PR #10, opened 2020-10-18
reconfigure baud rate in cargo-espflash when creating the serial connection

hanhossain:
`cargo espflash --board-info <serial>` was failing when leaving out the chip id. Fixed it by reconfiguring the baud rate right after creating the serial connection (just like in espflash).


icewind1991:
Thanks!

Issue #7, opened 2020-10-01
ESP32-PICO-D4 - Error: flash chip not supported


tobymurray:
I have an ESP32-PICO-D4 with 4MB of flash, I believe something like (but potentially not exactly) this: https://www.aliexpress.com/item/4000473398801.html

I'm following along the [xtensa-rust-quickstart](https://github.com/MabezDev/xtensa-rust-quickstart), and rying to use cargo-espflash from Crates.io got me some unexpected errors:

```
$ cargo espflash --chip esp32 --example esp32 --target xtensa-esp32-none-elf --speed 460800 --features="xtensa-lx-rt/lx6,xtensa-lx/lx6" /dev/ttyUSB0
thread 'main' panicked at 'Unable to parse command-line arguments: unused arguments left: --target, --speed, --features=xtensa-lx-rt/lx6,xtensa-lx/lx6', /home/toby/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-espflash-0.1.1/src/main.rs:13:29
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: cargo_espflash::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Installing from this repo directly (via `cargo install --git https://github.com/esp-rs/espflash/ cargo-espflash`) seems to have gone better, but I get "flash chip not supported":

```
cargo espflash --chip esp32 --example esp32 --speed 460800 --features="xtensa-lx-rt/lx6,xtensa-lx/lx6,esp32-hal" /dev/ttyUSB0
   Compiling xtensa-quickstart v0.1.0 (/home/toby/git/esp32/xtensa-rust-quickstart)
    Finished dev [unoptimized + debuginfo] target(s) in 0.66s
Error: flash chip not supported
```

Is this a limitation of the library currently, or potentially a bug? Whatever the solution, this may ultimately fix https://github.com/MabezDev/xtensa-rust-quickstart/issues/30

tobymurray:
As a sanity check that I have switched to an ESP-WROOM-32 and was successfully able to flash with the same command

icewind1991:
Can you try with the latest master, which should give you the detected flash id in the error message

tobymurray:
Tried with ad983cb3:

> Error: flash chip not supported, flash id: 0xff

ubergeek801:
Greetings,
[...]

PR #13, opened 2020-12-11
implementing alternate flash detection for e.g. ESP32-PICO-D4

ubergeek801:
This PR extends the flash chip detection mechanism as discussed in https://github.com/esp-rs/espflash/issues/7. To summarize:

1. Flash detection is performed as normal.
1. If detection succeeds, processing continues as normal.
1. If detection returns `0xFF`, alternate detection is attempted:
   a. `SpiAttach` is performed using parameters corresponding to the flash SPI pin configuration used by the ESP32-PICO-D4 (and possibly other devices)
   b. If detection succeeds, processing continues using the alternate SPI configuration; otherwise processing is aborted.

Thanks for reviewing!


icewind1991:
Nice,

I've reworked things a bit to make it easier to add more spi params later, can you enable "Allow edits from maintainers" on this PR?

ubergeek801:
"Allow edits by maintainers" is enabled. Might there be some repo-level settings I also have to tweak?

icewind1991:
I think it might be because the the PR comes from your `master`, I think I've had issues with that before.

I'll just merge this and push my changes, then you can verify if things are still working with the pico-d4

PR #20, opened 2021-08-02
Fix build std presumptions

### LINKED ISSUE #19: Cannot build rust-esp32-std-hello with `cargo espflash`
Calling `cargo espflash` in the root of the [rust-esp32-std-hello](https://github.com/ivmarkov/rust-esp32-std-hello) project like this:
```cargo espflash --tool cargo --chip esp32 /dev/tty.usbserial-0204D0A0```

Fails with:

>    Compiling rust-esp32-std-hello v0.4.0 (/Users/imarkov/projects/rust-esp32-std-hello)
error[E0463]: can't find crate for `panic_abort`

The reason is this [line](https://github.com/esp-rs/espflash/blob/master/cargo-espflash/src/main.rs#L204) where you are unconditionally passing `-Z build-std` to Cargo.

However, rust-esp32-std-hello expects `-Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort`, as you can see [here](https://github.com/ivmarkov/rust-esp32-std-hello/blob/main/.cargo/config.toml#L10).

To make callouts into `cargo build` a bit more flexible, I would suggest that you are not passing anything explicit to `cargo build` (aside from the optional `--release` flag) and perhaps support a `--` set of arguments at the end of the `cargo espflash` command line, where the semantics of `--` would be that all of those would be passed down to `cargo build` as is. E.g.:
```cargo espflash --tool cargo --chip esp32 /dev/tty.usbserial-0204D0A0 -- -Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort```

or just
```cargo espflash --tool cargo --chip esp32 /dev/tty.usbserial-0204D0A0```
... which, if you remove the explicit passing of `-Z build-std` would pick up all cargo build-std arguments from `.cargo/config.toml` and work.

MabezDev:
* Remove hardcoded passing of build-std args
* Instead note that the `--tool cargo` requires configuration in
  `.cargo/config{.toml}`

Closes #19

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -202,8 +202,8 @@ fn build(
     };
 
     if let "cargo" = tool {
-        args.push("-Z".to_string());
-        args.push("build-std".to_string());
+        println!("NOTE: --tool cargo currently requires the unstable build-std, ensure .cargo/config{{.toml}} has the appropriate options.");
+        println!("See: https://doc.rust-lang.org/cargo/reference/unstable.html#build-std")
     };
 
     args.push("--target".to_string());


icewind1991 (approved):

MabezDev:
Thanks @icewind1991, would you mind cutting a new release for `cargo-espflash`?

PR #21, opened 2021-08-06
Improve flashing reliability at higher speeds

### LINKED ISSUE #18: Compatibility with Slower Board
Hi,

I have [this older board](https://randomnerdtutorials.com/esp32-built-in-oled-ssd1306/) and it fails to work with espflash, but works fine with esptool.py.

After a bit of debugging, I found the issue to be here:

https://github.com/esp-rs/espflash/blob/45d5034d376f9d267bbab7dea29250cb706194a3/espflash/src/connection.rs#L52

When I change the first sleep from 100ms to 300ms, everything works fine. I suspect that something on that board is slower than usual, since the original version worked fine on an M5Stack.

Changing this delay time shouldn't have any impact on other boards, just 200ms slower writes.

KerryRJ:
I think that the USB to serial chips may be impacting the reliability of the transfer at higher speeds.
I have basically duplicated the esptool retry mechanisms and made sure that buffers are flushed during retries rather than delaying for longer periods.

The boards that I have that wouldn't flash at speeds of 460800 are now flashing faster than 115200.

Might fix #18.

### DIFF
--- a/espflash/src/flasher.rs
+++ b/espflash/src/flasher.rs
@@ -229,17 +229,18 @@ impl Flasher {
         self.connection
             .with_timeout(Duration::from_millis(100), |connection| {
                 let data = &[
-                    0x07u8, 0x07, 0x012, 0x20, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55,
+                    0x07, 0x07, 0x12, 0x20, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55,
                     0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55,
-                    0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55,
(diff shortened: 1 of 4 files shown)


MabezDev:
Thanks, I've tried this out and it hasn't made anything worse. I find with my board I need hold the reset button whilst flashing starts else I just get `Error: Failed to connect to the device`. Perhaps that's just my board.

I think this are positive changes non the less, so thank you!

MabezDev (approved):

MabezDev:
After switching board, this PR reliably flashes - so I think this is definitely worth merging.

ForsakenHarmony:
What's stopping this from being merged? the conflicts?

espflash/src/flasher.rs
```diff
@@ -249,16 +250,17 @@ impl Flasher {
...
-                        match connection.read_response()? {
-                            Some(_) => break,
-                            _ => continue,
-                        }
-                    }
-                }
                 Ok(())
-            })
+            })?;
+        for _ in 0..7 {
```
ForsakenHarmony, line 255: The nested loop here seems a bit awkward

icewind1991:
merged manually to deal with conflicts.

Thanks for the contribution

PR #29, opened 2021-08-18
Remove references to Xargo and cargo xbuild

### LINKED ISSUE #28: [Proposal] Remove Xargo/xbuild and only use build-std
The rationale behind this is simple - the less we have to support the better.

* All tools *require* nightly, with `build-std` being the only path to stable (even if it might be far away)
* With the introduction of the `build-std-features` flag, AFAIK `build-std` now has feature parity with `Xargo` and `xbuild`
* Switching means we can rely on cargo even more, for example cargo can handle this bit of code and avoid us hard-coding target names: https://github.com/esp-rs/espflash/blob/4921f6e50e0c08dfa33e3fd371cfa4978215eb80/espflash/src/chip/mod.rs#L106-L113
(Note that that bit of code is mainly why I am opening this, as it broke when we started upstreaming std targets).

Thoughts @jessebraham @icewind1991 @ivmarkov ?

MabezDev:
* Default to always use build-std.

Closes #28

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -8,7 +8,7 @@ use std::process::{exit, Command, ExitStatus, Stdio};
 use crate::cargo_config::has_build_std;
 use cargo_project::{Artifact, Profile, Project};
 use color_eyre::{eyre::WrapErr, Report, Result};
-use espflash::{Chip, Config, Flasher};
+use espflash::{Config, Flasher};
 use pico_args::Arguments;
 use serial::{BaudRate, SerialPort};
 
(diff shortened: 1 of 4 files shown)


cargo-espflash/src/main.rs
```diff
@@ -192,27 +160,14 @@ fn build(
...
-        "xbuild" => command.arg("xbuild"),
-        _ => unreachable!(),
-    };
-
-    if "cargo" == tool && !has_build_std(".") {
+    if !has_build_std(".") {
         println!("NOTE: --tool cargo currently requires the unstable build-std, ensure .cargo/config{{.toml}} has the appropriate options.");
-        println!("See: https://doc.rust-lang.org/cargo/reference/unstable.html#build-std")
+        println!("See: https://doc.rust-lang.org/cargo/reference/unstable.html#build-std");
+        // TODO early exit here
```
jessebraham, line 166: If you add the following lines at the top of the file:

```rust
#[cfg(unix)]
use std::os::unix::process::ExitStatusExt;
#[cfg(not(unix)]
use std::os::windows::process::ExitStatusExit;
```

Then you can return using:

```rust
exit_with_process_status(ExitStatus::from_raw(1))
```

Alternatively, it may be enough just to call `std::process::exit(1)` instead of using these wrapper functions, I can't actually remember why I included them.

jessebraham:
Thanks for this! Other than the one comment I left it looks great to me.

cargo-espflash/src/main.rs
jessebraham: Ahh I see you've resolved this in your other PR, please ignore :)

jessebraham (approved):

PR #31, opened 2021-08-19
Esp32s2 support

MabezDev:
A WIP PR for esp32s2 support.

* Removes the DATE_REG detection in favour of the CHIP_DETECT magic instead.
* Adds the esp32s2 module for s2 support.

I can't actually flash anything at the moment, the bootloader returns `InvalidMessage`:
```rust
Command: (10, DATA, 0)
Detected Esp32s2
Command: (13, DATA, 0)
Command: (10, DATA, 0)
Command: (10, DATA, 0)
Command: (9, DATA, 0)
Command: (9, DATA, 0)
Command: (9, DATA, 0)
Command: (9, DATA, 0)
Command: (9, DATA, 0)
Command: (10, DATA, 0)
Command: (10, DATA, 0)
Command: (9, DATA, 0)
Command: (9, DATA, 0)
Command: (13, DATA, 0)
Erase size: 12288
Command: (2, DATA, 0)
Error: bootloader returned an error: InvalidMessage
```
Any ideas?

Once I figure out whats going on, I think we can probably merge the esp32 & esp32s2 modules by expanding the `Chip` trait.

Builds on top of #29 & #30


jessebraham:
Refer to the following lines in `esptool.py`:

```python
params = struct.pack('<IIII', erase_size, num_blocks, self.FLASH_WRITE_SIZE, offset)
if isinstance(self, (ESP32S2ROM, ESP32S3BETA2ROM, ESP32S3ROM, ESP32C3ROM, ESP32C6BETAROM, ESP32H2ROM)) and not self.IS_STUB:
     params += struct.pack('<I', 1 if begin_rom_encrypted else 0)
```

Chips newer than the ESP32 must have an additional field in `BeginParams` denoting whether or not ROM encryption is being used. The command is being rejected by the bootloader because its length is 4 bytes too short.

icewind1991:
added the rom encryption parameter in a very hacky way for now, the flashing looks to go fine now.

It now crashes in the bootloader with an "invalid partition table error"

```
I (24) boot: ESP-IDF v4.4-dev-1404-gc13afea635-dirty 2nd stage bootloader
I (24) boot: compile time 20:23:13
I (25) boot: chip revision: 0
I (29) boot.esp32s2: SPI Speed      : 80MHz
I (33) boot.esp32s2: SPI Mode       : DIO
I (38) boot.esp32s2: SPI Flash Size : 2MB
I (43) boot: Enabling RNG early entropy source...
E (48) flash_parts: partition 0 invalid - offset 0x10000 size 0x3f0000 exceeds flash chip size 0x200000
E (58) boot: Failed to verify partition table
E (63) boot: load partition table error!
```

espflash/src/flasher.rs
```diff
@@ -138,6 +138,16 @@ struct BeginParams {
     offset: u32,
 }
 
+#[derive(Zeroable, Pod, Copy, Clone, Debug)]
+#[repr(C)]
+struct BeginParamsS2 {
```
jessebraham, line 143: This `encrypted` field applies to all chips except the ESP8266/ESP32 as far as I know. This should probably be the default struct, and we can have a `BeginParamsLegacy` or something to that effect instead.

icewind1991: I was thinking of having some way of having per-chip command/parameter structs or otherwise moving some of the flashing logic to the chips.

PR #35, opened 2021-09-02
Add support for flashing the ESP32-C3

jessebraham:
I was able to successfully build and flash [MabezDev/esp32c3-experiments](https://github.com/MabezDev/esp32c3-experiments) using this PR.

Note that the bootloader for this chip has configured the minimum revision as 0, and the flash size as 4MB. This should support most boards until we have a better solution.

~~This builds upon #31, and will need to be rebased before it can be merged.~~
I have modified this PR to be based off `master` instead. I have included most of the refactoring that had previously been done in #31.


icewind1991 (approved):
Code looks ok and flashing works

MabezDev (approved):
Looks good to me!

PR #37, opened 2021-09-03
Write the ESP32 & ESP32-C3 partition table with nvs and phy_init partitions

kelnos:
Uses the ESP-IDF defaults for the offsets and sizes of the nvs and phy_init partitions.  The factory partition was originally being written in the ESP-IDF default location, so it didn't need to move.

This also changes the `Partition` constructor; there's no need to pass the partition type since it can be inferred from the subtype.


espflash/src/chip/esp32.rs
```diff
@@ -49,7 +55,31 @@ impl ChipType for Esp32 {
     ) -> Box<dyn Iterator<Item = Result<RomSegment<'a>, Error>> + 'a> {
         let bootloader = include_bytes!("../../bootloader/esp32-bootloader.bin");
 
-        let partition_table = PartitionTable::basic(0x10000, 0x3f0000).to_bytes();
+        let partition_table = PartitionTable::new(
```
ForsakenHarmony, line 58: Might also want to do this for the c3?

ForsakenHarmony: Seems it's the exact same layout
https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-guides/partition-tables.html

ForsakenHarmony: Might make sense to just edit it in `partition_table.rs`, because it's the same for the s2 as well

kelnos: Ah, for some reason I thought the c3 was different -- at least it might be sometimes?  I remembered seeing this: https://github.com/ivmarkov/rust-esp32-std-hello/#flashing-for-esp32-s2--esp32-c3

But if the s2 is the same, at least, makes sense to put it somewhere common.  And either way, I guess this 3-part table is perhaps more useful for all boards than the basic one there now.

kelnos: Anyhow, this is done

icewind1991 (approved):

espflash/src/chip/esp32.rs
```diff
@@ -62,7 +62,8 @@ impl ChipType for Esp32 {
             PHY_INIT_DATA_SIZE,
             APP_ADDR,
             APP_SIZE,
-        );
+        )
+        .to_bytes();
```
kelnos, line 66: Ugh, that was dumb. Thanks for fixing that up.

PR #38, opened 2021-09-03
Use clap instead of pico-args for cargo-espflash

jessebraham:
I've been wanting to do this for awhile and finally got around to it. While part of the reason was purely personal preference, this also (IMO) makes it easier for us to add options, performs better validation of command-line arguments, and doesn't require manual updates of the argument parsing function and help messages.

As a note, I did remove the `--chip` option as it was not actually being used (we perform auto-detection). I am fine with adding it back for compatibility reasons, but I personally don't see the need.

### DIFF
--- a/cargo-espflash/Cargo.toml
+++ b/cargo-espflash/Cargo.toml
@@ -1,21 +1,34 @@
 [package]
 name = "cargo-espflash"
 version = "0.1.4"
-description = "Cargo subcommand for flashing the ESP8266 and ESP32 over serial"
-license = "GPL-2.0"
 authors = [
     "Robin Appelman <robin@icewind.nl>",
     "Jesse Braham <jesse@beta7.io>",
 ]
-repository = "https://github.com/esp-rs/espflash"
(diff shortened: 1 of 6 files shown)


N3xed:
Why not [`structopt`](https://crates.io/crates/structopt), I think that would make it more readable. But that's just my opinion.

MabezDev (approved):
I agree that structopt is nice! However either is a big improvement over picoargs, so either works for me :)

jessebraham:
I would rather focus my efforts on other functionality at this point. Anybody is of course welcome to make further changes on top of mine, I just don't feel strongly enough either way to justify spending any more time on this.

N3xed:
> I would rather focus my efforts on other functionality at this point. Anybody is of course welcome to make further changes on top of mine, I just don't feel strongly enough either way to justify spending any more time on this.

All good, was just a suggestion (I just like `structopt` :D).

icewind1991 (approved):

Issue #26, opened 2021-08-18
Flashing custom partition table & bootloader


ivmarkov:
The crate has a very nice support for describing custom partition tables. It is just that this functionality is not exposed yet to the outside world, and currently, a partition table with a single factory app is [always flashed](https://github.com/esp-rs/espflash/blob/2c86f48ea3a61624494f35add1ac590d4ca12138/espflash/src/chip/esp32.rs#L65).

We need a command-line syntax in the form of an option to read a partition table description from a file, so that one can flash their own partition tables. Ideally, the partition table file format should follow the one used in the [ESP-IDF CSV format](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html), so that we do not bring more confusion.

Regarding flashing a custom bootloader - this is less of a priority, but same here - we just need a command line option, where the user can provide their own elf or bin image of the bootloader (not sure which one would be more appropriate - perhaps elf, for symmetry with the app image).

Finally, having these two options (a custom bootloader and a custom partition table) brings the question: should it be possible to separately flash each of these (bootloader, partition table, app image), or shall we stick with the "platformio" approach, where all of these are flashed always together. Not sure what is the answer here.

jessebraham:
I have begun working on the custom partition table portion of this issue. I have an initial implementation ready and just need to do some testing/verification on hardware, then write some unit tests. Hopefully won't be long, though. Once this is complete I will implement the bootloader feature.

At present I am keeping the flashing process as-is just to minimize my pain while developing these features :grin: I do not feel strongly one way or another regarding the flashing process (it's adequate for my needs), so I would like to hear some other people's opinions.

@MabezDev @icewind1991 @georgik

MabezDev:
> Finally, having these two options (a custom bootloader and a custom partition table) brings the question: should it be possible to separately flash each of these (bootloader, partition table, app image), or shall we stick with the "platformio" approach, where all of these are flashed always together. Not sure what is the answer here.

Perhaps in the future, we could leverage cargo fingerprinting to only flash the partition table and bootloader if they have changed. I think for now, flashing everything is the simplest way forward.

jessebraham:
With #40 and #41 merged this issue can likely be closed. I agree with @MabezDev that just doing everything in one go is the simplest option for now. If there is a need for splitting out this process it can always be done when it's needed.

PR #40, opened 2021-09-07
Add support for providing custom partition tables

jessebraham:
Adds a `--partition-table` option to _cargo-espflash_ which, if provided, attempts to parse the CSV file at the given path and produce a usable `PartitionTable`. If this option is not provided then the default table is generated via `PartitionTable::basic` instead.

This was requested in #26.

### DIFF
--- a/espflash/src/partition_table.rs
+++ b/espflash/src/partition_table.rs
@@ -1,58 +1,94 @@
-use std::io::Write;
-
 use md5::{Context, Digest};
+use regex::Regex;
+use serde::{Deserialize, Deserializer};
+
+use std::{error::Error, io::Write};
 
 const MAX_PARTITION_LENGTH: usize = 0xC00;
 const PARTITION_TABLE_SIZE: usize = 0x1000;
+const MAX_PARTITION_TABLE_ENTRIES: usize = 95;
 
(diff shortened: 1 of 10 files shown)


espflash/src/partition_table.rs
```diff
@@ -209,29 +336,62 @@ impl<W: Write> HashWriter<W> {
...
-        PHY_INIT_DATA_SIZE,
-        APP_ADDR,
-        APP_SIZE,
-    );
-
-    let result = table.to_bytes();
-
-    assert_eq!(expected.len(), result.len());
-    assert_eq!(expected, result.as_slice());
+#[allow(unused)]
```
jessebraham, line 339: For some reason without this line  the `use super::*;` line as well as the `PTABLE_0` and `PTABLE_1` constants get marked as unused, and fail the clippy checks. Has anybody experienced this before?

icewind1991: the `unused` check is done without the `test` cfg, so anything that is only used by tests is seen as unused, adding a `#[cfg(test)]` should fix this

jessebraham: Oops, forgot about that. Thanks!

icewind1991 (approved):

PR #44, opened 2021-09-11
improved error messages

icewind1991:
uses [`miette`](https://github.com/zkat/miette) for error reporting and extends the error messages to hopefully be more useful for users.

Some examples:

image
image
image


jessebraham (approved):
This looks pretty good to me. I wasn't familiar with `miette`, very nice error messages!

PR #45, opened 2021-09-15
allow specifying partition table and bootloader in the projects Cargo.toml

icewind1991:
Allow specifying the partition table and bootloader in the project configuration to save users from having to specify the command line options every time.

This currently has a hard requirement on the file extensions of partition table and bootloader files, to keep the option open to have different non-filename options for these fields in the future (maybe specifying a crate name for the bootloader or something)


jessebraham (approved):
LGTM!

PR #46, opened 2021-09-17
use elf sections instead of elf segments for flashing

icewind1991:
While esp8266, esp32 and esp32c3 seem to work fine with the "almost the same data" in the segments so far, esp32s2 bootloader seems to be a bit more picky (or maybe the linker script differences?)

### DIFF
--- a/espflash/src/elf.rs
+++ b/espflash/src/elf.rs
@@ -3,7 +3,8 @@ use std::cmp::Ordering;
 
 use crate::chip::Chip;
 use crate::flasher::FlashSize;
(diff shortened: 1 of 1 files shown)


icewind1991:
esp8266, 32 and 32c3 all still work from my quick testing

jessebraham (approved):
LGTM! I have tested on the ESP32 and ESP32-C3 with no problems, and I rebased my branch on top of here and was able to flash the ESP32-S2 as well. Thank you for figuring this out.

PR #48, opened 2021-09-17
merge adjecent segments and better fit ram segments between rom segments

icewind1991:
Makes the resulting image closer the `esptool` output and reduces the image size in some cases


jessebraham (approved):
Sorry for the delay in getting to this. Everything here looks okay to me, seems to be working as intended. Tested on ESP32-C3 and ESP32-S2.

Thank you again for your work!

PR #49, opened 2021-09-17
fallback to regular flashing if compressed flashing times out

icewind1991:
For some unknown reason, sometimes the `FlashDeflateBegin` times out for me, while `FlashBegin` never seems to have this issue.

To prevent this from being an issue for users while still providing the highest potential flashing speeds, this adds a failover mechanism, where it first tries to do a compressed flash. And if the compressed flash times out it will fallback to the regular flash.

### DIFF
--- a/espflash/src/chip/mod.rs
+++ b/espflash/src/chip/mod.rs
@@ -4,13 +4,14 @@ use strum_macros::Display;
 use crate::{
     elf::{update_checksum, CodeSegment, FirmwareImage, RomSegment},
     error::{ChipDetectError, FlashDetectError},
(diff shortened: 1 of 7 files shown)


icewind1991:
If I first reset the device with `esptool.py flash_id` the compressed flash always works, so `esptool` is doing something different during reset/connect to make things work...

icewind1991:
Even more interesting, just opening the serial port from python before flashing makes things work...

icewind1991:
Ok, got it: https://github.com/esp-rs/espflash/pull/50

If we find another case where compressed flashing fails but regular flashing work we can reopen this

PR #50, opened 2021-09-19
disable software flowcontrol for serial

icewind1991:
Having software flow control enabled causes `FlashDeflateBegin` to time out for some reason


jessebraham (approved):
Good find, thanks!

PR #56, opened 2021-09-22
use the correct chip for determining if an address is a flash or rom segment

icewind1991:
missed this one while factoring out flash image building for esp32 chips :see_no_evil:


Vandell63:
@icewind1991  thx for this. I'm starting using it, but i expected it was an issue coming from my configuration.
Indeed it was not. Thanks again. Best regards

PR #58, opened 2021-09-25
split image formats into it's own modules

icewind1991:
for now it still only supports the default esp8266 and esp32 formats, but this paves the way for adding support for alternative formats later on


jessebraham (approved):
Seems okay to me, thanks! Tested with _ESP32_, _ESP32-C3_, and _ESP32-S2_.

PR #59, opened 2021-09-27
(in)validate docker cache for rust-xtensa clone

nim65s:
Hi,

This allow to let Docker know when it should update its git clone of esp-rs/rust, by using the Last-Modified HTTP header provided by Github,
It is not exact, but I find it a good compromise  between cache hit, speed, and rebuild triggering when there is anything new.


icewind1991 (approved):
Nice!

PR #65, opened 2021-10-01
Add a method of reading eFuse values

### LINKED ISSUE #39: Add a way to read eFuse values
There is some additional information we can get from chips by reading the eFuse, such as revision, MAC address, etc. It would be nice to have this functionality, however I believe this may require per-chip implementations.

jessebraham:
As the title says, adds a way to read eFuse values. Various information such as silicon revision, MAC address, chip features, etc. are accessible via the eFuse memory.

This PR contains a small refactor of the `chip` module as well, just slightly better separation of concerns.

I have additionally moved the `board_info` function out of `cargo-espflash` into the `Flasher` module, so that it can also be used by `espflash` without duplication.

Closes #39.

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -141,11 +141,13 @@ fn main() -> Result<()> {
         None
     };
 
-    // Connect the Flasher to the target device. If the '--board-info' flag has been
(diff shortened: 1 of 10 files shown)


icewind1991 (approved):
Looks good

PR #68, opened 2021-10-03
add cargo-espflash subcommand to save the generated image to disk

### LINKED ISSUE #57: Feature request: provide functionality similar to `esptool.py elf2image`
Subject says it all.

While having the flasher processing ELF files directly is gorgeous (and saves a ton of explanations to folks who are just starting with embedded development with the ESP), generating a `.BIN` file from an `.ELF` image does have one use case:
- OTA updates

icewind1991:
Allows saving the rom image instead of flashing it

- No attempt to connect to a device is made
- Image format is determined by matching against the configured target to auto detect the chip
- If an image contains multiple segments (esp8266) the output paths are prefixed by the address for the segment

Fixes #57

### DIFF
--- a/cargo-espflash/src/error.rs
+++ b/cargo-espflash/src/error.rs
@@ -44,12 +44,31 @@ pub enum Error {
     #[error(transparent)]
     #[diagnostic(transparent)]
     UnsupportedTarget(UnsupportedTargetError),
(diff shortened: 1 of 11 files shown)


jessebraham (approved):
Looks good!

PR #75, opened 2021-10-18
Serial port auto-detection

### LINKED ISSUE #52: Serial port auto-detection
esptool.py implements serial port auto-detection logic, which lets users omit the port name in many cases. Similar logic could be implemented in this project.

Additionally, if the serial port is provided by a USB device and we can get enough information from the OS about the serial port to map it to a USB device, we can use VID:PID pairs to detect likely candidates for the development board serial ports. This should work pretty well e.g. for ESP32-C3 built-in USB_SERIAL_JTAG.

jessebraham:
Will eventually close #52.

Tested various chips and functionality on macOS, and minimal testing on Windows 10.

* The `serial` crate was swapped out with `serialport`
    * `serial` has not received an update in ~4 years, `serialport` is somewhat maintained
    * `serialport` was almost a drop-in replacement, and gives us the `available_ports` function as well
* Implements fairly basic port detection for `cargo espflash` and `espflash`
    * If no port has been specified by CLI or config, attempts auto-detection
    * If only one port is detected, it is selected automatically
    * If multiple ports are detected, the user is prompted to select one

Presently I am only filtering serial ports by type, allowing USB ports only. Questions regarding this:
1) Is there any reason to support other port types (other options are `BluetoothPort` or `PciPort`)?
2) If we decide to move forward with VID/PID filtering, what should we begin with? Should we omit unrecognized ports entirely or just mark them as such?

~~I will likely move the port detection logic into the `espflash` crate, especially if there is any substantial filtering logic, but have just included it for `cargo-espflash` until we flesh things out.~~

cc @igrr @icewind1991

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,9 +10,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: musl-tools
+      - name: build dependencies
         run: |
-          sudo apt-get install musl-tools
+          sudo apt-get install musl-tools libudev-dev
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -7,6 +7,9 @@ jobs:
     name: Check
     runs-on: ubuntu-latest
     steps:
+      - name: build dependencies
+        run: |
+          sudo apt-get install musl-tools libudev-dev
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
@@ -22,6 +25,9 @@ jobs:
     name: Doc Test
     runs-on: ubuntu-latest
     steps:
+      - name: build dependencies
+        run: |
+          sudo apt-get install musl-tools libudev-dev
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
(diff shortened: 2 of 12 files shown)


icewind1991:
Looks good overall

> Is there any reason to support other port types (other options are BluetoothPort or PciPort)?

I would say ignore those until people ask for it

> If we decide to move forward with VID/PID filtering, what should we begin with? Should we omit unrecognized ports entirely or just mark them as such?

I would suggest the following.

1. If one serial is detected that matches the VID/PID, use it without confirmation
2. If one serial is detected that doesn't match the VID/PID, ask for confirmation
3. When multiple ports are detected, show both the matching and non matching, but distinguish them clearly.
4. Allows users to add their own VID/PID pairs in the config (and hint them how in case 2.)

jessebraham:
Thank you for the feedback! I believe this should be functioning as described now. Please let me know if you spot any issues or would like any further changes made.

icewind1991:
Did some changes:

- enable default features for `serialport` to get port info working on linux
- allow configuring multiple usb devices in config
- show product name when selecting serial
- allow exiting out of port selection

Will merge this soon unless something shows up, then collect some common vid/pid pairs for a follow up so we can "do the right thing"™ in most cases

jessebraham:
Great, LGTM! I don't currently have a Linux box that I can physically access so thank you for catching and fixing that issue.

PR #81, opened 2021-10-22
Add support for flashing the ESP32-S3

jessebraham:
Should be pretty straight-forward. Not 100% sure about the values in `Esp32Params` but it does work at least for basic examples.


jessebraham:
This has been rebased on `master` and should be (I think) ready to merge.

MabezDev (approved):

PR #86, opened 2021-10-30
Unify cli

icewind1991:
- Move to clap 3 beta and it's derive macros
- Move some bits from `cargo-espflash` to `espflash` so it can be reused in both
- Add missing cli features from `cargo-espflash` to `espflash` (`--monitor` and `save-image`)

Fixes #77, #78

PR #87, opened 2021-11-01
Fix typo in usage help

a1ien:


a1ien:
I saw  Unify cli #86  sorry for noise

PR #88, opened 2021-11-02
Allow for the build target to be specified via the command line as well

jessebraham:
See title. Command-line argument takes priority over cargo config.

cc @georgik

PR #89, opened 2021-11-04
Explicitly set build target to use --target value if provided

jessebraham:
I made a pretty silly mistake in #88 in that I wasn't actually providing the target to cargo. This should fix that.

I also made the `supports_target` function more strict and changed it to use a default implementation. Using the `starts_with` method it is still possible to pass an invalid target value.

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -159,14 +159,23 @@ fn build(
     let target = build_options
         .target
         .as_deref()
-        .or(cargo_config.target())
+        .or_else(|| cargo_config.target())
         .ok_or_else(|| NoTargetError::new(chip))?;
+
+    let chip = if chip.is_some() {
+        chip
+    } else {
+        Chip::from_target(target)
+    };
 
     if let Some(chip) = chip {
         if !chip.supports_target(target) {
             return Err(Error::UnsupportedTarget(UnsupportedTargetError::new(target, chip)).into());
         }
+    } else {
+        return Err(Error::UnknownTarget(target.to_string()).into());
     }
+
     // The 'build-std' unstable cargo feature is required to enable
     // cross-compilation for xtensa targets.
     // If it has not been set then we cannot build the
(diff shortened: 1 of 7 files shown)


cargo-espflash/src/main.rs
```diff
@@ -252,9 +263,17 @@ fn save_image(
...
     cargo_config: CargoConfig,
 ) -> Result<()> {
-    let target = cargo_config
-        .target()
-        .ok_or_else(|| NoTargetError::new(None))?;
+    let target = matches
+        .build_args
+        .target
+        .as_deref()
+        .or_else(|| cargo_config.target());
```
icewind1991, line 270: why not use the `ok_or_else` here as before, instead of the `is_none` and `unwrap` there is now

espflash/src/chip/mod.rs
```diff
@@ -69,7 +69,9 @@ pub trait ChipType: ReadEFuse {
         Ok(bytes_to_mac_addr(bytes))
     }
 
-    fn supports_target(target: &str) -> bool;
+    fn supports_target(target: &str) -> bool {
```
icewind1991, line 72: I intentionally made the `supports_target` check loser than the `SUPPORTED_TARGETS` list to allow using customized/future targets for the device that `espflash` doesn't know about.

icewind1991 (changes requested):

cargo-espflash/src/main.rs
jessebraham: Of course, thanks for pointing that out.

espflash/src/chip/mod.rs
jessebraham: Okay that's fair, thank you for giving your rationale.

My main concern with this approach is that you are able to provide any arbitrary number of targets assuming they have the correct prefix, regardless of whether or not these targets exists. This bypasses any error handling currently done by `miette` and will result in `rustc` throwing an error instead. To me, this defeats the purpose of even checking the target in the first place.

I'm not sure that custom targets will really be a common issue, and since we control the Rust compiler fork we will know of any new targets being added ahead of time and can act accordingly.

If you feel strongly about this then I will revert these changes, though.

icewind1991 (approved):

Issue #90, opened 2021-11-08
Wait for board to enter flashing mode


hongquan:
Some boards have hard buttons to manually switch to flashing mode. In that case, espmonitor often fails because it doesn't wait for human to press the button.

Please support this case.

jessebraham:
Now that #92 has been merged `espflash` will make multiple connection attempts, so this should give you a chance to press the BOOT button.

If you feel this issue is still not resolved please feel free to re-open it explaining why!

PR #92, opened 2021-11-18
Retry failed connection attempts

### LINKED ISSUE #18: Compatibility with Slower Board

jessebraham:
~~This is built on top of #75, so I will convert it from a draft once that gets merged.~~

I believe this should resolve #18 and #90.

This PR emulates the connection logic performed by `esptool.py` and should help with connection issues on older devices. Essentially this adds two things:

1) Retry the connection until we connect to the device, or reach the attempt limit
2) Alternate between the default delay and a longer delay, which helps with older devices (eg. ESP32 rev 0)

### DIFF
--- a/espflash/src/connection.rs
+++ b/espflash/src/connection.rs
@@ -56,7 +56,7 @@ impl Connection {
         Ok(())
     }
 
-    pub fn reset_to_flash(&mut self) -> Result<(), Error> {
+    pub fn reset_to_flash(&mut self, extra_delay: bool) -> Result<(), Error> {
         self.serial.write_data_terminal_ready(false)?;
         self.serial.write_request_to_send(true)?;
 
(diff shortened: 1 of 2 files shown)


jessebraham:
Rebased on master, should be good to go.

PR #94, opened 2021-11-19
Update all READMEs, add missing information

jessebraham:
Our READMEs were getting a little outdated in some places, so I just went ahead and revamped all of them. I believe I have included all the information which was previously in them, as well as added a few more points.

Please let me know if there are any issues or anything else you'd like to see added.

#### Rendered:

[README.md](https://github.com/jessebraham/espflash/blob/docs/readme-updates/README.md)  
[cargo-espflash/README.md](https://github.com/jessebraham/espflash/blob/docs/readme-updates/cargo-espflash/README.md)  
[espflash/README.md](https://github.com/jessebraham/espflash/blob/docs/readme-updates/espflash/README.md)

PR #96, opened 2021-11-22
Add `--package` option

ducktec:
Examples in a project (or just the project in general) might be organized as packages within a cargo workspace.

To make `cargo espflash` handle those scenarios, the `--package` option has to be supported and forwarded to the build step. If a binary crate was selected with the `--package` option, the build step will return the path to the generated binary in the shared `target` folder just like in the other scenarios. 

If a library crate was selected with `--package`, an error `cargo_espflash::no_artifact` will be issued. The help text for that error has been updated as part of the PR.

I've also updated the READMEs by just adding the option to the command help section, but since these sections are not in line with the code right now (see #94), I propose to rebase and properly document this PR once #94 has been merged. Therefore I've marked it as WIP.

### DIFF
--- a/cargo-espflash/README.md
+++ b/cargo-espflash/README.md
@@ -46,6 +46,9 @@ OPTIONS:
         --monitor
             Open a serial monitor after flashing
 
+        --package <PACKAGE>
+            Specify a (binary) package within a workspace to be built
+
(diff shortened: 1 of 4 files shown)


jessebraham (approved):
Looks good, thanks for this!

PR #99, opened 2021-11-27
add list of known serial adapters used on dev boards

icewind1991:
serial ports from these adapters are used by default when auto detecting serial port.


jessebraham (approved):
Looks good, thanks for adding this!

PR #100, opened 2021-11-27
ask to save vid/pid in config if an unrecognized adapter is used

icewind1991:
When a user picks a serial adapter that isn't from a known dev board, offer to add the vid/pid pair to the config for easier use next time.

Depends on:
- [x] https://github.com/esp-rs/espflash/pull/99


jessebraham (approved):
LGTM!

PR #102, opened 2021-12-09
Update clap dependency and address the breaking changes

jessebraham:
Clap's move from `beta.5` to `rc.0`/`rc.1` introduced some breaking changes, namely the splitting-out of `clap_derive`. I have updated the `clap` dependencies, added `clap_derive`, and fixed the resulting errors.


jessebraham:
This seems to have caused some issues with the command-line argument parsing now, according to the new help messages. Will address this shortly.

PR #106, opened 2021-12-13
Support multiple storage partitions and empty partition offsets

### LINKED ISSUE #104: `partitions.csv` rejected with blank offset fields
According to https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html it is perfectly valid to leave the offset fields blank. However, espflash rejects these `csv`s as malformed.

### LINKED ISSUE #105: `partitions.csv` rejected when it contains 2 spiffs partitions
https://github.com/esp-rs/espflash/blob/master/espflash/src/partition_table.rs#L289 checks for equality of partition subtypes and rejects the partition table when it finds two with the same subtype. It's not exactly well specified what should happen with component-specific subtypes, but I've been using two spiffs partitions in my esp-idf based project for the better part of the year without issues, so I believe those (`spiffs` for sure, maybe also `fat`) can be added as an exception.

bugadani:
Closes #104 
Closes #105

### DIFF
--- a/espflash/src/partition_table.rs
+++ b/espflash/src/partition_table.rs
@@ -115,6 +115,12 @@ pub enum DataType {
     Spiffs = 0x82,
 }
 
+impl DataType {
+    fn is_multiple_allowed(self) -> bool {
+        matches!(self, Self::Fat | Self::Spiffs)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


icewind1991:
Thanks

PR #110, opened 2021-12-19
add fallback serial detect when musl

icewind1991:
serialport's autodetect doesn't provide any port information when using musl linux we can do some manual parsing of sysfs to get the relevant bits without udev.

This doesn't handle as many cases as udev proper (it doesn't give the product name for half of the devices in my testing) and probably has more things not covered but it's better than nothing :)


jessebraham (approved):
Looks fine to me, though I'm unfortunately not really familiar with `musl` so I'm kind of assuming it's correct 😄

PR #112, opened 2021-12-21
Support flashing to partitions other than `factory`

### LINKED ISSUE #111: Cannot flash OTA partition
I have a project where I have 2 ota partitions and no factory partition. espflash.exe fails flashing an image with the error `  ╰─▶ No factory app partition was found`.

bugadani:
Resolves #111 

Hoping this is sufficient and correct. This PR changes valid partition table criteria to accept the partition table if any partition exists with the "app" type. I'm not sure if app with subtype test should be rejected here.


bugadani:
Opened as draft because I haven't tried the built executable yet.

icewind1991:
instead of accepting partition tables without factory it might make more sense to add a "target partition" option to determine which app partition to flash to

bugadani:
That might be an improvement, but I'd like to keep a "just pick one for me" option. esp-idf seems to do this, although it can lead to some weirdness: the active ota firmware might be on the second partition, and then flashing the first partition would be silently ineffective.

There is certainly room for improvement here, I've only implemented what I needed right now.

jessebraham (approved):
Thanks for this, I think it is sufficient. We can expand on it in the future if necessary. I was mistaken initially in thinking that the `factory` partition was required, so this is closer in line with the documentation.

espflash/src/error.rs
```diff
@@ -469,19 +469,19 @@ impl InvalidSubTypeError {
 }
 
 #[derive(Debug, Error, Diagnostic)]
-#[error("No factory app partition was found")]
+#[error("No app partition was found")]
 #[diagnostic(
-    code(espflash::partition_table::no_factory_app),
-    help("Partition table must contain a factory app partition")
+    code(espflash::partition_table::no_app),
```
bugadani, line 474: One question: I'm not familiar with the `diagnostic` macro, is this change valid?

icewind1991 (approved):

PR #118, opened 2021-12-31
Fix image creation for direct boot

bjoernQ:
This goes together with: https://github.com/MabezDev/esp32c3-experiments/pull/5

~~It just doesn't produce gaps between the segments.~~

Edit:
Turned out the code used the VMA addresses of the sections but it's needed to use the LMA

Some additional logging statements revealed this:
```
CodeSegment addr=42000000 len=8, new addr = 0
CodeSegment addr=42000008 len=b378, new addr = 8
CodeSegment addr=3c00b380 len=32d8, new addr = b380
CodeSegment addr=3fc80000 len=c, new addr = 80000
length = 524300
```

So, it added 478336 zero bytes before the _.data_ section

Now it looks like this
```
CodeSegment addr=0 len=8, new addr = 0
CodeSegment addr=8 len=b378, new addr = 8
CodeSegment addr=b380 len=32d8, new addr = b380
CodeSegment addr=e660 len=c, new addr = e660
length = 58988
```

Also, the final image looks like this when using _probe-rs_ for flashing.

### DIFF
--- a/espflash/src/elf.rs
+++ b/espflash/src/elf.rs
@@ -7,6 +7,7 @@ use crate::flasher::FlashSize;
 use std::fmt::{Debug, Formatter};
 use std::mem::take;
 use std::ops::AddAssign;
+use xmas_elf::program::Type;
 use xmas_elf::sections::{SectionData, ShType};
 use xmas_elf::ElfFile;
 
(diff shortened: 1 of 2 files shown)


icewind1991:
I don't think this is correct, the padding should be required when merging segments that have blank space between them, otherwise the memory mapping wont line up with the flash addresses

bjoernQ:
In my understanding the flash just gets mapped twice linearly - certainly the LMA and VMA needs to be configured accordingly in the linker script

The current implementation generates a 500k image for the Rust sample because of the gaps.

bjoernQ:
But I agree it's probably solvable just with the linker script and no change here.  Would be better but not sure yet

Will look into that tomorrow

jessebraham (approved):
LGTM, thanks! Seems to have fixed the incorrect output for `DATA_SECTION_TEST` in the [esp32c3-experiments](https://github.com/MabezDev/esp32c3-experiments/blob/master/src/main.rs#L36) repository as intended.

Issue #74, opened 2021-10-18
Not working in windows


stevefan1999-personal:
Hi, I wonder why it is not doing okay in Windows? I'm using [nanoESP32-C3](https://github.com/wuxx/nanoESP32-C3) which should be compatible, but:
```pwsh
PS E:\Git\github.com\MabezDev\esp32c3-experiments> cargo espflash --speed 115200 COM4
Serial port: COM4
Connecting...

Error: espflash::connection_failed
  × Error while connecting to device
  ╰─▶ Failed to connect to the device
  help: Ensure that the device is connected and the reset and boot pins are not being held down

```
Which I can flash correctly with esptool

mladedav:
This exact error is printed when you're not in flashing mode, i.e. you didn't restart the device while G0 connected to GND. I don't know if that's your issue (since you're able to flash with esptool) but maybe this can help someone.

JurajSadel:
Hi @stevefan1999-personal, did [this]( https://github.com/esp-rs/espflash/commit/85c8eb84db7f0a534d73772069f864475ff4dfe6) resolve your issue or you are still having the same issue?

PR #122, opened 2022-01-17
Allow for flashing via the USB Serial JTAG peripheral

### LINKED ISSUE #93: builtin serial/jtag - unable to flash via serial
The builtin serial/jtag peripheral of the c3 & s3 also exposes a serial port.

With recent versions of esptool it is possible to flash via the serial port, however it is not possible to flash with espflash currently (it hangs).

To reproduce:

1. Wire up a USB cable to the serial/jtag peripheral pins.
2. lsusb should report something like this:
    ```
    Bus 000 Device 010: ID 303a:1001 303a USB JTAG/serial debug unit  Serial: 7C:DF:A1:61:E4:80
    ``` 
3. A serial port should also show up, on mac/linux it comes up under `/dev/ttyX` (windows untested)
4. Try and use espflash over this port

jessebraham:
Thanks to @JurajSadel for his work on this!

As stated in the title, this adds support for flashing via the ESP32-C3's USB Serial JTAG peripheral. This required performing a small refactor for `get_serial_port` to instead return a `SerialPortInfo` struct, as we needed access to the `PID` later on.

Has been tested on macOS x86_64 successfully. I will test it on Windows later today, and I believe @MabezDev will be able to test on an M1 Mac.

Closes #93.

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -1,31 +1,38 @@
-use self::clap::ConnectArgs;
 /// CLI utilities shared between espflash and cargo-espflash
 ///
-/// No stability guaranties applies
+/// No stability guaranties apply
 use config::Config;
 use miette::{Result, WrapErr};
-use serialport::FlowControl;
+use serialport::{FlowControl, SerialPortType};
 
-use crate::cli::serial::get_serial_port;
-use crate::{error::Error, Flasher};
+use self::clap::ConnectArgs;
+use crate::{cli::serial::get_serial_port_info, error::Error, Flasher};
 
 pub mod clap;
 pub mod config;
-mod line_endings;
 pub mod monitor;
+
+mod line_endings;
 mod serial;
 
 pub fn connect(matches: &ConnectArgs, config: &Config) -> Result<Flasher> {
-    let port = get_serial_port(matches, config)?;
+    let port_info = get_serial_port_info(matches, config)?;
 
     // Attempt to open the serial port and set its initial baud rate.
-    println!("Serial port: {}", port);
+    println!("Serial port: {}", port_info.port_name);
     println!("Connecting...\n");
-    let serial = serialport::new(&port, 115_200)
+    let serial = serialport::new(&port_info.port_name, 115_200)
         .flow_control(FlowControl::None)
         .open()
         .map_err(Error::from)
-        .wrap_err_with(|| format!("Failed to open serial port {}", port))?;
+        .wrap_err_with(|| format!("Failed to open serial port {}", port_info.port_name))?;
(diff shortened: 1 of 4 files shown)


MabezDev:
Hmm this isn't working for me anymore on linux (Mac M1 won't work until we fix the serial library), but the [original branch](https://github.com/MabezDev/espflash/tree/usb-serial-jtag) still works for me via the usb jtag serial?

MabezDev (approved):
I take that back actually, it does work! However sometimes it seems it can get into a state where no matter how many retries it won't work until its power cycled :thinking:. Possibly this is a quirk of the hardware peripheral, maybe we can test with esptool.py.

For now, this is definitely an improvement, so lets get it landed and figure out the quirks later on.

jessebraham:
Thanks for taking a look! I did have that happen once right when I started testing, but it did not happen again so I chalked it up to PEBKAC. I will take a second look at it today and compare with `esptool.py` so see if we're missing anything.

bjoernQ:
I tried it on Windows (Windows 11) with mixed results. Seems I had tinkered around with Zadig before and I had to revert to the original drivers first.

After attaching the USB cable, it (always?) gets stuck like this
```
Serial port: com3
Connecting...

test
Reset done
```

When I do a board-info first it often is able to do the flashing. It doesn't reset the target however after flashing so I need to manually push the reset button on the dev-board.

After one or more successful flash attempts it ends up in the `Reset done` - hanging like above (there should be a ctrl-c handler since there is no way to get out of that other than pulling the USB cable currently)

jessebraham:
Thanks for testing it out. Very strange indeed. I guess this will require some more digging!

JurajSadel:
It looks like it has different behavior on different machines. I was able to flash on Windows 10 with pretty high success even after repeating flashing ~10 times. However, I encountered the same behavior @bjoernQ  described with `monitor`. The board was hung in download mode until I manually pressed the reset button and I couldn't get out of that download mode by changing states with `write_request_to_send()`- it doesn't change the voltage level on the physical pin for me.

MabezDev (approved):
> write_request_to_send()- it doesn't change the voltage level on the physical pin for me.

Seems this is an issue with serial-port-rs on windows (similar behavior already reported in https://github.com/esp-rs/espflash/issues/74)? Works well on x86 linux & mac.

Overall, I think this is probably good to go, the issues seem to be independent to this PR.

jessebraham:
Yeah, I generally agree with that assessment. It seems to be working _most_ of the time as-is from what I can tell, which is a good start at the very least. I'm only really running macOS these days, so makes sense that I wouldn't see the same behaviour described above.

PR #125, opened 2022-01-26
Update `clap` to stable version, reset baud when starting serial monitor

jessebraham:
- Move off of the `clap` release candidate to the stable release
- Reset the serial port's baud to 115,200 prior to starting the serial monitor, allowing the use of `--speed` and `--monitor` together

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,9 +205,9 @@ dependencies = [
 
 [[package]]
 name = "cargo_toml"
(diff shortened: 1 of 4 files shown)


MabezDev (approved):
> Reset the serial port's baud to 115,200 prior to starting the serial monitor, allowing the use of --speed and --monitor together

Ran into this so many times! Thanks for fixing!

PR #132, opened 2022-02-10
Correct the handling or partition table flags

jessebraham:
(This is for an upcoming PR adding support for flash encryption)

The `encrypted` flag should be a symbol, not an integer. See:  
https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/security/flash-encryption.html#encrypted-partition-flag

This adds the `Flags` type (of which there is currently only one, `encrypted`) and updates the relevant code. Also performed some simplification, cleanup, and formatting while I was in here.


bjoernQ (approved):
LGTM

PR #133, opened 2022-02-10
Refactor the command-line argument handling

jessebraham:
There are two major things accomplished by this PR:

1) Some CLI-related types were declared in `espflash` which were only used in `cargo-espflash`, so these have been moved
2) The subcommand actions have been refactored to be simpler and to duplicate less code

On top of this, I performed some normalization of type/variable names, just to make things more consistent throughout. I also changed some of the command-line arguments' types from `String` to `PathBuf`, where applicable.


bjoernQ:
Nice improvements code-wise, however the behavior changed a bit

* running `espflash` w/o args previously just printed the help but now it enumerates the ports and let me choose one (if more than one) and connects to it before printing help, if there is no serial port found it gives an error

* passing a serial port on the command line gives an error

```
❯ cargo run --package espflash -- com12 \projects\esp\esp-hal\target\xtensa-esp32s2-none-elf\debug\examples\timer_interrup
t
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target\debug\espflash.exe com12 \projects\esp\esp-hal\target\xtensa-esp32s2-none-elf\debug\examples\timer_interrupt`
Error: cargo_espflash::no_serial

  × No serial port specified in arguments or config
  help: Add a command line option with the serial port to use
```

bjoernQ:
> passing a serial port on the command line gives an error

mhhh - seems that wasn't introduced with these changes - looks like the same happens on `master`

jessebraham:
Great, thanks for the input! I'm just going to convert this to a draft until I have some time to come back to it (which will hopefully be soon).

jessebraham:
I believe I have addressed the issues which were raised in the most recent commit.

bjoernQ (approved):
LGTM - all my tests were successful

PR #134, opened 2022-02-17
Add a check to the monitor for interrupted system calls

### LINKED ISSUE #128: [Linux] Interrupted system call
When leaving the monitor open for a while (and sometimes when flashing) on Linux I see this error:

```
Error: 
  × Interrupted system call
```

I quick google tells me this "interrupted" signal can just be retried, but I'm not sure whether this needs to happen inside espflash or the underlying serial library.

jessebraham:
Fixes #128

@MabezDev can you please confirm (I have only tested on macOS, but was originally able to reproduce the error).


MabezDev (approved):
Works great! Thank you!

PR #136, opened 2022-02-24
Support the riscv32imac target, which can be used via atomic emulation

jessebraham:
We're able to perform atomic emulation for the C3:  
https://github.com/esp-rs/riscv-atomic-emulation-trap

However trying to use the `riscv32imac-unknown-none-elf` target results in an error saying it's unsupported.


MabezDev (approved):
Looks great! Thank you!!

PR #139, opened 2022-03-09
Re-work the serial error types to make the problem more clear

jessebraham:
Previously there was a single error variant for serial related problems, `NoSerial`, which did not necessarily describe the problem. This PR has re-worked the serial related error handling to hopefully better describe what went wrong. While I was at it I did some cleanup which (IMO) makes parts of this code a bit more readable/understandable.

Is this new approach adequate? Are there any other cases which may require an additional error variant?


bjoernQ (approved):
Much more user-friendly error messages now. Also, the other code changes make sense to me

Issue #140, opened 2022-03-10
Allow customization of firmware image headers


MabezDev:
Currently every firmware image we create with `from_elf` has hard coded flash attributes

https://github.com/esp-rs/espflash/blob/eeb0e8f95d184209cc42be9c4fec15c7172729fc/espflash/src/elf.rs#L49-L57

We should add CLI options to configure these, for the cases where the flash size of a chip may be different (see: https://github.com/esp-rs/esp-idf-template/issues/15)

jessebraham:
Should hopefully have a PR ready for this tomorrow. I managed to implement most of this today, just need to clean up some rough edges and do a bunch of testing still.

PR #141, opened 2022-03-11
Customize firmware image header

MabezDev:
This PR now updates the boot loader header _and_ the app image header only when a user supplies override via the `-s`, `-m` and `-f` flags. 

If no overrides are provided, the header is extracted from the esp-idf bootloader, and the same settings are used for the app header, by doing this we allow esp-idf SDKCONFIG settings to work seamlessly with espflash.

Closes https://github.com/esp-rs/espflash/issues/140

### DIFF
--- a/espflash/src/image_format/esp32bootloader.rs
+++ b/espflash/src/image_format/esp32bootloader.rs
@@ -1,6 +1,6 @@
 use std::{borrow::Cow, io::Write, iter::once};
 
-use bytemuck::{bytes_of, Pod, Zeroable};
+use bytemuck::{bytes_of, Pod, Zeroable, from_bytes};
 use sha2::{Digest, Sha256};
 
 use crate::{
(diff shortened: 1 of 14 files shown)


espflash/src/image_format/esp32bootloader.rs
```diff
@@ -33,21 +33,49 @@ impl<'a> Esp32BootloaderFormat<'a> {
...
-            flash_config: encode_flash_size(image.flash_size)? + image.flash_frequency as u8,
-            entry: image.entry,
-        };
+        // fetch the generated header from the bootloader
+        let mut header: EspCommonHeader = *from_bytes(&bootloader[0..8]);
+        if header.magic != ESP_MAGIC {
+            return Err(Error::InvalidBootloader);
+        }
+        // update the header if a user has specified any custom arguments
+        if image.flash_frequency.is_some()
```
bjoernQ, line 50: nitpicking but the outer if seems redundant

bjoernQ (approved):
looks good to me and seems to work fine

jessebraham:
Looks good overall, thanks for figuring out that final bit for me!

`--flash-size` and `--flash-freq` seem to be working as intended, however `--flash-mode` does not. I'm not sure if I'm missing something here or if there is an actual problem, so I will do some digging.

jessebraham (approved):
Okay, `--flash-mode` seems to work; I suppose there's a gap in my understanding here as it only works for some chips, but I assume this just comes down to the capabilities of the specific flash chips.

Will merge for now, if there are any issues we can tackle them in another PR.

PR #151, opened 2022-03-27
Scale the app partition based on flash size.

maspetsberger:
The pre-defined `app_sizes` for the various `Esp32Param` assume a flash size which is not necessarily correct, e.g. there are esp32-c3 chips with 2MB. 

This pull request removes the need to supply a custom partition-table for those chips.


jessebraham:
Good catch, thanks!

If you wouldn't mind fixing the formatting issues (sorry, I know it's a bit of a hassle) I'm happy to merge this.

jessebraham (approved):
Thanks for getting that fixed up!

PR #160, opened 2022-04-23
Refactor merge option a bit

### LINKED ISSUE #159: Running --merge for any path but the current local dir results in file not found
Easy to reproduce.
This works:
```
cargo espflash save-image --merge working
```
This doesn't:
```
cargo espflash save-image --merge /tmp/broken
Error: 
  × No such file or directory (os error 2)
```

svenstaro:
I think the earlier way the merge option was implemented was a bit surprising to the user and entirely undocumented in that. It created a new file `_merged` instead of the actual given path name. I think if a user provides `--merge` with a `file`, that `file` being the merged file is the expected outcome.

This refactors the merge functionality while providing a clear separation in terms of code paths for merged and unmerged paths while being overall less complex and less code to maintain.

This also fixes #159.

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -97,7 +97,7 @@ pub fn board_info(opts: ConnectOpts, config: Config) -> Result<()> {
 pub fn save_elf_as_image(
     chip: Chip,
     elf_data: &[u8],
-    path: PathBuf,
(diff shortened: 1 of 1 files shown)


jessebraham (approved):
Thank you for this! I agree this is more in line with how I would expect it to work.

Issue #149, opened 2022-03-19
Error: espflash::connection_failed


pengkebao:
pengkebao@localhost esp-rust-test % espflash /dev/tty.usbserial-0001 target/riscv32imc-esp-esp                      
Serial port: /dev/tty.usbserial-0001
Connecting...

Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...
Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...
Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...
Unable to connect, retrying with extra delay...
Error: espflash::connection_failed

  × Error while connecting to device
  ╰─▶ Failed to connect to the device
  help: Ensure that the device is connected and the reset and boot pins are
        not being held down

pengkebao@localhost esp-rust-test % espflash --version                         
espflash 1.3.0

pengkebao:
pengkebao@localhost esp-rust-test % espflash /dev/tty.usbserial-0001 target/riscv32imc-esp-esp
Serial port: /dev/tty.usbserial-0001
Connecting...

Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...
Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...
Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...
Unable to connect, retrying with extra delay...
Error: espflash::connection_failed

  × Error while connecting to device
  ╰─▶ Failed to connect to the device
  help: Ensure that the device is connected and the reset and boot pins are
        not being held down

pengkebao@localhost esp-rust-test % espflash --version                         
espflash 1.4.0-dev

jessebraham:
Sorry to hear you're having issues. Would you mind please sharing which board you are trying to flash? I have seen this issue with the ESP32 before but not with the ESP32-C3, so I'm interested where the problem lies.

LeDominik:
I have exactly that problem... Running on Intel, MacOS Monterey, board is a [FeatherS2 Neo](https://unexpectedmaker.com/feathers2-neo) so it's using the ESP32S2 _onboard_ USB capabilities. The device shows up as follows 

<img width="275" alt="image" src="https://user-images.githubusercontent.com/213234/164017547-efc8df36-baf1-4095-88d6-02db48ea07ab.png">

[...]

PR #164, opened 2022-04-26
fix bug #149

niceskylei:
I have same issue, fixed by add a line in  espflash/src/connection.rs:93 add sleep 10ms.
In my MacBook Pro 12.3.1, It's test OK.


jessebraham (approved):
Thanks for this! Glad it was a simple fix :)

PR #169, opened 2022-04-26
Hack to disable the rtc watchdog when connecting under the usb-serial-jtag

### LINKED ISSUE #154: Broken pipe error on ESP32-C3 for "large" applications
It seems that on my ESP32-C3 (The QT PY board from adafruit), after having flashed an application in bootloader mode, flashing fails for "large" applications (> ~200 chunks).

All these situations lead to a successful flash:
* Reset in bootloader mode (hold BOOT, press RESET), then `cargo espflash`
* First flash an application from the Arduino IDE (I suppose it includes its own bootloader), then try `cargo espflash`
* Flash a "small" application (less than ~200 chunks)

Here is an example of flashing a "larger application" (~500 chunks), right after having flashed from Arduino IDE:
```
Serial port: /dev/ttyACM0
Connecting...

    Finished release [optimized] target(s) in 0.05s

Chip type:         ESP32-C3 (revision 3)
Crystal frequency: 40MHz
Flash size:        4MB
Features:          WiFi
MAC address:       10:91:a8:31:06:e4
[00:00:00] ########################################      12/12      segment 0x0
[00:00:00] ########################################       1/1       segment 0x8000
[00:00:08] ########################################     489/489     segment 0x10000

Flashing has completed!
```

Then, running `cargo espflash` right after gives this output:

```
Serial port: /dev/ttyACM0
Connecting...

    Finished release [optimized] target(s) in 0.05s

Chip type:         ESP32-C3 (revision 3)
Crystal frequency: 40MHz
Flash size:        4MB
Features:          WiFi
MAC address:       10:91:a8:31:06:e4
[00:00:00] ########################################      12/12      segment 0x0
[00:00:00] ########################################       1/1       segment 0x8000
Error: espflash::serial_error

  × Communication error while flashing device
  ├─▶ IO error while using serial port: Broken pipe
  ╰─▶ Broken pipe
```
(Before the error message, I see chunks being uploaded; it fails around chunk 200.)

So I suspect the bootloader included in here is not quite the same as the one Arduino IDE includes; in particular, it doesn't seem to support well applications larger than 200 chunks.

MabezDev:
When the chip is reset via the USB-SERIAL-JTAG not everything is fully reset, the RTC domain stays active. Therefore, if the RTC watchdog is enabled before being put into download mode it will trigger when flashing, hence resetting the board and interrupting flashing.

The reason it always works on Arduino is that the RTC watchdog is explicitly disabled very early on in the boot process which doesn't seem to be the case for normal esp-idf applications, and therefore Rust applications. 

Closes #154

### DIFF
--- a/espflash/src/connection.rs
+++ b/espflash/src/connection.rs
@@ -16,7 +16,7 @@ use crate::{
 };
 
 const DEFAULT_CONNECT_ATTEMPTS: usize = 7;
-const USB_SERIAL_JTAG_PID: u16 = 0x1001;
+pub const USB_SERIAL_JTAG_PID: u16 = 0x1001;
 
 #[derive(Debug, Copy, Clone, BinRead)]
 pub struct CommandResponse {
--- a/espflash/src/flash_target/esp32.rs
+++ b/espflash/src/flash_target/esp32.rs
@@ -1,5 +1,5 @@
 use crate::command::{Command, CommandType};
-use crate::connection::Connection;
+use crate::connection::{Connection, USB_SERIAL_JTAG_PID};
 use crate::elf::{FirmwareImage, RomSegment};
 use crate::error::Error;
 use crate::flash_target::FlashTarget;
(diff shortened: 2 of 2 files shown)


jessebraham (approved):
LGTM, thanks!

PR #172, opened 2022-04-30
Add partition-table subcommand

### LINKED ISSUE #163: Expose partition table csv->bin and bin->csv parsers as subcommand
This tool is quite handy in that it does the partition table juggling for you. However, it only does it as part of its internal process. What do you think about exposing that stuff explicitly as a sub-command so that we may use in in place of [the esp-idf way](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html#generating-binary-partition-table)?

### LINKED ISSUE #165: Add image info subcommand
It would be cool to have a subcommand to pretty print (with optional machine-readable output) the layout of the image (according to the partition table) as well as the actual app size within. This should be possible with either a connected board or a provided local image file.

svenstaro:
Fixes #165 and fixes #163. It's quite pretty too:
image

This is quite a bit more than I bargained for but I think this makes sense as parsing of the binary stuff is also in esp-idf's `gen_esp32part.py` and that's Python so we can't have that!

I had to bump the MSRV to 1.58 but I hope that's ok.

### DIFF
--- a/espflash/src/main.rs
+++ b/espflash/src/main.rs
@@ -1,4 +1,4 @@
-use std::{fs, mem::swap, path::PathBuf, str::FromStr};
+use std::{fs, io::Write, mem::swap, path::PathBuf, str::FromStr};
 
 use clap::{IntoApp, Parser};
 use espflash::{
@@ -6,7 +6,7 @@ use espflash::{
         board_info, connect, flash_elf_image, monitor::monitor, save_elf_as_image, ConnectOpts,
         FlashConfigOpts, FlashOpts,
     },
-    Chip, Config, ImageFormatId,
+    Chip, Config, ImageFormatId, InvalidPartitionTable, PartitionTable,
 };
 use miette::{IntoDiagnostic, Result, WrapErr};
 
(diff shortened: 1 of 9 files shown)


espflash/src/main.rs
```diff
@@ -60,13 +62,34 @@ pub struct SaveImageOpts {
...
+pub struct PartitionTableOpts {
+    /// Convert CSV parition table to binary representation
+    #[clap(long, required_unless_present_any = ["info", "to-csv"])]
+    to_binary: bool,
+    /// Convert binary partition table to CSV representation
+    #[clap(long, required_unless_present_any = ["info", "to-binary"])]
+    to_csv: bool,
+    /// Show information on partition table
+    #[clap(short, long, required_unless_present_any = ["to-binary", "to-csv"])]
+    info: bool,
```
icewind1991, line 75: it might make more sense to have these as sub commands

svenstaro: I think it's fine as they terminate the chain of commands and I don't like too deeply nested subcommands.

jessebraham (changes requested):
Thank you so much for this, sorry again for the delays!

Overall I think this looks quite good, I don't have much else to say about the submitted code which isn't nitpicking. There are a couple things I'd like to see added though, if you are willing; I can do it myself if you are not feeling up to it however, please just let me know.

1. Could you please update the top-level README to include the new MSRV?
2. You have updated `espflash/src/main.rs`, but not `cargo-espflash/src/main.rs`; would it be possible for you to add this functionality to `cargo-espflash` as well?

svenstaro:
Cool, all done!

jessebraham (approved):
Awesome, thank you so much!

PR #174, opened 2022-05-08
Update README.md

andywswan:
Typo, getting "Error: × invalid type: map, expected a sequence for key `usb_device` at line 4 column 1" otherwise


MabezDev:
Thanks!

PR #179, opened 2022-05-17
bugfix: flash with wsl2

### LINKED ISSUE #166: espflash 1.3.0+ doesn't not work under WSL with usbipd
As the title suggests, commits post 1.2.0 break compatibility with WSL and usbipd.
Connecting to the device over uart with the usual Python IDF utilities works fine.

Narrowing down the issue, it appears it started with commit https://github.com/esp-rs/espflash/commit/85c8eb84db7f0a534d73772069f864475ff4dfe6 .

The error shows as below:
```

  × The serial port '/dev/ttyUSB0' could not be found

```

However the device is definitely connected and is listed under `/dev/`, as well as works with the Python IDF utilities.

espflash 1.4.1 works fine under a Linux host, and on Windows itself as expected.

JurajSadel:
Closes #166 

Note: It is not possible to flash `usb-serial-jtag` chips with `WSL2` because the reset also resets `serial-jtag-peripheral` which disconnects the chip from WSL2.

### DIFF
--- a/cargo-espflash/README.md
+++ b/cargo-espflash/README.md
@@ -104,6 +104,9 @@ vid = "303A"
 pid = "8000"
 ```
 
+## WSL2
+It is not possible to flash `usb-serial-jtag` chips with `WSL2` because the reset also resets `serial-jtag-peripheral` which disconnects the chip from WSL2.
+
 ## Package Metadata
 
 You can specify the bootloader, partition table, or image format for a project in the package metadata in `Cargo.toml`:
--- a/espflash/README.md
+++ b/espflash/README.md
@@ -87,6 +87,9 @@ vid = 12346 # 0x303A
 pid = 32768 # 0x8000
 ```
 
+## WSL2
+It is not possible to flash `usb-serial-jtag` chips with `WSL2` because the reset also resets `serial-jtag-peripheral` which disconnects the chip from WSL2.
+
 ## Use as a Cargo Runner
 
 You can also use `espflash` as a Cargo runner by adding the followin to your project's `.cargo/config` file:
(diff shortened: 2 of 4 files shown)


andywswan:
Definitely works, thank you so much!

```
            SerialPortType::Unknown => &UsbPortInfo {
                vid: 0,
                pid: 0,
                serial_number: None,
                manufacturer: None,
                product: None,
            }
```

I thought VID and PID were necessary for the crate to work, but this simplifies everything.

MabezDev:
> Definitely works, thank you so much!
> 
> ```
>             SerialPortType::Unknown => &UsbPortInfo {
>                 vid: 0,
>                 pid: 0,
>                 serial_number: None,
>                 manufacturer: None,
>                 product: None,
>             }
> ```
> 
> I thought VID and PID were necessary for the crate to work, but this simplifies everything.

It is required for flashing via [USB-SERIAL-JTAG](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-guides/usb-serial-jtag-console.html), as it has a different reset routine, but other USB serial chips should just work :). In the future, we might want to add a CLI flag to choose which reset routine to use.

jessebraham:
@JurajSadel this looks good, thank you for pursuing this issue!

Would you mind also adding notes to both `cargo-espflash/README.md` and `espflash/README.md` mentions that USB Serial JTAG will not work in WSL please? Just a header and a single sentence is probably adequate, I just want it to be clearly noted somewhere.

andywwright:
I like the commit as it is, it works and this is all I needed, but you might want to add some tiny warning like "The type of the tty port is unknown". I think it would really help reasoning with some future issues, especially with [USB-SERIAL-JTAG](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-guides/usb-serial-jtag-console.html). Left as it is you'll never know if you have real vid and pid detected or they are silent fake zeros.

jessebraham (approved):
Thanks for making those changes!

PR #180, opened 2022-05-18
Add --locked and --frozen options

svenstaro:
These are easy to support but important to ensure reproducibility and as such are a must for production deployments of modern firmware. :)


MabezDev:
Thanks!

PR #182, opened 2022-05-19
Make all library modules public

MabezDev:
I'm trying to use espflash as library and needed to open up espflash a lil bit :D. Is there any reason not to do this? 

Also snuck in a trait bound on `ImageFormat` to make `dyn ImageFormat` `Send`able across threads, happy to open another PR if that's more suitable!

Edit: If this seems like a reasonable idea lmk and I'll fix the clippy issues.

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -112,6 +112,7 @@ pub fn board_info(opts: ConnectOpts, config: Config) -> Result<()> {
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
 pub fn save_elf_as_image(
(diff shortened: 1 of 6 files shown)


jessebraham:
Yeah I can't think of any reason not to do this, fine with me. `cargo-espflash` is the only listed dependent on [crates.io](https://crates.io) anyways 😅

MabezDev:
Didn't really know what to do with `clippy::too_many_arguments`, so I just allowed it :shrug:.

jessebraham (approved):
All good, I'm not a fan of that rule anyways. Thanks for getting that fixed up!

PR #183, opened 2022-05-20
remove hyphen from chip string

MabezDev:
esptool uses the chip name _without_ the hypen, e.g esp32c3 instead of esp32-c3. I see no reason not to do the same here, typing a hyphen is annoying :D.

### DIFF
--- a/espflash/src/chip/mod.rs
+++ b/espflash/src/chip/mod.rs
@@ -141,11 +141,16 @@ impl FromStr for Chip {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s.to_ascii_lowercase().as_str() {
+        let s: String = s
+            .chars()
+            .filter(|&c| c != '-')
+            .map(|c| c.to_ascii_lowercase())
+            .collect();
+        match s.as_str() {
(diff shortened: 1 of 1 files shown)


bjoernQ (approved):
👍  I always tried to use the non-hyphen version just to run into an error

In theory we could also support both? No idea why we should
Only reason that comes to my mind is if someone uses this in a script which will fail after the change

igrr:
Although we can't (and shouldn't) force everyone to do the same, we generally are trying to stick to two ways of writing chip names:

* upper case, with hyphens (the actual product name, like ESP32-C3) — in the docs, error messages, and so on
* lower case, without hyphens (like esp32c3) — when we need an identifier, such as in function or variable names, file names, package names.

Other options we try to avoid: ESP32C3, esp32-c3, esp32_c3. Of course, in informal discussions anything works (like "c3").

That said, some of our tools (like `idf.py set-target`) accept all of the above options by lowercasing the input and removing hyphens in underscores.

MabezDev:
> That said, some of our tools (like idf.py set-target) accept all of the above options by lowercasing the input and removing hyphens in underscores.

That's a good idea! Just pushed some code that will do just that.

PR #184, opened 2022-05-20
CI Add workflow for building artifatcs for armv7 and aarch64 linux

### LINKED ISSUE #170: Build release artifacts for the Raspberry Pi

JurajSadel:
closes #170 

This PR builds artifacts for cargo-espflash and espflash for Linux armv7 and armv7 (aarch68).
I wasn't successful with [cross](https://github.com/cross-rs/cross).


jessebraham (changes requested):
Thanks for this, sorry in the delay for reviewing!

I did notice that unfortunately both the `espflash-aarch64-unknown-linux-gnu.zip` and `espflash-armv7-unknown-linux-gnueabihf.zip` files seem to be missing the binaries (they're only 22 bytes, which indicates an empty archive). If you're able to resolve this I'm otherwise pretty happy with this.

JurajSadel:
That makes sense, I zipped `espflash` before build :D Should be fixed, thank you @jessebraham

jessebraham (approved):
Thanks for fixing that!

PR #186, opened 2022-05-26
Detect the bootloader and partition table from ESP-IDF when able

### LINKED ISSUE #185: Detect and use bootloader/partition table from ESP-IDF when able
When users are building projects via `cargo-espflash` and are using `esp-idf-sys`, it's possible to locate the `bootloader.bin` and `partition_table.bin` files and use them automatically. We should do this while still allowing for users to provide their own via cargo package metadata and/or command-line options.

jessebraham:
Closes #185 

When the `esp-idf-sys` package is built, we store the paths to the already-compiled `bootloader.bin` and `partition-table.bin` files. If no alternatives have been specified via either command-line arguments and/or Cargo package metadata, then these binaries will be used instead of the defaults included with `espflash`.

When users are _not_ building `esp-idf-sys`, (ie. bare-metal) then the behaviour does not change.

@MabezDev @bjoernQ @ivmarkov @n3xed will this cause any issues which I have not thought of?

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -132,6 +132,13 @@ fn main() -> Result<()> {
     }
 }
 
+#[derive(Debug, Clone)]
+struct BuildContext {
+    pub artifact_path: PathBuf,
+    pub bootloader_path: Option<PathBuf>,
+    pub partition_table_path: Option<PathBuf>,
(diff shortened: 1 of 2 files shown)


MabezDev (approved):
Looks good to me! Thanks for looking into this!

PR #188, opened 2022-05-28
Introduce FirmwareImage trait

tpambor:
This PR makes `FirmwareImage` a trait and moves the original implementation to `ElfFirmwareImage` and `ElfFirmwareImageBuilder` so that formats other than ELF could be supported.


svenstaro:
While this is probably a better architecture, is there _currently_ a use for a non-ELF image?

tpambor:
Beside ELF there is currently only the binary [App Image Format](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/app_image_format.html).

jessebraham:
Thanks for this, looks fine to me. If you're able to rebase I don't have a problem merging.

Just as a note, we are already using the App Image Format, which is what the `EspCommonHeader` and `EspExtendedHeader` structs are for. So I'm not really sure that additional formats are much of a concern. With that said, I still think this is probably architecturally better than what we have currently.

tpambor:
@jessebraham I rebased it

jessebraham (approved):
Thanks!

PR #189, opened 2022-05-30
Remove dependency to FirmwareImage from flash_targets

tpambor:
I am looking into using espflash as a API to flash single a binary like the phy init data. Currently, the flash_targets depend on a `FirmwareImage` which requires to have loaded a valid ELF file.

The `FlashTarget` is currently only used to get the entrypoint for the `RamTarget`. This PR makes the entrypoint specific to the `RamTarget` and makes the `FlashTarget` trait more generic for other use-cases.

### DIFF
--- a/espflash/src/chip/mod.rs
+++ b/espflash/src/chip/mod.rs
@@ -254,8 +254,8 @@ impl Chip {
         }
     }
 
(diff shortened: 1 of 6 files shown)


jessebraham (approved):
LGTM, thanks!

Issue #171, opened 2022-04-29
Broken pipe while flashing over USB, on macOS only


allison-n-h:
Trying to test flashing an image using the USB port on an ESP32S3 DevKit-C-1-N8. For larger images (like the [wifi station example](https://github.com/espressif/esp-idf/tree/v4.4/examples/wifi/getting_started/station) from the idf), flashing fails about ~75% through the image:

```
Serial port: /dev/tty.usbmodem14101
Connecting...

Chip type:         ESP32-S3
Crystal frequency: 40MHz
Flash size:        8MB
Features:          WiFi, BLE
MAC address:       7c:df:a1:e2:4a:a8
[00:00:00] ########################################      13/13      segment 0x0
[00:00:00] ########################################       1/1       segment 0x8000
Error: espflash::serial_error

  × Communication error while flashing device
  ├─▶ IO error while using serial port: Broken pipe
  ╰─▶ Broken pipe
```

This seems to only be an issue with espflash on macOS and using the USB port on the devkit. Using the UART port works as expected, and using the native `idf.py flash` works as expected. It also seems to occur only 'larger' images - for example, flashing the hello_world example from the IDF completes successfully. Testing this on a Windows 11 machine, we don't see this issue either, so it does seem to be macOS specific.

We did verify we are using the 1.4.1 version of espflash.

MabezDev:
Can you try using master? We recently merged #169 which should fix this issue.

allison-n-h:
I seem to get a different failure when using the latest from master:

```
Connecting...

Chip type:         ESP32-S3
Crystal frequency: 40MHz
Flash size:        8MB
Features:          WiFi, BLE
MAC address:       7c:df:a1:e0:31:3c
Error: espflash::timeout

  × Communication error while flashing device
  ╰─▶ Timeout while running WriteReg command


```

jessebraham:
I've unfortunately never seen this error in practice, so I'm a little stumped as to what the issue is. I run macOS as well and have never run into issues flashing the ESP32-S3, although I haven't used it much.

Just out of curiosity, which version of macOS are you running? Also if you're able to share your project (either a repo or a zip file, or whatever is easiest) that would be helpful so I could at least try to reproduce this.

allison-n-h:
We were seeing this on macOS Monterey 12.1, and trying to flash the app from this esp-idf example: https://github.com/espressif/esp-idf/tree/v4.4/examples/wifi/getting_started/station

[...]

Issue #126, opened 2022-01-28
Flashing not working


luukvankooten:
```zsh
$ espflash /dev/cu.usbserial-0001 target/xtensa-esp32-espidf/release/led
Serial port: /dev/cu.usbserial-0001
Connecting...

Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...
Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...
Unable to connect, retrying with extra delay...
Unable to connect, retrying with default delay...

Error: espflash::timeout

  × Error while connecting to device
  ╰─▶ Timeout while running ReadReg command
```

Every single time I run espflash this output is sent to me. It looks like it's doing something because my previous program which is flashed stops. Looked into the issues, it is like #85 but not sure. I'm also on Mac M1 don't know if it has todo something with that. And other programs are successfully uploaded with the Arduino IDE.

luukvankooten:
Got a workaround: install espflash with this specific comment: `cargo install espflash --git https://github.com/esp-rs/espflash.git --rev fd205c7d21cf7eb339ccb76d270d1821577c0254 --force`

pyaillet:
I faced the same problem. Reverting to v1.1.0 solved it for me.

I can provide

jessebraham:
Would either or both of you be able to check whether this issue persists using the newest version (`1.4.0`) please?

pyaillet:
I switched to 1.4.0-dev last week to test panic backtrace of `espmonitor` and did not face any problems since then.
I will try with the `1.4.0` release though.

pyaillet:
So I just installed the released `1.4.0` version and it works perfectly on my side.

Thank you for this update :)

PR #190, opened 2022-06-01
Include the base address of the RTC peripheral when disabling the RTC WDT

MabezDev:
Closes [#126 ](https://github.com/esp-rs/espflash/issues/171)

Interestingly even though flashing now succeeds, the application doesn't seem to start - I'm not sure if this is just my devkit though.


jessebraham:
Why were certain boards previously working if we were writing to invalid addresses? The linked issue seems to only effect certain boards, I'm not really sure what's going on here.

jessebraham (approved):
And with all that said, the person reporting the issue has confirmed this resolved it so good enough for me.

PR #194, opened 2022-06-10
Add "write-bin-to-flash" subcommand

rdeterre:
This subcommand adds support for writing a binary file at a specific address in flash.


jessebraham (approved):
Seems to be working as intended for me, thanks for this!

PR #202, opened 2022-07-04
Allow hex or decimal addresses to be passed to the write bin subcommand

### LINKED ISSUE #201: The `write-bin-to-flash` subcommand should accept hex or decimal addresses

jessebraham:
Closes #201

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,7 +34,7 @@ dependencies = [
  "gimli",
  "object 0.27.1",
  "rustc-demangle",
- "smallvec 1.8.0",
+ "smallvec 1.9.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


bjoernQ:
Looks good / will test that tomrorrow but just FYI the probe-rs project has an interesting solution for this:

https://github.com/probe-rs/probe-rs/blob/67a180fcb6121b60b2dd27930d05cc6cc88ca669/cli/src/main.rs#L96
https://github.com/probe-rs/probe-rs/blob/67a180fcb6121b60b2dd27930d05cc6cc88ca669/cli/src/main.rs#L490

They use https://crates.io/crates/parse_int and a custom CLAP parse function. Didn't knew that is possible but looks like a good alternative

jessebraham:
Oh interesting, wasn't aware of that approach. Thanks! I'll probably re-work it to go that route instead.

bjoernQ (approved):
LGTM

PR #207, opened 2022-07-19
Fix direct-boot image check for ESP32-S3

bjoernQ:
Unlike the ESP32-C3 the magic bytes to identify a direct-boot image is at offset 0x400.

### DIFF
--- a/espflash/src/image_format/esp32directboot.rs
+++ b/espflash/src/image_format/esp32directboot.rs
@@ -12,7 +12,7 @@ pub struct Esp32DirectBootFormat<'a> {
 }
 
 impl<'a> Esp32DirectBootFormat<'a> {
-    pub fn new(image: &'a dyn FirmwareImage<'a>) -> Result<Self, Error> {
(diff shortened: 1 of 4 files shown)


espflash/src/image_format/esp32directboot.rs
```diff
@@ -12,7 +12,7 @@ pub struct Esp32DirectBootFormat<'a> {
 }
 
 impl<'a> Esp32DirectBootFormat<'a> {
-    pub fn new(image: &'a dyn FirmwareImage<'a>) -> Result<Self, Error> {
+    pub fn new(image: &'a dyn FirmwareImage<'a>, magic_position: usize) -> Result<Self, Error> {
```
MabezDev, line 15: minor nitpick I would maybe rename this to offset

MabezDev (approved):
LGTM! Just a small nitpick that you can leave if you like :)

jessebraham (changes requested):
Looks good overall, thanks for the fix! I agree with @MabezDev, let's change `magic_position` to `offset` and then I think it's good to go.

jessebraham (approved):
Thanks for updating that!

PR #212, opened 2022-07-27
Consistently support binary partition tables

### LINKED ISSUE #211: Failed to open partition table when using save-image with esp-idf custom partition table
My setup is like this:
sdkconfig.defaults
```
CONFIG_PARTITION_TABLE_CUSTOM=y
CONFIG_PARTITION_TABLE_CUSTOM_FILENAME="custom_partitions.csv"
CONFIG_PARTITION_TABLE_FILENAME="custom_partitions.csv"
```
I do:
```
cargo espflash save-image --merge ESP32-S3 lol.bin
```
I get:
```
Finished release [optimized] target(s) in 0.23s
Error: 
  × Failed to open partition table
  ╰─▶ stream did not contain valid UTF-8
```
I expect it to just work!

Instead to make it work I need to set this in my `Cargo.toml`:
```
[package.metadata.espflash]
partition_table = "custom_partitions.csv"
```
I think this should not be required. `espflash` can parse binary partition tables just fine and I think it should just use the binary partition table by esp-idf.

kyrias:
Fixes #211.

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -178,15 +178,15 @@ pub fn save_elf_as_image(
         };
 
         // If the '-T' option is provided, load the partition table from
-        // the CSV at the specified path.
+        // the CSV or binary file at the specified path.
         let partition_table = if let Some(partition_table_path) = partition_table_path {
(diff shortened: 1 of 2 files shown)


jessebraham (approved):
Great, thank you!

PR #214, opened 2022-08-03
Check if a new version if available upon invocation

jessebraham:
Checks [crates.io](https://crates.io) for a new version at the time of invocation, and prints a message to the user if an update is available.

You can test locally by hard-coding in a version below `1.6.0`.


MabezDev (approved):
Very nice! Thanks for looking into this!

Issue #177, opened 2022-05-17
espflash does not follow symlinks on Linux


alxpettit:
```
$ espflash -V
espflash 1.5.1
$ uname -a
Linux ada-DT0 5.13.0-41-generic #46~20.04.1-Ubuntu SMP Wed Apr 20 13:16:21 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

In modern versions of Linux, the devfs system includes symlinks that allow you to unambiguously reference items by ID -- this includes serial devices.

e.g.
```
$ ls /dev/serial/by-id/ -lah
total 0
drwxr-xr-x 2 root root 80 May 16 19:55 .
drwxr-xr-x 4 root root 80 May 16 16:58 ..
lrwxrwxrwx 1 root root 13 May 16 19:19 usb-Raspberry_Pi_Picoprobe_[REDACTED]_if00 -> ../../ttyACM0
lrwxrwxrwx 1 root root 13 May 16 19:55 usb-Silicon_Labs_CP2102_USB_to_UART_Bridge_Controller_0001-if00-port0 -> ../../ttyUSB0
```

Instead of following symlinks passed for serial device file paths, it returns an error, like so:

```
$ espflash /dev/serial/by-id/usb-Silicon_Labs_CP2102_USB_to_UART_Bridge_Controller_0001-if00-port0 target/xtensa-esp32-espidf/debug/rust-esp32-std-demo
Error: espflash::serial_not_found

  × The serial port '/dev/serial/by-id/usb-Silicon_Labs_CP2102_USB_to_UART_Bridge_Controller_0001-if00-port0' could not be found
  help: Make sure the correct device is connected to the host system
```

This leads me to use workarounds with `basename` and `readlink`, which definitely seems subpar and unnecessary. :P

MabezDev:
I've just opened #178, could you try it out and report back?

Seems to work for me

```rust
cargo run -p espflash -- --monitor /dev/serial/by-id/usb-Espressif_USB_JTAG_serial_debug_unit_84:F7:03:41:79:AC-if00 target/riscv32imc-esp-espidf/debug/esp-fs-tests
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/espflash --monitor '/dev/serial/by-id/usb-Espressif_USB_JTAG_serial_debug_unit_84:F7:03:41:79:AC-if00' target/riscv32imc-esp-espidf/debug/esp-fs-tests`
[...]

PR #218, opened 2022-08-15
bugfix: Don't use canonicalized strings on Windows

### LINKED ISSUE #209: "The parameter is incorrect" error when flashing from windows
Hi, 

I'm new to rust and embedded in general but I believe there is a regression in espflash 1.6.0 when running from windows. 

```plain
espflash COM3 .\target\xtensa-esp32-espidf\debug\test
Error: ←[31mespflash::serial_error←[0m

  ←[31mx←[0m Error while connecting to device
←[31m  |-> ←[0mIO error while using serial port: The parameter is incorrect. (os error 87)
←[31m  `-> ←[0mThe parameter is incorrect. (os error 87)
```

Running the same command on 1.5.1 is successful. My environment is Windows 10 (build 19044). Happy to provide any additional details if required.

JurajSadel:
closes #209

### DIFF
--- a/espflash/src/cli/serial.rs
+++ b/espflash/src/cli/serial.rs
@@ -19,12 +19,21 @@ pub fn get_serial_port_info(
     // select a serial port. If some VID and PID were provided then the user will
     // also be prompted to select a port, unless there is only one found and its VID
     // and PID match the configured values.
+    //
+    // The call to canonicalize() was originally added to resolve https://github.com/esp-rs/espflash/issues/177,
+    // however, canonicalize  doesn't work (on Windows) with "dummy" device paths like `COM4`.
+    // That's the reason we need to handle Windows/Posix differently.
+
     let ports = detect_usb_serial_ports().unwrap_or_default();
 
     if let Some(serial) = &matches.serial {
-        find_serial_port(&ports, &std::fs::canonicalize(serial)?.to_string_lossy())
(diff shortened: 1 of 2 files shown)


MabezDev:
Just passing by, I don't understand the full issue, but could it be the `to_string_lossy` that's the real issue? `find_serial_port` should _probably_ take a PathBuf?

MabezDev:
> Just passing by, I don't understand the full issue, but could it be the `to_string_lossy` that's the real issue? `find_serial_port` should _probably_ take a PathBuf, not a String?

We looked into this, it seems canonicalize doesn't work with "dummy" device paths like `COM4` for example, hence this error. Therefore this change is correct.

espflash/src/cli/serial.rs
```diff
@@ -22,9 +22,13 @@ pub fn get_serial_port_info(
     let ports = detect_usb_serial_ports().unwrap_or_default();
 
     if let Some(serial) = &matches.serial {
```
jessebraham, line 24: Would you mind adding a comment explaining this please? There are a couple things going on here and it isn't immediately clear to others what the motivation for this is:

- The call to `canonicalize()` was originally added to resolve #177
- This then causes issues on Windows, so we need to handle Windows/Posix differently

jessebraham (approved):

PR #220, opened 2022-08-18
Add support for custom partition types and sub-types

### LINKED ISSUE #219: Custom partition type
I would like to add a custom a partition for more or less raw runtime storage (NVS API is not applicable). The docs at https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html#creating-custom-tables recommend that application-defined partitions use the unreserved type values 0x40-0xFE.

Trying to flash a partition table like this:
```
# ESP-IDF Partition Table
# Name,   Type, SubType, Offset,  Size, Flags
nvs,data,nvs,0x9000,0x6000,
phy_init,data,phy,0xf000,0x1000,
factory,app,factory,0x10000,0x100000,
custom,0x40,0x00,0xf00000,0x100000,
```
gives an error
```
Error: espflash::partition_table::mallformed

  × Failed to parse partition table
  ├─▶ Malformed partition table
  ├─▶ CSV deserialize error: record 3 (line: 6, byte: 171): unknown variant `0x40`, expected `app` or `data`
  ╰─▶ unknown variant `0x40`, expected `app` or `data`
   ╭─[5:1]
 5 │ factory,app,factory,0x10000,0x100000,
 6 │ custom,0x40,0x00,0xf00000,0x100000,
   · ─────────────────┬─────────────────
   ·                  ╰── unknown variant `0x40`, expected `app` or `data`
   ╰────
  help: See the espressif documentation for information on the partition table format:
        
        https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html#creating-custom-tables

```
i.e. only the textual types "app" and "data" are currently supported. A similar error occurs when trying to provide a binary partition table.

I suppose for the time being I could use partition type "data" and subtype "undefined".

jessebraham:
Adds support for custom partition types and sub-types. Refer to the ESP-IDF partition table documentation for more information on these:  
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html

@ollpu would you be able to test out this PR and confirm it is working for you, please?

Closes #219

### DIFF
--- a/espflash/src/partition_table.rs
+++ b/espflash/src/partition_table.rs
@@ -3,7 +3,6 @@ use std::{
     fmt::{Display, Formatter, Write as _},
     io::{Cursor, Write},
     ops::Rem,
-    str::FromStr,
 };
 
 use binread::{BinRead, BinReaderExt};
@@ -12,7 +11,7 @@ use md5::{Context, Digest};
 use regex::Regex;
 use serde::{Deserialize, Deserializer, Serialize};
 use strum::IntoEnumIterator;
-use strum_macros::{EnumIter, EnumString};
+use strum_macros::EnumIter;
 
 use crate::error::{
     CSVError, DuplicatePartitionsError, InvalidChecksum, InvalidPartitionTable,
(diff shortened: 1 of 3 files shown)


espflash/src/partition_table.rs
```diff
@@ -56,58 +56,113 @@ impl Type {
...
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, PartialEq, Eq, BinRead, EnumString)]
 #[repr(u8)]
 #[br(little, repr = u8)]
 pub enum AppType {
     #[serde(rename = "factory")]
+    #[strum(serialize = "factory")]
     Factory = 0x00,
     #[serde(rename = "ota_0")]
+    #[strum(serialize = "ota_0")]
     Ota0 = 0x10,
```
jessebraham, line 108: I'm currently using `strum_macros::EnumString` to convert from a `String` to an `AppType` discriminant. I unfortunately cannot use `#[strum(serialize_all = "snake_case")]` here as we need to insert the underscores into the serialized names.

Is there any way I can perform this conversion using `serde` instead? I'm not very experienced with it.

ollpu: Something like this?

 ```rust
#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Copy, Clone, BinRead)]
#[serde(untagged)]
pub enum SubType {
    App(AppType),
    Data(DataType),
    #[serde(deserialize_with = "deserialize_custom_partition_sub_type")]
    Custom(u8),
}
```
`untagged` already handles that "first successful parse" logic for you.

**Comment by ollpu** [comment]
Seems to work for the most part, however, when serializing a partition table to CSV, custom partition types still use textual names. So

```
custom,0x40,0x00,0xf00000,0x100000,
```
gets turned into
```
custom,0x40,factory,0xf00000,0x100000,
```

I don't know if IDF still accepts this, but at the very least it's somewhat semantically wrong.

jessebraham:
Ahh yeah that doesn't look right, thanks. I'll add a test case for this as well.

jessebraham:
I have a fix for the issue locally, but I don't really like my solution so I'm going to spend a bit of time thinking on it. Converting to draft for now.

espflash/src/partition_table.rs
jessebraham: You're right, I was able to re-work and simplify this using this approach. Thanks!

jessebraham:
I believe at this point your custom partitions should now be handled correctly. I don't really like my fix for handling the custom sub-types, but I'm not really sure how else to go about accomplishing this without some more involved reworking, so I suppose it's fine for now.

MabezDev (approved):
Looks good to me :)

PR #223, opened 2022-08-30
Wait for FlashDeflateEnd ack

mertzt89:
When using the stub, the data sent in FlashDeflateData is not actually
written to flash until after the ack is sent by the stub. Waiting for
FlashDeflateEnd ensures that all data sent during with FlashDeflateData
has been fully written.


jessebraham (approved):
Thanks for this!

Just out of curiosity, what did this failure look like? Was there a specific error printed or was it just spurious flashing failures?

mertzt89:
> Thanks for this!
> 
> Just out of curiosity, what did this failure look like? Was there a specific error printed or was it just spurious flashing failures?

In my case the device would reset before the bootloader has fully written everything resulting in a section of 0xff's in the last segment programmed. 

In my case it was the tail end of my application binary that would end up missing which seems to contain important section information that the bootloader reads. So ultimately since my device would fail to boot.

PR #227, opened 2022-09-05
Refactor dependencies so that CLI dependencies are optional

Douile:
Add the "cli" feature that enables dependencies that are only used by
the CLI, this makes the library compilation smaller and easier to build
for some targets (such as android) as it eliminates TLS and such.

After this change to build the CLI you must run:
`cargo build --bin esptool --features cli`

`cargo build` with no arguments will only build the library.

# Notes
I attempted to split the CLI off into a separate workspace but it didn't work as intuitively as what is in this PR, both work it is mainly down to preference I believe.


jessebraham:
Sorry, I kind of forgot about this PR. I appreciate the contribution, but I've actually since merged my CLI rewrite (which had been in the works for awhile) which also does this. So again, thanks but this is already done!

(Note that the feature is enabled by default, however)

PR #228, opened 2022-09-05
Add support for flashing S2 via CDC UART

mertzt89:


mertzt89:
The bulk of the constants came from
https://github.com/espressif/esptool/blob/master/esptool/targets/esp32s2.py

Issue #224, opened 2022-08-31
RFC: Relicense espflash under Apache-2.0 and MIT license


tom-borcin:
Hello all,

firstly I want to emphesize that **this issue is an RFC**. Your comments are welcome.
I would like to open a discussion about the license we are using: GNU GENERAL PUBLIC LICENSE Version 2 (GPL-2.0).

### The problem

Espressif currently maintains 4 codebases for flashing of microcontrollers:

1. [Esptool](https://github.com/espressif/esptool) for flashing from OS
2. [esp-serial-flasher](https://github.com/espressif/esp-serial-flasher) for flashing from host microcontroller
3. [esptool-js](https://github.com/espressif/esptool-js) for flashing from web browser
4. [espflash](https://github.com/esp-rs/espflash) for flashing from OS

Maintaining 4 different codebases is difficult, costs a lot of resources, is prone to mistakes and defies DRY (Don't Repeat Yourself) principle. Our end goal is to have one common codebase that will be used by other three codebases. And we chose espflash as our common codebase.

Unfortunately, there is one problem that prevents us to use it as library whitin other projects: its [GPL-2.0 license](https://www.gnu.org/licenses/old-licenses/gpl-2.0.html).

[esp-serial-flasher](https://github.com/espressif/esp-serial-flasher) and [esptool-js](https://github.com/espressif/esptool-js) are both licensed under Apache 2.0, which is incompatible with GPL-2.0 due to the restriction that terminates the grant of patent rights if the license sues over patent infringement. You can read about it in more detail [here](https://www.apache.org/licenses/GPL-compatibility.html).

Any library licensed under GPL-2.0 will enforce this licensing on any program that uses it. You can read about it [here](https://www.gnu.org/licenses/old-licenses/gpl-2.0-faq.en.html#IfLibraryIsGPL). Because of this we would have to relicense [esp-serial-flasher](https://github.com/espressif/esp-serial-flasher) and [esptool-js](https://github.com/espressif/esptool-js) to GPL-2.0 as well as any SW that uses these codebases.

### Solution

From Espressif's and my perspective, the easiest and best solution to aforementioned problems would be relicensing of [espflash](https://github.com/esp-rs/espflash) to dual license it under MIT and Apache 2.0. Dual licensing under MIT and Apache 2.0 is widely accepted within Rust community and provides the most permissive licensing (MIT) that gives us patent grant (Apache 2.0) and as a whole is compatible with GPL-2.0.

### Open Discussion

[...]

PR #229, opened 2022-09-06
Add support for erasing otadata

### LINKED ISSUE #119: When doing espflash→OTA→espflash, (sometimes?) the boot partition is not set
Basic overview, might be wrong:

1. Flash with a partition table that has `ota_0` and `ota_1`; it presumably flashes to `ota_0`
2. Perform an OTA update to a firmware with changed behavior, which will write to `ota_1` (the not-in-use partition) and set that as the boot partition
3. Flash with espflash again, again using a firmware that behaves differently

The result is that the ESP32 continues to behave as if step 3 never happened. I *assume* that it's always flashing to `ota_0` and not setting it as the boot partition, so the ESP simply continues booting from `ota_1`.

I'm using current git version of espflash and an ESP32-C3. I accidentally made a String that slowly expands infinitely, did an OTA, then tried to fix it repeatedly using USB flashes that did not change the running code. Above is my best *guess* at what's going wrong.

ESP-IDF's `otatool` can [set the boot partition](https://github.com/espressif/esp-idf/blob/b0ca71f302b721236c3eaf0fe216482b04d70e40/components/app_update/otatool.py#L89), according to [this doc](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/ota.html#codecell5). Alternatively erasing the `otadata` partition should also reset to `ota_0`.

Although I'm not entirely sure the description/repro is right, running `otatool.py --port COM3 switch_ota_partition --slot 0` resulted in it running the firmware I was trying to run, so I'm fairly certain that the basic 'flashed firmware not being set to boot' issue is real.

Note that I might be doing something horribly wrong and I previously used the Arduino core to work with ESP32s, not the IDF or Rust.

<details>
  <summary>Partition table</summary>

```
# Name,   Type, SubType, Offset,   Size, Flags
# Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
nvs,      data, nvs,     ,        0x6000,
otadata,  data, ota,     ,        0x2000,
phy_init, data, phy,     ,        0x1000,
ota_0,    app,  ota_0,   ,        0x1C0000,
ota_1,    app,  ota_1,   ,        0x1C0000,
```
</details>

maximeborges:
Closes #119

This PR adds a `--erase-otadata` argument when flashing.

This will force using stub since it's using the `ERASE_REGION` command.
I also implemented the `ERASE_FLASH` command but didn't do any particular test nor did I implement an argument switch to use it.

I guess implementing two other commands for both `ERASE_REGION` and `ERASE_FLASH` would make sense, but having a first support for erasing the otadata partition thus allowing to reflash the chip after an OTA update seems more of a priority.

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -21,9 +21,9 @@ use crate::{
     cli::monitor::monitor,
     cli::serial::get_serial_port_info,
     elf::{ElfFirmwareImage, FlashFrequency, FlashMode},
-    error::Error,
+    error::{Error, NoOtadataError},
     flasher::FlashSize,
-    Chip, Flasher, ImageFormatId, InvalidPartitionTable, PartitionTable,
+    Chip, Flasher, ImageFormatId, InvalidPartitionTable, PartitionTable, MissingPartitionTable,
 };
 
 pub mod config;
@@ -59,6 +59,10 @@ pub struct FlashOpts {
     /// Open a serial monitor after flashing
     #[clap(long)]
     pub monitor: bool,
+    /// Erase the OTADATA partition
+    /// This is useful when using multiple OTA partitions and still wanting to be able to reflash via espflash
+    #[clap(long)]
+    pub erase_otadata: bool,
 }
 
 #[derive(Debug, Clone, Parser)]
@@ -280,6 +284,7 @@ pub fn flash_elf_image(
     flash_mode: Option<FlashMode>,
     flash_size: Option<FlashSize>,
     flash_freq: Option<FlashFrequency>,
+    erase_otadata: bool,
 ) -> Result<()> {
     // If the '--bootloader' option is provided, load the binary file at the
     // specified path.
@@ -306,6 +311,23 @@ pub fn flash_elf_image(
     } else {
         None
     };
+
+    if erase_otadata {
+        let partition_table = match &partition_table {
+            Some(partition_table) => partition_table,
+            None => return Err((MissingPartitionTable {}).into()),
+        };
+
+        let otadata = match partition_table.find("otadata") {
+            Some(otadata) => otadata,
+            None => return Err((NoOtadataError {}).into()),
+        };
+
+        let offset = otadata.offset();
+        let size = otadata.size();
+
+        flasher.erase_region(offset, size)?;
+    }
 
     // Load the ELF data, optionally using the provider bootloader/partition
     // table/image format, to the device's flash memory.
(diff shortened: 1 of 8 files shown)


espflash/src/cli/mod.rs
```diff
@@ -307,6 +312,23 @@ pub fn flash_elf_image(
         None
     };
 
+    if erase_otadata {
+        let partition_table = match &partition_table {
+            Some(partition_table) => partition_table,
+            None => return Err((MissingPartitionTable {}).into()),
+        };
+
+        let otadata = match partition_table.find("otadata") {
```
jessebraham, line 321: I don't believe it's _required_ for this partition to be named `otadata`, only that it has type `data` and subtype `ota`. The first line of the [OTA Data Partition](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/ota.html#ota-data-partition) documentation seems to indicate this is the case.

jessebraham (approved):
Awesome, thanks for getting that fixed up. LGTM!

maximeborges:
>One other small thing, we're in the process of re-licensing the project so if you wouldn't mind please commenting in https://github.com/esp-rs/espflash/issues/224 stating that this is okay with you that'd be much appreciated!

Looking at the last comment I guess it's already decided, but I'm definitely in favor of this change.

espflash/src/cli/mod.rs
```diff
@@ -59,6 +60,10 @@ pub struct FlashOpts {
     /// Open a serial monitor after flashing
     #[clap(long)]
     pub monitor: bool,
+    /// Erase the OTADATA partition
+    /// This is useful when using multiple OTA partitions and still wanting to be able to reflash via espflash
+    #[clap(long)]
+    pub erase_otadata: bool,
```
svenstaro, line 66: Can't you model the dependency using `requires = "use-stub"`?

espflash/src/main.rs
```diff
@@ -92,6 +92,10 @@ fn main() -> Result<()> {
 
     debug!("options: {:?}", opts);
 
+    if opts.flash_opts.erase_otadata {
+        opts.connect_opts.use_stub = true;
+    }
```
svenstaro, line 97: This dependency could perhaps be modeled using the clap option `requires`?

espflash/src/cli/mod.rs
maximeborges: Thanks for the idea, used in https://github.com/esp-rs/espflash/pull/273

PR #230, opened 2022-09-06
Allow skipping the padding for image files

Janrupf:
Currently `espflash` creates a 16MB image when specifying a flash size of 16MB. When mass producing devices a flash size of 16MB significantly slows down the flashing process without any benefit.

In order to not break backwards compatibility and expectations a new command line switch has been introduced `--skip-padding`, which reduces the image size to only what is really used.

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -124,6 +124,9 @@ pub struct SaveImageOpts {
     /// Custom partition table for merging
     #[clap(long, short = 'T')]
     pub partition_table: Option<PathBuf>,
+    /// Don't pad the image to the flash size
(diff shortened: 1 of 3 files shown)


jessebraham (approved):
This seems reasonable enough, thank you for contributing!

Before I merge, I just wanted to note that we are in the process of re-licensing the project (see https://github.com/esp-rs/espflash/issues/224 for details), so if you could please comment in that issue or in this PR stating that you consent to the new licensing (assuming you do, of course!) then I will happily merge this.

Janrupf:
Sure thing!
https://github.com/esp-rs/espflash/issues/224#issuecomment-1240971135

jessebraham:
Thanks!

PR #232, opened 2022-09-09
Check (and display) app vs partition size before flashing

jendakol:


jessebraham (changes requested):
This looks fine overall, thank you for contributing! I know it's a bit of a pain, but if you could just get the CI green that would be very much appreciated and I can get this merged.

Additionally, we are in the process of trying to re-license this project to MIT/Apache 2. Assuming you consent to your contributions being under these licenses rather than GPL2 could you please just say so in [this RFC](https://github.com/esp-rs/espflash/issues/224)?

jendakol:
Hey @jessebraham, I don't really care about the licensing so I've posted the consent 😄  
As for this PR, I'll try to fix the build. What bothers me more is that I realized that my change is related only to ESP32, not to all other variants of the chip. It's kind of hard to develop it as I have only ESP32 available right now :-) What do you think about that?

jessebraham:
Thank you for doing that!

This does actually work on the ESP32, ESP32-C3, ESP32-S2, and ESP32-S3 (I've tested them all). The `esp32bootloader.rs` file is probably not the best name, this refers to all devices starting from the ESP32 and moving forward; this only really excludes the ESP8266.

jendakol:
Does it? 😮 It didn't look like that when I was exploring the code 😂 (maybe because I was looking at ESP8266 🤔)  
Anyway, let me fix the build and we're good to go then 😎

jessebraham (approved):
Awesome, thanks again!

PR #233, opened 2022-09-10
Add option --bin for choosing one binary when there are more than one binaries

wzhd:
A beginner in esp decided to build several slightly different binaries by placing the source codes in `src/bin` in the same project. The following error message is produced when the binary to flash is not specified:

```
 Error: cargo_espflash::multiple_artifacts
 
   × Failed to build project
   ╰─▶ Multiple build artifacts found
  help: Please specify which artifact to flash using --bin
```

Running `cargo espflash --bin myapp1 /dev/ttyACM0` doesn't seem to solve it, though, with this message:

```
error: Found argument '--bin' which wasn't expected, or isn't valid in this context

      If you tried to supply `--bin` as a value rather than a flag, use `-- --bin`
```

This commit adds the option `--bin` to build_options, which has worked in my case.

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -84,6 +84,9 @@ pub struct BuildOpts {
     /// Example to build and flash
     #[clap(long)]
     pub example: Option<String>,
+    /// Binary to build and flash
+    #[clap(long)]
+    pub bin: Option<String>,
     /// Specify a (binary) package within a workspace to be built
     #[clap(long)]
     pub package: Option<String>,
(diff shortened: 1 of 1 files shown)


jessebraham (approved):
Good catch, thank you for fixing this! I've relaxed the Clippy checks a bit on `master` since this has been opened so I won't worry about the failing check in this PR, it's a pretty silly reason.

Just one final thing before I merge, we are in the process of trying to re-license this project to MIT/Apache 2. Assuming you consent to your contributions being under these licenses rather than GPL2 could you please just say so in [this RFC](https://github.com/esp-rs/espflash/issues/224)?

__EDIT:__ the re-licensing has completed, so I hope you're okay with that as I'm going to go ahead and merge this 😁

PR #216, opened 2022-08-07
enable support for loader stubs (via --use-stub arg)

ryankurte:
adds a rough impl for loading flash stubs into ram towards #215 (and https://github.com/esp-rs/esp-idf-template/issues/48), enabled with the `--use-stub` argument so this is a strictly opt-in addition and doesn't change existing behaviour.

(for some reason `ConnectOpts` is part of `main::Opts` and `Subcommand::(BoardInfo|SerialMonitor)` which can be a bit confusing / seems like it should only be in the root?)

tested `board-info` and flashing with on my `esp32-pico-v3`:

```
> cargo run --bin espflash -- board-info --use-stub /dev/ttyUSB0
Serial port: /dev/ttyUSB0
Connecting...

Using flash stub
Loading flash stub for chip: Esp32
Write 3432 byte stub text
Write 4 byte stub data
Finish stub write
Stub written...
Re-detected chip: Esp32
Chip type:         ESP32 (revision 3)
Crystal frequency: 40MHz
Flash size:        8MB
Features:          WiFi, BT, Dual Core, 240MHz, Embedded Flash, Embedded PSRAM, Coding Scheme None
MAC address:       50:02:91:a5:80:48
```

```
> cargo +esp espflash --use-stub /dev/ttyUSB0 --monitor
Serial port: /dev/ttyUSB0
Connecting...

Matched SerialPortType::Unknown
Using flash stub
Loading flash stub for chip: Esp32
Write 3432 byte stub text
Write 4 byte stub data
Finish stub write
Stub written...
Re-detected chip: Esp32
    Finished dev [optimized + debuginfo] target(s) in 0.10s
Chip type:         ESP32 (revision 3)
Crystal frequency: 40MHz
Flash size:        8MB
Features:          WiFi, BT, Dual Core, 240MHz, Embedded Flash, Embedded PSRAM, Coding Scheme None
MAC address:       50:02:91:a5:80:48
[00:00:01] ########################################      16/16      segment 0x1000                                                                [00:00:00] ########################################       1/1       segment 0x8000                                                                [00:00:29] ########################################     305/305     segment 0x10000                                                               
Flashing has completed!
Commands:
    CTRL+R    Reset chip
    CTRL+C    Exit

���ets Jul 29 2019 12:21:46
rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 271414342, SPIWP:0xee
[...]

PR #239, opened 2022-09-16
Re-design the CLI interface

### LINKED ISSUE #222: save-image argument order doesn't work as a runner for cargo
The order of arguments for the save-image subcommand is `<chip> <source> <destination>`.  If it was `<chip> <destination> <source>`, then the command could be used as a runner for cargo.  This is because cargo appends the source file to the end of the given command.  Alternatively there could be flags for the arguments so that the ordering is user decidable: `-c <chip> -d <destination> -s <source>)`.

If there is another way for me to solve this without changing the cli then please let me know!  I do see how the current ordering makes sense aesthetically.

### LINKED ISSUE #236: Re-work the command-line interface to be more ergonomic and less confusing
This work is already very much in-progress, just creating a tracking issue for it.

The biggest issue is flashing when no subcommand is provided. Any arguments passed _prior_ to a subcommand are silently ignored, so it was not always clear to users why their settings were not being applied.

Any dependencies related to the command-line interface should also be made optional behind a `cli` feature, which will be enabled by default.

jessebraham:
This PR includes a re-design of the command-line interface, most notably adding a `flash` subcommand.

In the current interface, since there is no flashing subcommand, it's possible to pass arguments _before_ a subcommand which are accepted but silently ignored. This is bad UX and is a common source of confusion for users. Adding a flashing subcommand eliminates this issue entirely.

While I was breaking things, I did a bit of refactoring as well. This was mostly just moving types to the module they belong in, simplifying some functions, etc. Nothing too interesting. I've also renamed a number of types/fields/functions/etc. to be more in line with other parts of the code, eg.) CLI arguments now use `arg` instead of `opt`, as their types are even `clap::Arg`.

Additionally I have created a `cli` feature which all CLI-related dependencies are gated behind. This allows these packages to be excluded when using `espflash` as a library. This feature is enabled by default. The executable was renamed from `main.rs` to `bin/espflash.rs` to accommodate this.

Beyond addressing whatever changes are requested, I'm not sure that there is much more to add to this PR. There are additional UX-related changes I would like to make but they are out of scope and deserve their own PRs, I think.

I would like to spend some time reviewing this and really picking it apart before we merge it. I have marked this as a draft until I am comfortable with its state. Small details such as short/long names, docstrings (which are displayed in help messages) and argument names are important, so please nitpick!

Closes #222 
Closes #236


hnidoaht-101:
hi @jessebraham I come here via esp-rs chat channel. Its clear that I'm totally newbie, so please forgive if I have some naive questions 😅. Is there any reasons you decided to switch from `tracing-subcriber` crate to `env-logger`?

jessebraham:
@hnidoaht-101 for context, `tracing-subscriber` was added in a [recent PR](https://github.com/esp-rs/espflash/pull/216), we did not have any logging prior and I just really didn't scrutinize this inclusion (partially because I was already working on revamping the CLI anyways, so it didn't really matter much).

`env-logger` is what I would normally use in these situations, and it's a much simpler package. I frankly had never even heard of `tracing-subscriber` before that PR, so if there is some tangible benefit in using it then I would consider reverting, but I'm not sure what said benefit(s) would be.

bjoernQ:
Haven't looked much into this yet but definitely thumbs-up for adding a `flash` command and the other changes described here

espflash/src/cli/mod.rs
```diff
@@ -32,94 +29,99 @@ pub mod monitor;
...
+    pub file: PathBuf,
+    /// Boolean flag to merge binaries into single binary
+    #[clap(long)]
+    pub merge: bool,
+    /// Custom partition table for merging
+    #[clap(long)]
+    pub partition_table: Option<PathBuf>,
+    /// Don't pad the image to the flash size
+    #[clap(long)]
+    pub skip_padding: bool,
```
svenstaro, line 120: Please change this as per https://github.com/esp-rs/espflash/pull/241/

svenstaro (changes requested):

espflash/src/cli/mod.rs
jessebraham: These have been updated in [0c63abf](https://github.com/esp-rs/espflash/commit/0c63abf179b7ce158ce2f1445d1501a11ddf0905).

jessebraham:
Thanks for taking a look!

I wasn't sure when to bump the version, I can set it to `2.0.0-dev` for now though I suppose. I do plan on submitting a handful more PRs prior to releasing (mostly just refactoring, but now is the time to break public APIs if we're gonna do it!)

PR #240, opened 2022-09-16
cargo-espflash: disable build-std check to allow users of cmdline -Zbuild-std

### LINKED ISSUE #221: cargo-espflash refuses to work when using `-Zbuild-std` instead of `.cargo/config.toml` `unstable.build-std`
I normally don't configure my `.cargo/config.toml` to do `-Zbuild-std` because I have a project where some targets are built with `build-std`, and some are not. Cargo's config doesn't provide for enabling `unstable.build-std` depending on target and/or cfg, so I pass it on the command line (via a helper script or directly).

cargo-espflash checks the `.cargo/config.toml` and rejects my configuration, blocking the build.

As a workaround, I've switched to either using a modified `cargo-espflash` with the `cargo_config.has_build_std()` commented out, or using `espflash` instead.

codysch:
Fixes #221

This check does not examine the cmdline arguments, resulting it in improperly blocking use of `cargo-espflash` in some cases for xtensa based esp's.

### DIFF
--- a/cargo-espflash/src/cargo_config.rs
+++ b/cargo-espflash/src/cargo_config.rs
@@ -7,26 +7,13 @@ use std::path::{Path, PathBuf};
 #[derive(Debug, Deserialize, Default)]
 pub struct CargoConfig {
     #[serde(default)]
-    unstable: Unstable,
-    #[serde(default)]
     build: Build,
 }
 
 impl CargoConfig {
-    pub fn has_build_std(&self) -> bool {
-        !self.unstable.build_std.is_empty()
-    }
-
     pub fn target(&self) -> Option<&str> {
         self.build.target.as_deref()
     }
-}
-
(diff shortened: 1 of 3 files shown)


jessebraham:
Thanks for the PR! I think I'd prefer to solve the problem rather than just ripping this feature out, building without this flag causes a large number of build errors and it's not immediately clear what the issue is if you haven't seen this before. The only real problem with the current check is that it's ignoring the command-line arguments.

I have a pretty simple fix for this locally so if you would like to change your PR that would be fine, but otherwise I'm happy to push my solution instead.

jmesmon:
@jessebraham let's go ahead with your version, then we can close this PR.

codysch:
@jessebraham let's go ahead with your version, then we can close this PR

jessebraham:
I've opened #242 which implements a fix for this.

PR #241, opened 2022-09-19
Correctly declare some arguments as requiring --merge

svenstaro:
In the context of `save image`, the `partition_table` and `skip_padding` arguments currently are only consumed inside a conditional block that's behind `if merge`. As such, we should declare this here as well to make it less surprising for users.


svenstaro:
Ok, done. You can close this MR once you have made the changes.

jessebraham:
My other PR has been updated.

PR #242, opened 2022-09-20
Fix the build-std check in cargo-espflash

jessebraham:
Closes #240


MabezDev (approved):
Looks good to me! Maybe one day our xtensa targets will be tier 2 :D

PR #245, opened 2022-09-27
Refactor: reduce the number of top-level modules, combine common functionality

jessebraham:
This is the first of three PRs performing various refactorings.

This PR is fairly simple, and does two things:

- Moves the `encoder` module into the `connection` module, as this is the only place it is referenced (and the encoder is an implementation detail).
- Moves the `stubs` module into the `flasher` module. Same reasons as above.


bjoernQ (approved):
LGTM - totally makes sense to me

PR #246, opened 2022-09-27
Refactor release CI and fix release binary permissions for cargo-binstall

SergioGasquez:
I've combined the 3 yaml files that were used for releases into a single one. If you think this adds more complexity, I am happy to go back as it was and only apply the permission changes, which were caused by `papeloto/action-zip@v1` action.
You can test the changes using the release [`v1.7.21` at my fork](https://github.com/SergioGasquez/espflash/releases/tag/v1.7.21):
1. Clone my fork
2. Edit `cargo-espflash/Cargo.toml` and `espflash/Cargo.toml`
   1. Set `version = "1.7.21"`
   2. Set `repository = "https://github.com/SergioGasquez/espflash"`
4. Install:
   1. `cargo binstall  --manifest-path cargo-espflash/Cargo.toml cargo-espflash`
   2. `cargo binstall  --manifest-path espflash/Cargo.toml espflash`


jessebraham (changes requested):
Thank you for your work on this!

Sorry to to do this to you, but if you don't mind I'd prefer to have these files split up. I think that `package.yml` and `rpi.yml` can likely be combined, but I do prefer having `release.yml` being quite simple.

If you're not willing to make these changes I understand and we can merge this as-is. I just prefer having a bit more separation because YAML sucks to read.

SergioGasquez:
Hi Jesse!
No worries, your point is more than fair, so happy to change it! I've merged `rpi.yml` and `package.yml` into a single file while leaving `release.yml` as it was. If you still think the previous approach was better or this can be improved in any way, do not hesitate to ask me to revert or make more changes!

jessebraham (approved):
Looks great, thank you for changing that!

PR #249, opened 2022-10-03
Refactor: rename image format modules to allow for future expansion

jessebraham:
This is the third in my series of planned refactors, following #245 and #247. This should be it for now, I hope.

Nothing too crazy here:

- Renamed the image formats to be a bit more specific, paves the way for adding additional image formats in the future
    - This is done in anticipation of eventually adding `mcuboot` support
- Moves a couple functions into the modules they belong in

### DIFF
--- a/espflash/src/elf.rs
+++ b/espflash/src/elf.rs
@@ -244,29 +244,3 @@ impl<'a> From<CodeSegment<'a>> for RomSegment<'a> {
         }
     }
 }
-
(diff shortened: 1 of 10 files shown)


JurajSadel (approved):
LGTM!

MabezDev:
LGTM! I always found the old layout confusing when trying to find the extended header stuff :D.

PR #251, opened 2022-10-03
Update to `clap@4.0.x`

### LINKED ISSUE #248: Update to `clap@4.0.x`
There is a new major release of [clap](https://github.com/clap-rs/clap), so we should update to this. I have already completed this task, but I am holding off on opening a PR until the `clap` releases have stabilized a bit. There have been 5 releases in the last ~24 hours at the time of writing.

jessebraham:
Of course, almost immediately after reworking the CLI they had to publish a new major version on us 😁 `cargo_toml` published an update requiring changes as well, so I've updated it and any other dependencies with new versions.

Unfortunately in this new version of `clap` they've taken away the `possible_values` attribute, so I've had to add a simple macro to replace it. This is incredibly unfortunate, but also not the end of the world.

There were also some small issues with our existing structures for the CLI arguments, so I had to address those as well. The `--monitor-baud` arg should not have been added to `ConnectArgs` (this was my bad) so I moved it out. Addressed some visual stuff too, just lowercasing options, etc.

Closes #248

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -212,7 +212,7 @@ version = "2.0.0-dev"
 dependencies = [
  "cargo_metadata",
  "cargo_toml",
- "clap",
+ "clap 4.0.8",
  "env_logger",
  "espflash",
  "log",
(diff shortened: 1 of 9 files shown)


SergioGasquez:
Maybe it's also worth updating the `#[clap(...)]` attributes with the new derive attributes: https://epage.github.io/blog/2022/09/clap4/#more-specific-derive-attributes

jessebraham:
Thanks for the link, would be really nice if they had some sort of migration guide or something. Half of the issues I ran into i only fixed because I spent a bunch of time searching through the issues and discussions in their repository 😅 Looks like there are a few things I should update though so I'll get that done shortly!

PR #254, opened 2022-10-04
Remove the `espmonitor` dependency, re-implement our own monitor

### LINKED ISSUE #253: The `espmonitor` dependency must be removed (as it is GPL)
Since we have relicensed to MIT/Apache-2.0 we are no longer able to use dependencies licensed under GPL. As such we must replace `espmonitor` with a bespoke solution.

I have a clean-room implementation which is working as far as I can tell, so I will PR it once it's been cleaned up and tested some more.

jessebraham:
Unfortunately since `espmonitor` is licensed under GPL we are no longer able to use it after the relicensing. This adds back in some of the code which was removed when adding `espmonitor`, and then re-implements the necessary functionality to get us back to where we were.

One added benefit of this new implementation is that `print!()` actually works now, as we do not buffer until a newline like `espmonitor` does.

Closes #253.


espflash/src/cli/monitor/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+use self::{line_endings::normalized, symbols::Symbols};
+use crate::{connection::reset_after_flash, interface::Interface};
+
+mod line_endings;
+mod symbols;
+
+// Pattern to much a function address in serial output.
+// All function addresses begin with '0x4', followed by exactly 7 hex digits.
+lazy_static! {
+    static ref RE_FN_ADDR: Regex = Regex::new(r"0x4[[:xdigit:]]{7}").unwrap();
```
bjoernQ, line 25: > All function addresses begin with '0x4

Not true for functions in IRAM and also sometimes the names of statics are interesting

bjoernQ (approved):
LGTM - works great
Really happy to not need new-lines to print anything anymore!
Regarding the regex - it's up to you if you want to adapt it now or not - not that much important

MabezDev (approved):

PR #255, opened 2022-10-05
Add `raspberry` feature checks in CI.

### LINKED ISSUE #250: Modify CI to build and release both with and without the `raspberry` feature
Presently the `raspberry` feature is not enabled in our CI workflow. Since we're unable to test builds for this without actually doing so on a Pi (or messing around with cross-compilation) it'd be best if we could check this automatically.

SergioGasquez:
- Went back to the old approach of having two files for generating release packages: One for raspberry targets, and another for the rest of the targets.
- Added `raspberry_rust.yml` which is the equivalent of `rust.yml` but for raspberry targets (I removed `fmt` check as it does not change between targets AFAIK)

Closes #250

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -24,38 +24,30 @@ jobs:
       fail-fast: false
       matrix:
         crate: ["espflash", "cargo-espflash"]
-
     steps:
       - uses: actions/checkout@v2
-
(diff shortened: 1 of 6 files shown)


jessebraham (approved):
LGTM, thanks for getting that fixed up!

PR #262, opened 2022-10-18
Re-write all the READMEs

jessebraham:
Pretty self-explanatory.

I wanted to pull as much out of the repo-level `README` as possible, as you only ever see it looking at the repository on GitHub. The other two files are rendered on [crates.io](https://crates.io) so they contain all relevant information, with plenty of duplication between the two files as necessary.

Rendered views:
- [`cargo-espflash/README.md`](https://github.com/esp-rs/espflash/tree/d0402d964bf029ac1aa81a31b938eedd03de96df/cargo-espflash)
- [`espflash/README.md`](https://github.com/esp-rs/espflash/tree/d0402d964bf029ac1aa81a31b938eedd03de96df/espflash)
- [`README.md` ](https://github.com/esp-rs/espflash/blob/d0402d964bf029ac1aa81a31b938eedd03de96df/README.md)

Please let me know if you feel anything is missing, or if any formatting/wording/etc. can be improved.

### DIFF
--- a/espflash/README.md
+++ b/espflash/README.md
@@ -1,89 +1,94 @@
 # espflash
 
-Serial flasher utility for Espressif SoCs and modules.
+[![Crates.io](https://img.shields.io/crates/v/espflash?labelColor=1C2C2E&color=C96329&logo=Rust&style=flat-square)](https://crates.io/crates/espflash)
+[![docs.rs](https://img.shields.io/docsrs/espflash?labelColor=1C2C2E&color=C96329&logo=rust&style=flat-square)](https://docs.rs/espflash)
+![MSRV](https://img.shields.io/badge/MSRV-1.60-blue?labelColor=1C2C2E&logo=Rust&style=flat-square)
+![Crates.io](https://img.shields.io/crates/l/espflash?labelColor=1C2C2E&style=flat-square)
 
-Currently supports the **ESP32**, **ESP32-C3**, **ESP32-S2**, **ESP32-S3**, and **ESP8266**.
+A library and command-line tool for flashing Espressif devices over serial.
+
+Supports the **ESP32**, **ESP32-C2**, **ESP32-C3**, **ESP32-S2**, **ESP32-S3**, and **ESP8266**.
 
 ## Installation
 
-```shell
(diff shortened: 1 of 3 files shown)


SergioGasquez:
Just added a bunch of minor suggestions, feel free to reject them if you want. Also, I prefer not having `$` on the bash snippets so it's easy to copy, but, again this is a personal preference, so feel free to leave it as it is!

Good job reworking the readmes, I think they are way cleaner now!

espflash/README.md
```diff
@@ -1,89 +1,94 @@
...
 
-Currently supports the **ESP32**, **ESP32-C3**, **ESP32-S2**, **ESP32-S3**, and **ESP8266**.
+A library and command-line tool for flashing Espressif devices over serial.
+
+Supports the **ESP32**, **ESP32-C2**, **ESP32-C3**, **ESP32-S2**, **ESP32-S3**, and **ESP8266**.
 
 ## Installation
 
-```shell
+If you are installing `espflash` from source (ie. using `cargo install`) then you must have `rustc>=1.60.0` installed on your system. Additionally [libuv] must be installed; this is available via most popular package managers.
```
bjoernQ, line 14: maybe mention there is nothing to do on Windows?

espflash/README.md
```diff
@@ -1,89 +1,94 @@
...
 
 ```toml
-[connection]
-serial = "/dev/ttyUSB0"
-
-[[usb_device]]
-vid = "303A"
-pid = "8000"
+[target.'cfg(any(target_arch = "riscv32", target_arch = "xtensa"))']
+runner = "espflash --baud=921600 --ram --monitor /dev/ttyUSB0"
```
bjoernQ, line 74: the "--ram" option is maybe a bit confusing to first time users?

JurajSadel (approved):
I agree with Sergio's point about removing `$`.
LGTM, thanks for doing this @jessebraham!

espflash/README.md
jessebraham: I had just copy-pasted this from the previous README but I think you're right, it probably does more harm than good being there. Will update it!

PR #269, opened 2022-10-24
connection: Swap error and status in CommandResponse

maximeborges:
In [stub_flasher.c:319](https://github.com/espressif/esptool/blob/52278a9c695d9d4298bed00085947b8cb28b1539/flasher_stub/stub_flasher.c#L319) we can see that error is sent before status:
```c
SLIP_send_frame_data(error);
SLIP_send_frame_data(status);
SLIP_send_frame_delimiter();
```

This PR swap those two fields in the `CommandResponse` struct.


jessebraham (approved):
Sneaky, good catch!

PR #270, opened 2022-10-24
connection: error handling

maximeborges:
Since we just swapped the two fields, we need to check `error` instead of `status` now.

I've been mainly working with the stub since I'm erasing the OTA partition when flashing.
After taking a better look at the error handling there I'm not really sure where the values from `RomErrorKind` are coming from, since they are neither in the C stub code nor the `bootloader` and `bootloader_support` components.

I added the generic errors from the stub. There is other values used in other places in there but they are not under an enum and it would take quite a bit of time to reference all of them for now.

If anyone has any clue where those are coming from, otherwise I'd say that we can remove them at some point.

### DIFF
--- a/espflash/src/connection.rs
+++ b/espflash/src/connection.rs
@@ -226,7 +226,7 @@ impl Connection {
         for _ in 0..100 {
             match self.read_response().for_command(ty)? {
                 Some(response) if response.return_op == ty as u8 => {
-                    return if response.status == 1 {
+                    return if response.error != 0 {
                         let _error = self.flush();
                         Err(Error::RomError(RomError::new(
                             command.command_type(),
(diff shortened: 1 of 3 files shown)


maximeborges:
Mmh I'm not able to flash anymore, I'm getting a `NotEnoughData` at the `FlashDeflateEnd` step, switching to draft to investigate.

maximeborges:
The `size` field for `Command::FlashDeflateBegin` is the uncompressed size of the segment.
It was rounded up to the memory block size, and the stub was expecting more data when the flash ended,
even though the inflator was returning successfully.

jessebraham (changes requested):
Thank you for this! Overall it looks good to me.

Sorry but there's just one small typo I'd like to see fixed please 😁

jessebraham (approved):
Thanks for fixing that!

PR #281, opened 2022-11-01
Print app size when using any image format, cleanup/use new features

jessebraham:


espflash/src/image_format/esp8266.rs
```diff
@@ -87,9 +88,16 @@ impl<'a> Esp8266Format<'a> {
             data: Cow::Owned(common_data),
         };
 
+        let app_size = irom_data
+            .clone()
+            .map(|d| d.data.len() as u32)
+            .unwrap_or_default()
+            + flash_segment.data.len() as u32;
```
jessebraham, line 95: @bjoernQ @MabezDev Do either of you know if this is correct for the ESP8266? I wasn't sure if the IROM data should be included in the total "application" size or not.

MabezDev: I believe so, IROM being instruction ROM, aka flash

MabezDev (approved):
Looks good to me! The new `#[default]` macro is a nice QoL change, going to start using that!

PR #283, opened 2022-11-04
Only display progress bars when `cli` feature is enabled, allow passing progress callback

### LINKED ISSUE #274: Allow retrieving progress in write_bin_to_flash()
Currently `write_bin_to_flash()` does all its own printing and the progress is internal. I use `espflash` as a library with my own flashing application on top of it and it's currently impossible to retrieve the current flash progress from it. Could you perhaps change the function to take a `FnMut` that could be used to set the progress if desired?

jessebraham:
Mostly just decoupling the progress bars from the flash targets in this. I think now all of the CLI-only dependencies are optional. Not 100% happy with it, but frankly it's good enough.

Closes #274

### DIFF
--- a/espflash/Cargo.toml
+++ b/espflash/Cargo.toml
@@ -47,7 +47,7 @@ directories-next = { version = "2.0.0", optional = true }
 esp-idf-part = "0.1.1"
 env_logger = { version = "0.9.1", optional = true }
 flate2 = "1.0.24"
-indicatif = "0.17.1"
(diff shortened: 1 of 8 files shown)


jessebraham:
@svenstaro could you please confirm that this will work for you?

svenstaro:
I think it should work, yes.

MabezDev (approved):
Looks good to me :)

PR #294, opened 2022-11-14
Read metadata from specified package when in a workspace

### LINKED ISSUE #291: can't specify partition table in a workspace
In a cargo workspace like:

```
[workspace]
resolver = "2"
members = [
    "thing",
    "some-other-thing",
]
```

Where `thing` has 

```
[package.metadata.espflash]
partition_table = "partitions.csv"
```

The default partition table is used instead of the specified one. Looks like only the workspace  `Cargo.toml` is parsed leaving no place to declare a custom partition file. Ideally this should detect the workspace and load the `Cargo.toml` from the specific crate being processed.

jessebraham:
Closes #291

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -133,19 +133,18 @@ fn main() -> Result<()> {
     // displayed.
     check_for_update(env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"));
 
-    // Load any user configuraiton and/or package metadata, if present.
+    // Load any user configuration, if present.
     let config = Config::load().unwrap();
     let cargo_config = parse_cargo_config(".")?;
-    let metadata = CargoEspFlashMeta::load("Cargo.toml")?;
 
     // Execute the correct action based on the provided subcommand and its
     // associated arguments.
     match args {
         Commands::BoardInfo(args) => board_info(args, &config),
-        Commands::Flash(args) => flash(args, &config, &cargo_config, &metadata),
+        Commands::Flash(args) => flash(args, &config, &cargo_config),
         Commands::Monitor(args) => serial_monitor(args, &config),
         Commands::PartitionTable(args) => partition_table(args),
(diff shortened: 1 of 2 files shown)


jessebraham:
@liebman Does this seem reasonable and work for your use case?

liebman:
This gives me an error:
```
➜ cargo espflash flash --package candle --monitor -b 921600 -p /dev/tty.usbmodem14644301
Error: cargo_espflash::no_project

  × No Cargo.toml found in the current directory
  help: Ensure that you're running the command from within a cargo project
```
and if I run from the package directory:
```
➜ cargo espflash flash --monitor -b 921600 -p /dev/tty.usbmodem14644301 
Error: 
  × Failed to parse Cargo.toml
  ╰─▶ value from workspace hasn't been set
```

Cargo build does this just fine:
```
➜ cargo build --package candle
   Compiling candle v0.1.0 (/Users/chris.l/Dropbox/git/cat-mouse/crates/candle)
   Compiling esp-idf-ws2812 v0.1.0 (/Users/chris.l/Dropbox/git/cat-mouse/crates/esp-idf-ws2812)
    Finished dev [optimized + debuginfo] target(s) in 15.63s
```

liebman:
I found this parses the entire workspace and gives you package names and paths (and a ton of un-needed info in this case).
```
use cargo::core::Workspace;
use cargo::util::Config;
use std::env;
use std::fs;

fn main() {
    let args: Vec<String> = env::args().collect();

    for p in &args[1..] {
        let path = fs::canonicalize(p).unwrap();
        let config = Config::default().unwrap();
        let ws = Workspace::new(path.as_path(), &config).unwrap();
        print!("{:#?}", ws);
    }
}
```

jessebraham:
@liebman sorry forgot to comment when I closed this, I'm working on a (much better) version using `cargo`, it never even occurred to me to use it as a dependency (though it's obvious now) 😁 Thanks for the pointer!

PR #300, opened 2022-11-16
Add support for using custom cargo metadata when in a workspace

### LINKED ISSUE #291: can't specify partition table in a workspace

jessebraham:
Alright, let's try this again 😁 

I took the advice of @liebman and used `cargo` in place of the `cargo_toml` package, and this has resulted in a much more robust and flexible solution. Not entirely happy with the `load_metadata` function still, but it appears to work at least.

I did actually test this on hardware this time around, and it seems to be working as intended. @liebman if you could test this out one more time that would be appreciated!

Also cleaned up the errors module, bumped a couple dependencies, and some other little bits.

(Hopefully) closes #291

### DIFF
--- a/.github/workflows/raspberry_rust.yml
+++ b/.github/workflows/raspberry_rust.yml
@@ -32,7 +32,7 @@ jobs:
             cargo-command: check
             args: --lib --no-default-features --features=raspberry
           - name: Check MSRV
-            toolchain: "1.62"
+            toolchain: "1.63"
             cargo-command: check
             args: --features=raspberry
           - name: Unit Test
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -65,7 +65,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: "1.62"
+          toolchain: "1.63"
           override: true
       - uses: Swatinem/rust-cache@v1
       - uses: actions-rs/cargo@v1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,12 +65,18 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.19"
+version = "0.7.20"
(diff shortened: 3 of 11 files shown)


liebman:
Using a [small test project](https://github.com/liebman/espflash-test) I tested running at the projects root and in the binary crates directory.

I found that this needs to use a newer cargo than 0.63 or it does not have support for workspace inheritance and you get this
```
Error: cargo_espflash::invalid_workspace

  × The current workspace is invalid, and could not be loaded
  help: Ensure that a valid Cargo.toml file is in the executing directory
```

If I update it to 0.66 (latest) it works well. With one minor issue.

- Running from root works well (with .cargo/config.tom at workspace root level)
- Running from the binary crate fails if there is no `.cargo/config.toml` in the current directory as it can't find the target (it does not look in the workspace root directory)

Basically there currently has to be a `.cargo/config.toml` in the current directory to specify a target. Ideally it would look first in the package that will be flashed (--package option if running from root) and if one is not found there then the workspace root.

I see you have a comment about having to use an older version - would not that only apply if these were going to be run on an esp32 target?

jessebraham:
Sorry for the delays, I needed a little break from this project 😅 

Thank you for providing the example project. I have made a number of additional changes and I am able to flash your project both from the workspace root and from within the `candle` package, so I think we're in good shape! @liebman if you wouldn't mind just giving this one more test to verify it's working on your end that would be appreciated!

liebman:
This works!

SergioGasquez (approved):
LGTM, and seems to be working fine! Thanks for taking care of this!

PR #301, opened 2022-11-18
Re-export esp-idf-part

bugadani:
Library users may need to parse partition tables and it is surprising that the partition table is in a separate crate.

This PR re-exports the esp-idf-part crate for better developer experience. This way the users don't need to add the dependency and keep it in sync in the future.

### DIFF
--- a/espflash/src/lib.rs
+++ b/espflash/src/lib.rs
@@ -59,6 +59,8 @@ pub mod targets;
 
 mod command;
 
+pub use esp_idf_part;
+
 /// Logging utilties
 #[cfg(feature = "cli")]
 #[cfg_attr(docsrs, doc(cfg(feature = "cli")))]


jessebraham:
I had considered doing this previously, however this would mean that any semver-incompatible changes in `esp-idf-part` will force us to bump the major version here. Given that `esp-idf-part` is currently on `0.1.x` I'm a little hesitant to do this, but I will think on it some more.

bugadani:
> any semver-incompatible changes in esp-idf-part will force us to bump the major version here

Fair point, `cargo-semver-checks` to the rescue :P Jokes aside, pinning the dependency version would probably allow a breakage like this to be fixed.

The alternative is exposing the user to a weird error: if `espflash` depends on some version of `esp-idf-part` and the user manually adds another version, it's possible to end up with a compile error that goes something like "expected type X but found type X" which is a real head-scratcher for newcomers. But I don't expect many newcomers to start integrating espflash into their own projects, so this may not be an issue. 🤷‍♂️

Up to you. While I would find this a bit more convenient, semver concerns probably hold more weight.

jessebraham:
In our community meeting this week there were some changes requested for `esp-idf-part`, so I will be holding off on re-exporting it at least until this work is done.

PR #303, opened 2022-11-19
Make command module public

bugadani:
Functions like [`Connection::write_command`](https://docs.rs/espflash/1.7.0/espflash/connection/struct.Connection.html#method.write_command) are public, but the `Command` type, which is one of their argument, is not accessible outside of the crate. This PR makes the command module public so users have the opportunity to implement functionality other than flashing a binary.


jessebraham (approved):
Seems reasonable enough, thanks!

PR #322, opened 2022-12-14
Support ESPFLASH_BAUD env var

bjoernQ:
This makes it possible to use `ESPFLASH_BAUD` env-variable to specify the flashing baud-rate.

This might be useful since the max possible baud-rate might be different depending on the host


jessebraham (approved):
Thanks, good idea!

Issue #320, opened 2022-12-13
Prioritise recognized ports


MabezDev:
<img width="617" alt="Screenshot 2022-12-13 at 16 41 32" src="https://user-images.githubusercontent.com/6977289/207378134-ec2cd196-27e4-4a38-94e6-af6e67fc3c88.png">

The port(s) that the real board is connected to is recognized, but unfortunately at the bottom of the list. It would be nice to move this to the first item.

PR #324, opened 2022-12-18
Reorder ports so that known ports appear first in CLI

matoushybl:
Attempts to solve #320


jessebraham (approved):
Seems to be working as far as I can tell, thanks for the contribution!

PR #333, opened 2023-01-10
Add an optional callback trait which can be implemented and provided to most flashing functions

### LINKED ISSUE #310: espflash lib: There is no way to display progress unless the `cli` feature is enabled
As someone who integrates espflash into my own software, it would be nice if I could display progress to users. I can do this with the `cli` feature, but only if my program runs in the command line. Currently, I can't do the same in a TUI/GUI app, for example.

It would be nice if I could pass some trait implementation to the flash functions, or the flasher object, that is responsible for doing anything necessary.

jessebraham:
Hopefully this will make `espflash` a bit more usable as a library.

@bugadani does this seem reasonable enough to you? Would love to hear any feedback you might have.

I suppose this is sort of an extension of #283 as well, so @svenstaro will this still work for your use case as well? Sorry for changing this again 😁 

Closes #310.

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -10,8 +10,8 @@ use espflash::{
     cli::{
         self, board_info, config::Config, connect, erase_partitions, flash_elf_image,
         monitor::monitor, parse_partition_table, partition_table, print_board_info,
-        save_elf_as_image, serial_monitor, ConnectArgs, FlashConfigArgs, MonitorArgs,
-        PartitionTableArgs,
+        save_elf_as_image, serial_monitor, ConnectArgs, EspflashProgress, FlashConfigArgs,
+        MonitorArgs, PartitionTableArgs,
     },
     image_format::ImageFormatKind,
     logging::initialize_logger,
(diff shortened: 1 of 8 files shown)


bugadani:
Thanks!

This looks awesome, although I have two small comments:
 - I think you might want to pass in the `impl ProgressCallbacks` type by value. I think that if I really need a reference to some object, I might implement the trait for the reference myself. Explicitly borrowing sounds a bit unnecessary to me.
 - I think I would replace the optional arguments with an empty `ProgressCallbacks` implementation, which would allow cleaning up some of the progress report sites (like `write_segment`). Although, I don't know if this would be a good idea or a complete antipattern instead 🤷‍♂️

jessebraham:
@bugadani I briefly poked at this and tried to get it working with `impl ProgressCallbacks` but I kept running into issues regarding object safety. I am pretty busy with other projects right now so I can't really spend much time on this for a bit. If you're able to build on top of this that'd be great; otherwise, whenever I get back to it I will either update it myself or just merge it as-is. Even though the API isn't ideal it does at least work in its current form.

bugadani:
Oh, please don't feel like my comments are anything more than mere wishes. I'm pretty happy with the PR as is :)

jessebraham:
Okay, glad to hear that. In that case I'm just going to merge this I suppose, we still have the opportunity to tweak it if needed.

PR #345, opened 2023-01-17
Remove the USB Serial JTAG watchdog workaround now that we use the stub

jessebraham:


MabezDev:
I think we also need to update the stubs from https://github.com/espressif/esptool/commit/32e80148493a83d34bad635287cb97e458d2af58, I did report this last year but it only got fixed in the stub a few days ago.

MabezDev:
Actually, we probably shouldn't have removed this but instead executed if `use_stub` is false, sorry that comment was actually a bit misleading.

PR #354, opened 2023-02-14
Make `Interface` constructor public

bugadani:
As it turns out, users (me) might actually want to construct `Interface` objects. While the fields are public, the raspberry-enabled `new` does a few useful things that I don't necessarily want to duplicate in my program.


jessebraham (approved):
Seems reasonable enough, I think it was only protected because the `interface` module at one point was private, so this probably just never got updated. Thanks!

bugadani:
I missed it, too, when I was poking at raspberry support :)

PR #358, opened 2023-03-01
fix: esp32c3 direct boot

sethp:
Prior to this change, attempting to `espflash flash` with the `--format
direct-boot` argument targeting an esp32c3 would fail, reporting:

```
Error: espflash::unsupported_image_format

  × Image format direct-boot is not supported by the esp32c3 revision v0.3
  help: The esp32c3 only supports direct-boot starting with revision 3
```

However, direct-boot on this board was working prior to the changes in
https://github.com/esp-rs/espflash/commit/b25af06912e3c257261ac5ab9788217c0405d033, and looking at the [docs][esp32c3 revisions] it seems that the
mapping that was chosen for the `c3` specifically was `v0.REVISION`:

> ECO  | Revision (Major.Minor)
> ---------|-------------------
> ECO1 | v0.1
> ECO2 | v0.2
> ECO3 | v0.3
> ECO4 | v0.4

[esp32c3 revisions]: https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/system/chip_revision.html#revisions

Additionally, this PR contains a small cleanup change (https://github.com/esp-rs/espflash/commit/f0ab4dde79d18439bb499709d3ad9b6cf6ae26f9) and a proposal for moving/updating the error message emitted above.

### DIFF
--- a/espflash/src/error.rs
+++ b/espflash/src/error.rs
@@ -401,6 +401,7 @@ pub struct UnsupportedImageFormatError {
     format: ImageFormatKind,
     chip: Chip,
     revision: Option<(u32, u32)>,
+    context: Option<String>,
 }
 
 impl UnsupportedImageFormatError {
@@ -409,6 +410,7 @@ impl UnsupportedImageFormatError {
             format,
             chip,
             revision,
+            context: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


jessebraham (approved):
Sorry for taking so long to get to this, thanks for the fix!

sethp:
Not a problem, thanks!

PR #362, opened 2023-03-03
Reconnect to chip after build step since watchdog timer may have caused disconnect.

### LINKED ISSUE #361: `cargo-espflash` fails to flash when the build is too long
This appears to be another issue related to the watchdog behavior.  `cargo-espflash` connects to the chip before performing the build, and holds the connection.  If the duration of the build is longer than the watchdog timer, the chip resets.  When `cargo-espflash` finishes the build, it attempts to use the now-broken connection to flash, which fails.   

I've prepared two different approaches to fix this, both of which resolve the issue in my testing:
1. Don't connect to the chip until after the build is complete.  [changes](https://github.com/esp-rs/espflash/compare/main...junelife:espflash:joe/dont-connect-before-build) 
2. Reconnect to the chip after the build is complete. [changes](https://github.com/esp-rs/espflash/compare/main...junelife:espflash:joe/reconnect-after-build)

Solution 2, reconnecting results in a tidier patch IMO.  So I've put up a PR with that second solution: #362

Note that both solutions rely on reverting 1c054e84127c08a3ccfa60ada39a6185ed1c9ee0

jnross:
Fixes #361

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -157,13 +157,16 @@ fn flash(args: FlashArgs, config: &Config) -> Result<()> {
     let metadata = PackageMetadata::load(&args.build_args.package)?;
     let cargo_config = CargoConfig::load(&metadata.workspace_root, &metadata.package_root);
 
+    let flasher = connect(&args.connect_args, config)?;
+    let chip = flasher.chip();
+
+    let build_ctx =
+        build(&args.build_args, &cargo_config, chip).wrap_err("Failed to build project")?;
+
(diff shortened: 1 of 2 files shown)


jnross:
After using this for a couple of days, I realized a couple of deal-breaking drawbacks:
1.  If you don't specify the `--port`, you'll get the interactive selection prompt to pick a serial port twice.  This is a terrible experience.
2. If the watchdog timer doesn't expire because the build is fast or because you're using the ESP-PROG board, the reconnect attempt fails with `Device or resource busy`.

I'm still looking for a fix to #361, but this isn't it.

PR #363, opened 2023-03-06
Disable watchdog timer before build.

### LINKED ISSUE #361: `cargo-espflash` fails to flash when the build is too long

jnross:
Fixes #361

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ target/
 .vscode/
 .DS_Store
 *.swp
+.idea/
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -161,6 +161,7 @@ fn flash(args: FlashArgs, config: &Config) -> Result<()> {
     let chip = flasher.chip();
     let target = chip.into_target();
     let target_xtal_freq = target.crystal_freq(flasher.connection())?;
+    flasher.disable_watchdog()?;
 
     let build_ctx =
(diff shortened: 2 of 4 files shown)


jnross:
Sorry I forgot to double-check that I ran `rustfmt` before pushing.  Turns out there was a stray newline.  I've run rustfmt and pushed another commit with the fixes.

jessebraham (approved):
Sorry for taking so long to get to this, but LGTM! Thanks for fixing this!

jnross:
Thanks for considering and landing this :)

PR #366, opened 2023-03-15
Add `ESPFLASH_PORT` environment variable

matteocarnelos:
Simple PR that extends #322.

The addition of this env variable is particularly useful when you have multiple ESPs connected, as you can have different run configurations for the different ESPs with this env variable set accordingly.


jessebraham (approved):
Thanks! I've been meaning to add a few more of these variable, this being one of them 😁

PR #368, opened 2023-03-20
Update documentation

SergioGasquez:
- Fix clippy warnings
- Make CI fail if there are clippy warnings (`too_many_arguments` lint is allowed)
- Update docstrings


jessebraham:
Sorry, still haven't gotten around to going through all these changes.

Curious though, in the `Cargo.toml` files there are some formatting changes, and in various source files `super` imports are converted to `crate` imports; what was the reason for these changes? They just kind of seem like changing things for the sake of changing them, but I'm not sure if there was some other motivation for this or not.

espflash/src/flasher/mod.rs
```diff
@@ -153,7 +154,7 @@ impl FlashSize {
 
 impl FromStr for FlashSize {
     type Err = Error;
-
+    /// Create a `FlashSize` from a string
```
jessebraham, line 157:
```suggestion
    /// Create a [FlashSize] from a string
```

There may be other instances like this, but I'll just point out this one: we can link to internal types instead, which just makes the documentation a bit nicer to use IMO

jessebraham (changes requested):
Just a small change I'd like to see, plus the question I asked in the previous comment. Overall looks pretty good to me though, thanks for taking care of this!

jessebraham:
Also looks like I've made some changes which have caused conflicts here, so this will need a rebase. Sorry about that, I'll hold off on my other PR fiddling with the `Cargo.toml` files until this gets merged.

SergioGasquez:
> Curious though, in the Cargo.toml files there are some formatting changes, and in various source files super imports are converted to crate imports; what was the reason for these changes?

The` Cargo.toml` changes were the result of the formatting applied by VsCode, (I have "format on save" setting). As per the `super` vs `crate`, I basically change it in places where we already had an import of `crate` to unify all the "local" imports.  None of those changes add any value, tbh, so I am happy to revert them if you want.

jessebraham (approved):
Thanks for making those requested changes. Regarding the formatting stuff, I'm not too concerned out about it, just was curious. It's fine to just leave it as is, not worth the effort to change it back.

Thanks again for taking care of this!

PR #369, opened 2023-03-20
Add dependabot

SergioGasquez:
Adding dependabot can help us to update the dependencies, at least for CI actions. Should you have any other opinion, let me know, and I am more than happy to modify or close the PR.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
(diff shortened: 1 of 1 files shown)


jessebraham:
Thanks for the PR. I'm not generally a huge fan of dependabot, but I'll take some time to think about it. At the very least I'd like to hold off until we're ready for a full release I think, so we can leave this open for now.

jessebraham:
After thinking about this for awhile I don't really see any value in adding this. Dependencies are always updated prior to releases, there's no point bumping them over and over during development. I have enough notifications already 😁 

Thanks anyway, though!

PR #374, opened 2023-03-25
If exactly one port matches, use it

### LINKED ISSUE #367: option to auto-choose serial port if only one matches a known USB device
Because I often have a few USB-serial devices connected, my esp device gets a path like `/dev/ttyACMn` but the `n` is unpredictable. This prevents me from using the `serial` option in my `espflash.toml` file, and the presence of multiple serial ports also prevents `cargo espflash` from auto-connecting to one. But I usually connect only a single esp device, and so `cargo espflash` always highlights a single "known common dev board", and it's always the one I want.

Would you accept a PR to add an option (either a CLI flag or something in the config file) to say that if there's a single known dev board then auto-connect to it?

jneem:
Fixes #367.

Since I added the `matches` closure, it felt like some clean-up was appropriate. But if you'd prefer a more minimal PR I can do that too.

This also slightly changes the behavior of non-usb serial ports, for consistency. Previously, you could make them match by manually editing the config file and adding "vid=0, pid=0", but they wouldn't be sorted and highlighted as though they matched. After this change, non-usb serial ports won't auto-match at all.

### DIFF
--- a/espflash/src/cli/serial.rs
+++ b/espflash/src/cli/serial.rs
@@ -1,4 +1,3 @@
-use std::cmp::Ordering;
 #[cfg(not(target_os = "windows"))]
 use std::fs;
 
@@ -6,7 +5,7 @@ use crossterm::style::Stylize;
 use dialoguer::{theme::ColorfulTheme, Confirm, Select};
 use log::{error, info};
 use miette::{IntoDiagnostic, Result};
-use serialport::{available_ports, SerialPortInfo, SerialPortType, UsbPortInfo};
+use serialport::{available_ports, SerialPortInfo, SerialPortType};
 
 use crate::{
     cli::{config::UsbDevice, Config, ConnectArgs},
@@ -26,7 +25,7 @@ pub fn get_serial_port_info(
     // Users may optionally specify the device's VID and PID in the configuration
     // file. If no VID/PID has been provided, the user will always be prompted to
     // select a serial port. If some VID and PID were provided then the user will
-    // also be prompted to select a port, unless there is only one found and its VID
+    // also be prompted to select a port, unless there is only one found whose VID
     // and PID match the configured values.
     //
     // The call to canonicalize() was originally added to resolve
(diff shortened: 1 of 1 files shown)


jessebraham:
Thanks for this PR, glad to see you've cleaned things up a bit as well!

These changes look good to me, no issues testing on my local system. Unfortunately I've merged another PR updating some documentation and there's a small merge conflict now, sorry about that. If you don't mind giving this a rebase then I'm happy to merge it!

jneem:
Hm, I rebased but now when I test it I get an error when parsing my `~/.config/espflash/espflash.toml`:
```
Error:
  × TOML parse error at line 4, column 7
  │   |
  │ 4 | vid = "303a"
  │   |       ^^^^^^
  │ invalid type: string "303a", expected a borrowed byte array
  │
```
Was this an intentional change?

SergioGasquez:
> Was this an intentional change?

No, probably an error introduced when bumping the `toml` version in: https://github.com/esp-rs/espflash/pull/378, more specifically Id say this changes: https://github.com/esp-rs/espflash/pull/378/files#diff-16aedab0b163d4e6c5350f585959afd77901030799b31de2a9c3c47e64b9c253

espflash/src/cli/serial.rs
```diff
@@ -186,53 +185,50 @@ fn select_serial_port(
...
-                    } else {
-                        port_info.port_name.as_str().reset()
-                    };
-
-                    if let Some(product) = &info.product {
-                        format!("{} - {}", formatted, product)
-                    } else {
-                        formatted.to_string()
+            .map(|port_info| {
+                let formatted = if matches(&port_info) {
```
SergioGasquez, line 217: Clippy suggestion:
```suggestion
                let formatted = if matches(port_info) {
```

kyrias:
It would be nice if we could disable this behavior and always be prompted to select one.

Right now I commonly have both a ESP-C3-DevKitM-1 and an ESP-Prog connected and while the devkit is one of the "known" devices the ESP-Prog is not, which means that while previously I was always asked which one I wanted to flash to, now it always tries to flash the ESP-C3.

jneem:
I'm not opposed to adding a flag, but is there some reason you can't just add the not-yet-known device to `~/.config/espflash/espflash.toml` so that it becomes known?

kyrias:
For this specific case I don't have a problem with adding the ESP-Prog myself, but I personally don't _ever_ want it to automatically pick the serial port to use as I don't trust it to necessarily pick the correct one, and I don't trust the other side of various serial ports to handle the wrong one picked well.

kyrias:
As an example of why I don't want it to ever decide on its own, earlier today `espflash` ended up flashing an ELF file for an ESP32-S3 board onto an ESP32-C3 because I forgot to enable one of the ports on my USB hub.

Issue #370, opened 2023-03-21
Add the ability to specify a minimum compatible revision number


jessebraham:
There are going to be changes to the ROM API with an upcoming revision of the ESP32-C3 (ECO6), so we will need the ability to specify a minimum compatible revision so that we can verify the target device is compatible.

PR #375, opened 2023-03-25
Image header improvements and bugfixes

maxwase:
The original idea was to add ESP32S3 32MB flash support, but then I've discovered some bugs in current implementation such as
* Wrong image size & frequency serialization
* Incomplete header copying (min_chip_rev_full, max_chip_rev_full)

This PR also includes some documentation and interface improvements, as well as support for new flash sizes.
Workaround for #370

## Testing 

* This was tested on ESP32C3 4Mb and ESP32S3 32Mb
* [OTA test](https://github.com/maxwase/esp-ota)
* Regenerated image header diff (entry and max_chip_rev_full):
```diff
1,2c1,2
< 0000 0000: E9 04 02 20 E0 10 0D 40  EE 00 00 00 00 00 00 00  ... ...@ ........  
< 0000 0010: 00 FF FF 00 00 00 00 01  20 00 40 3F 6C 2E 00 00  ........  .@?l...  
---
> 0000 0000: E9 04 02 20 88 06 0D 40  EE 00 00 00 00 00 00 00  ... ...@ ........  
> 0000 0010: 00 00 00 00 00 00 00 01  20 00 40 3F C0 25 00 00  ........  .@?.%..  

```


espflash/src/flasher/mod.rs
```diff
@@ -93,49 +107,89 @@ pub enum FlashMode {
...
+                _128Mb => 7,
+                _256Mb => 8,
+                _ => return Err(Error::UnsupportedFlash(self as u8)),
+            },
+        };
+
+        Ok(encoded)
+    }
+
+    // components/esptool_py/esptool/esptool.py:151
```
SergioGasquez, line 173: Do you mind sharing a link to this reference? I assume this belongs to the esp-idf repo, but in `master` there is no such line in the esptool.py

SergioGasquez: Found it, here is a link to where it now lives: https://github.com/espressif/esptool/blob/f4d2510e2c897621884f433ef3f191e8fc5ff184/esptool/cmds.py#L42

espflash/src/flasher/mod.rs
```diff
@@ -93,49 +107,89 @@ pub enum FlashMode {
...
     /// 128 MB
-    _128Mb = 0x21,
+    _128Mb,
+    /// 256 MB
+    _256Mb,
 }
 
 impl FlashSize {
-    fn from(value: u8) -> Result<FlashSize, Error> {
+    // components/esptool_py/esptool/docs/en/advanced-topics/firmware-image-format.rst:19
```
SergioGasquez, line 138:
```suggestion
    // https://docs.espressif.com/projects/esptool/en/latest/esp8266/advanced-topics/firmware-image-format.html?highlight=0xe9#file-header 
```
I think its better to point at the rendered documentation when possible, this way, if the line or file name changes, it may still be valid

SergioGasquez (commented):
Thanks for your contribution! I think it's a more clean way to handle image headers, just left a few small comments. Do you mind resolving the conflicts? Once they are resolved I'll do tests it flashing a few devices

espflash/src/image_format/mod.rs
```diff
@@ -26,12 +31,70 @@ const WP_PIN_DISABLED: u8 = 0xEE;
 
 #[derive(Debug, Clone, Copy, Pod, Zeroable)]
 #[repr(C, packed)]
-struct EspCommonHeader {
+#[doc(alias = "esp_image_header_t")]
```
SergioGasquez, line 34: Maybe we could point to https://docs.espressif.com/projects/esptool/en/latest/esp32c3/advanced-topics/firmware-image-format.html#file-header and https://docs.espressif.com/projects/esptool/en/latest/esp32c3/advanced-topics/firmware-image-format.html#extended-file-header?

espflash/src/image_format/esp8266.rs
```diff
@@ -39,19 +38,20 @@ impl<'a> Esp8266Format<'a> {
...
         // Common header
         let flash_mode = flash_mode.unwrap_or_default() as u8;
-        let flash_freq = flash_freq.unwrap_or_default();
-        let flash_size = flash_size.unwrap_or_default();
-        let flash_config =
-            encode_flash_size(flash_size)? + encode_flash_frequency(Chip::Esp8266, flash_freq)?;
         let segment_count = image.ram_segments(Chip::Esp8266).count() as u8;
 
-        let header = EspCommonHeader {
+        let mut header = ImageHeader {
```
SergioGasquez, line 43: Maybe add a reference to ESP8266 file header: https://docs.espressif.com/projects/esptool/en/latest/esp8266/advanced-topics/firmware-image-format.html#file-header?

jessebraham:
I've just taken a cursory look at these changes and they look really great, thanks for fixing all of this. Clearly I missed a number of things during my big refactor last year 😅 

I will do some testing and take one more look at the code over the next couple days here. I have merged another PR adding some documentation so there are unfortunately a few merge conflicts, if you wouldn't mind rebasing this please in the meantime that would be appreciated! Sorry about that.

maxwase:
Thank you for review, rebased & updated some docs, let me know if you want to change doc format somehow

espflash/src/flasher/mod.rs
maxwase: On the one hand, permanent link is better because of the line, on the other hand, it would be nice to have a link to master version. I guess approach with the permalink is more durable

espflash/src/flasher/mod.rs
maxwase: Such a nice feature with `/esp-MODEL/`!

maxwase:
Btw, I'm not familiar with your version bumping policy, do I need to change it in this PR?

espflash/src/image_format/mod.rs
maxwase: Done

jessebraham:
I did some testing today and things are looking pretty good here, thanks again for all your work! With the exception of the ESP8266 (which I've commented on above) I think everything seems to be working as expected.

One small detail I did notice, since the ESP32-C2 does not support 40MHz as the flash frequency, and this is the default value, we do get an error when trying to flash without explicitly providing the `--flash-freq` option:

```text
Chip type:         esp32c2 (revision v1.0)
Crystal frequency: 40MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       10:97:bd:f1:e4:48
Error: espflash::unsupported_flash_frequency

  × The specified flash frequency '_40Mhz' is not supported by the esp32c2
```

This isn't great UX, but also can probably be handled in a separate PR. If you'd like to tackle it here you're of course welcome to, but it's up to you.

espflash/src/image_format/esp8266.rs
```diff
@@ -39,19 +38,20 @@ impl<'a> Esp8266Format<'a> {
...
-            flash_config,
             entry: image.entry(),
+            ..Default::default()
         };
+        header.write_flash_config(
+            flash_size.unwrap_or_default(),
+            flash_freq.unwrap_or_default(),
+            Chip::Esp8266,
+        )?;
         common_data.write_all(bytes_of(&header))?;
```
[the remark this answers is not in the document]
maxwase: Thanks for testing this! I've changed it back with some refactoring
I have decided to serialize while header, but write only 8 bytes of it. The operation itself is more expensive (16 zero bytes plus), but more durable and unified in my opinion

jessebraham (approved):
Thanks for making those changes! Also, thanks for the suggestions RE the C2; I'll open an issue for it.

PR #378, opened 2023-03-29
Update `toml` dependency and fix errors, feature gate `ctrlc` dependency

jessebraham:
- New version of `toml` with some breaking API changes, which were addressed
- I forgot to gate `ctrlc` behind the `cli` feature when I added it
    - #372


SergioGasquez (approved):
LGTM!

PR #381, opened 2023-04-02
Remove serde-hex crate due to save / load conflict

jamesgraves:
When using serde-hex, the vid and pid are saved as:

```
[connection]

[[usb_device]]
vid = "303a"
pid = "1001"
```

in the `~/.config/espflash/espflash.toml` file.

However, upon loading, we see an error message:

```
Error:
  × TOML parse error at line 4, column 7
  │   |
  │ 4 | vid = "303a"
  │   |       ^^^^^^
  │ invalid type: string "303a", expected a borrowed byte array
  │
```

The serde-hex source code hasn't been updated in 4 years, and is not apparently compatible with the toml crate.

Storing the vid and pid as decimal integers is not ideal, but it at least works correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,7 +34,7 @@ dependencies = [
  "gimli",
  "object",
  "rustc-demangle",
- "smallvec 1.10.0",
+ "smallvec",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


SergioGasquez:
Thanks for your contribution! I was about to have a look at this as we were discussing it in: https://github.com/esp-rs/espflash/pull/374. I wonder if we could easily use the [hex crate](https://github.com/KokaKiwi/rust-hex) with the [serde feature](https://docs.rs/hex/0.4.3/hex/serde/index.html) to keep it in hexadecimal.

SergioGasquez:
Closed by #383 to make it backward compatible, it needed to use hex instead of decimal

jamesgraves:
Thank you!

PR #382, opened 2023-04-04
Fix config file parsing

SergioGasquez:
- Remove `serde-hex` since it's no longer compatible with the new `toml` version
- Include the `hex` crate
- Add functions for serializing and deserializing
- Add test for serialization and deserialization methods

Not sure if this is the best way, but the file needed to be backward compatible.

### DIFF
--- a/espflash/src/cli/config.rs
+++ b/espflash/src/cli/config.rs
@@ -112,3 +112,53 @@ impl Config {
             .wrap_err_with(|| format!("Failed to write config to {}", self.save_path.display()))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde::Deserialize;
+
+    #[derive(Debug, Deserialize)]
+    struct TestData {
+        #[serde(deserialize_with = "parse_hex_u16")]
+        value: u16,
+    }
+
+    #[test]
+    fn test_parse_hex_u16() {
+        // Test no padding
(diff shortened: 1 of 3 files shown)


MabezDev (changes requested):

espflash/src/cli/config.rs
```diff
@@ -94,3 +120,81 @@ impl Config {
...
+    struct TestData {
+        #[serde(serialize_with = "parse_u16_hex", deserialize_with = "parse_hex_u16")]
+        value: u16,
+    }
+
+    #[test]
+    fn test_parse_hex_u16() {
+        // Test no padding
+        let input = "aaaa";
+        let result: Result<TestData, _> = toml::from_str(&format!("value = \"{}\"", input));
```
bjoernQ, line 139: probably a matter of taste but I would write something less verbose like
`let result: Result<TestData, _> = toml::from_str(r#"value = "aaaa""#);`

but really a matter of taste

bjoernQ:
maybe you also want to adapt `test_parse_u16_hex`

SergioGasquez:
> maybe you also want to adapt `test_parse_u16_hex`

Ups, forgot to push the changes after updating it. Thanks!

bjoernQ (approved):
LGTM

MabezDev (approved):
LGTM, thanks!

PR #397, opened 2023-05-03
Add a diagnostic to tell the user about the partition table format 

MabezDev:
When the image doesn't fit.

### DIFF
--- a/espflash/src/error.rs
+++ b/espflash/src/error.rs
@@ -44,6 +44,11 @@ pub enum Error {
     #[error(
         "Supplied ELF image of {0}B is too big, and doesn't fit configured app partition of {1}B"
     )]
+    #[diagnostic(
+        code(espflash::image_too_big),
+        help("Reduce the size of the binary or increase the size of the partition."),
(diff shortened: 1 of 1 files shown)


SergioGasquez (approved):
LGTM! Just fyi, miette supports a `url` param under `#[diagnostic]`, see https://github.com/zkat/miette#-diagnostic-code-urls

MabezDev:
> Just fyi, miette supports a url param under #[diagnostic], see https://github.com/zkat/miette#-diagnostic-code-urls

Thanks! I didn't know about that, I've updated the diagnostic to use the URL feature.

SergioGasquez (approved):

espflash/src/error.rs
```diff
@@ -44,6 +44,11 @@ pub enum Error {
     #[error(
         "Supplied ELF image of {0}B is too big, and doesn't fit configured app partition of {1}B"
     )]
+    #[diagnostic(
+        code(espflash::image_too_big),
+        help("Reduce the size of the binary or increase the size of the partition."),
```
jessebraham, line 49: Would have preferred this to say "increase the size of the __app__ partition", as it's pretty ambiguous now. Will fix it in another PR before release I guess.

PR #398, opened 2023-05-04
limit default partition size

### LINKED ISSUE #393: Flashing with ESP32-S3 with Octal flash flashes an unbootable image
After installing RC4 I noticed that flashing with the USB-SERIAL-JTAG port (in my case, S3) no longer correctly flashes the chip. The second stage bootloader complains about the partition table having partition size. I bisected from a known working state to find that #375 seems to be the culprit. 

```
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x3 (RTC_SW_SYS_RST),boot:0x18 (SPI_FAST_FLASH_BOOT)
Saved PC:0x403cdb4a
0x403cdb4a - __default_naked_level_7_interrupt
    at ??:??
SPIWP:0xee
Octal Flash Mode Enabled
For OPI Flash, Use Default Flash Boot Mode
mode:SLOW_RD, clock div:2
load:0x3fce3810,len:0x1690
0x3fce3810 - _stack_end_cpu0
    at ??:??
load:0x403c9700,len:0xbd0
0x403c9700 - __default_naked_level_7_interrupt
    at ??:??
load:0x403cc700,len:0x2dc8
0x403cc700 - __default_naked_level_7_interrupt
    at ??:??
SHA-256 comparison failed:
Calculated: ae126cb96d962b79703c4a4144654f513bc92dbad93c921b2e20c5558d0914dc
Expected: d1c78d7325094d95e7bb525e04f54c3212a0971c36efb48e3f3f056daece1a5e
Attempting to boot anyway...
entry 0x403c9900
0x403c9900 - __default_naked_level_7_interrupt
    at ??:??
I (60) boot: ESP-IDF v5.0-beta1-764-gdbcf640261 2nd stage bootloader
I (60) boot: compile time 11:32:39
I (60) boot: chip revision: V000
I (64) boot.esp32s3: Boot SPI Speed : 40MHz
I (69) boot.esp32s3: SPI Mode       : SLOW READ
I (74) boot.esp32s3: SPI Flash Size : 32MB
I (79) boot: Enabling RNG early entropy source...
I (84) boot: Partition Table:
I (88) boot: ## Label            Usage          Type ST Offset   Length
I (95) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (102) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (110) boot:  2 factory          factory app      00 00 00010000 01ff0000
I (117) boot: End of partition table
E (122) esp_image: partition size 0x1ff0000 invalid, larger than 16MB
E (129) boot: Factory app partition is not bootable
E (134) boot: No bootable app partitions in the partition table
```

Looking at the changes in #375, I'm not sure what would cause this. @maxwase, could you test flashing on your S3 board via USB serial jtag to verify this is a real issue?

MabezDev:
Some chips can have more than 16MB of flash, mine has 32MB. Historically espflash failed to detect this flash size (and defaulted to 4MB) but now it can. Without this change, the board will boot loop because the esp idf bootloader max size is only 16MB.

Closes #393

### DIFF
--- a/espflash/src/targets/mod.rs
+++ b/espflash/src/targets/mod.rs
@@ -31,6 +31,9 @@ use crate::{
     image_format::{ImageFormat, ImageFormatKind},
 };
 
+/// Max partition size is 16 MB
+const MAX_PARTITION_SIZE: u32 = 16 * 1000 * 1024;
+
 mod esp32;
 mod esp32c2;
 mod esp32c3;
@@ -187,7 +190,10 @@ impl Esp32Params {
                 Type::App,
                 SubType::App(AppType::Factory),
                 self.app_addr,
-                flash_size.map_or(self.app_size, |size| size - self.app_addr),
+                core::cmp::min(
+                    flash_size.map_or(self.app_size, |size| size - self.app_addr),
+                    MAX_PARTITION_SIZE,
+                ),
                 false,
             ),
         ])


SergioGasquez (approved):
LGTM! But, cant test it because I dont think I have a >16MB flash chip

maxwase:
Tested this on my ESP32-S3 32MB flash, works fine
I had a bootloop only when I explicitly set >16MB partition, maybe we should add a warning for it?
In 2 paces:  `flash` and `partition-table` sub-commands

jessebraham (approved):
LGTM, thanks!

PR #399, opened 2023-05-05
Fix Windows installation

SergioGasquez:
- Added [`resolver = "2"`](https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2) to the workspace `Cargo.toml`
- `cargo` dependency uses the feature `vendored-openssl` only on Unix systems
- Bump `cargo` version
  - This required bumping the MSRV to `1.65`
  
Those changes seem to make the installation succeed in all OSs (and does not require installing `perl` on Windows). I tested the installation in:
- MacOs
- Arch Linux
- Debian using Docker
- Windows 11

Any other test is appreciated: `cargo install cargo-espflash --git https://github.com/SergioGasquez/espflash --branch fix/windows-install`


jessebraham (approved):
Thanks for taking care of this, LGTM!

PR #407, opened 2023-05-19
espflash: add `--offline` flag

dimpolo:
This flag disables the update check.

Hope it's okay to add this.

Please let me know if there's anything I can improve :)

### DIFF
--- a/espflash/src/bin/espflash.rs
+++ b/espflash/src/bin/espflash.rs
@@ -26,6 +26,9 @@ use miette::{IntoDiagnostic, Result, WrapErr};
 pub struct Cli {
     #[command(subcommand)]
     subcommand: Commands,
+    /// Do not check for updates
+    #[arg(long, global = true)]
+    offline: bool,
 }
 
 #[derive(Debug, Subcommand)]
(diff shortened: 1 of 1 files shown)


SergioGasquez:
Hi! Do you mind elaborating the rationale behind this? Just did a small test and when I had no internet connection the check was just skipped:
```
# With internet connection
❯ espflash board-info
[2023-05-22T07:10:23Z INFO ] 🚀 A new version of espflash is available: v2.0.0-rc.3
✔ Use serial port '/dev/ttyS0'? · no

  × The serial port '/dev/ttyS0' could not be found


# With NO internet connection
❯ espflash board-info
✔ Use serial port '/dev/ttyS0'? · no

  × The serial port '/dev/ttyS0' could not be found
```

jessebraham:
I'm a little apprehensive of adding options to the base `Cli` struct like that. A big part of the reason we completely rewrote the command-line interface was to eliminate these options, as they tended to cause confusion if provided in the wrong location (ie. after the subcommand instead of before it).

On top of that, it doesn't really do much. As @SergioGasquez mentioned the check fails silently with no connection. If they don't trust this application the source code is obviously available for review.

dimpolo:
Those are valid points. For my niche I'm ok providing the client with my own build so I'll close this. Thanks for having a look

PR #412, opened 2023-05-26
Update to the latest version of `addr2line` and address breaking changes

jessebraham:


espflash/src/cli/monitor/symbols.rs
```diff
@@ -28,15 +29,19 @@ impl<'sym> Symbols<'sym> {
         //   3. if no function name is found, try to look it up in the object file
         //      directly
         //   4. return a demangled function name, if one was found
-        self.ctx
-            .find_frames(addr)
+        let mut frames = match self.ctx.find_frames(addr) {
+            LookupResult::Output(result) => result.unwrap(),
+            LookupResult::Load { .. } => unimplemented!(),
```
jessebraham, line 34: @MabezDev @bjoernQ is this safe? I stole it from the `addr2line` docs, and it has worked at least for my limited testing here. This is a little outside my realm of knowledge though, to be honest.

MabezDev: According to the addr2line docs:
> Operations that consult debug information may require additional files to be loaded if split DWARF is being used. This enum returns the result of the operation in the Break variant, or information about the split DWARF that is required and a continuation to invoke once it is available in the Continue variant.

I think given that we have single ELFs with all debug info, we don't need to cover this and can leave it as unimplemented.

SergioGasquez (approved):
LGTM!

PR #413, opened 2023-05-29
Do not require the `--partition-table` argument when erasing partitions

### LINKED ISSUE #390: cargo-espflash's `--erase-part` option requires `--partition-table` even when the partition table is set via metadata
When passing `--erase-part` to `cargo-espflash` for a package that has a partition table in package metadata (ie: in `Cargo.toml`'s `[package.metadata.espflash]` section), `cargo-espflash` still complains with:

```
error: the following required arguments were not provided:
  --partition-table <FILE>
```

Instead of returning an error, `cargo-espflash` should use the partition table provided by the metadata when a partition erase is requested (if that metadata is present for the package used).

jessebraham:
Rather than perform error handling within `clap` for this, we just move the check (which already existed) into `espflash` instead, after any package metadata has been read. This will still return an error if no partition table was able to be found, however it should be able to pick up partition tables provided via cargo metadata just fine.

Closes #390

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -89,21 +89,10 @@ pub struct FlashArgs {
     #[arg(long, value_name = "FILE")]
     pub bootloader: Option<PathBuf>,
     /// Erase partitions by label
-    #[arg(
-        long,
-        requires = "partition_table",
-        value_name = "LABELS",
-        value_delimiter = ','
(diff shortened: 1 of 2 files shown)


SergioGasquez (approved):
Changes LGTM, tested it a bit, and it behaves as expected, thanks for fixing this!

PR #416, opened 2023-05-30
Add CHANGELOG

SergioGasquez:
- Add changelog
- Add CI check to enforce changelog modifications

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -104,3 +104,19 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       # `too_many_arguments` is relatively arbitrary
       - run: cargo clippy -- -A clippy::too_many_arguments -D warnings
+
+  changelog:
+    name: Changelog
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -104,3 +104,19 @@ jobs:
...
+
+  changelog:
+    name: Changelog
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+
+      - name: Check that changelog updated
+        uses: dangoslen/changelog-enforcer@v3
```
jessebraham, line 116: This action drives me crazy, it does not work consistently with forks. I'm fine with enforcing the `CHANGELOG` but I would prefer to use something else.

SergioGasquez: I've only found https://github.com/Zomzog/changelog-checker but it seems like is not actively developed (it still uses NodeJS12 and has a issue open for it with no action taken so far) so I will remove the check for the time being.

jessebraham: Okay, thanks for checking at least.

jessebraham (approved):
LGTM, thanks again

PR #417, opened 2023-05-30
Override the flash size in `Flasher` if provided via command-line argument

### LINKED ISSUE #396: Flash size question
I was previously using the old (stable) version of the tool (`1.7.0`) and ran into an issue where my program got too large for the app partition (which is apparently by default 2MB ?). I was already using the `-s 16MB` option to tell it to increase the flash size (for other reasons), and didn't think this was affecting the app partition size. However, upon updating the tool to the `2.0.0-rc.3` version, my problem of exceeding the app partition size appears to have disappeared, which seems to imply that it's now using the `-s 16mb` option as app partition size, or maybe the default partition is just larger now. In both versions of the tool, it says it could not detect flash size and defaults to `4MB`, and yet it also says that the app partition size is paradoxically `16MB`. Here's the output I get:

```
    Finished dev [optimized + debuginfo] target(s) in 0.17s
     Running `espflash flash -s 16mb --monitor target/xtensa-esp32s3-espidf/debug/netsblox-vm-esp32`
[2023-05-03T15:36:46Z INFO ] Serial port: '/dev/ttyACM0'
[2023-05-03T15:36:46Z INFO ] Connecting...
[2023-05-03T15:36:46Z INFO ] Using flash stub
[2023-05-03T15:36:46Z WARN ] Could not detect flash size (FlashID=0x3980C2, SizeID=0x39), defaulting to 4MB
Chip type:         esp32s3 (revision v0.1)
Crystal frequency: 40MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       68:b6:b3:3d:79:d4
App/part. size:    3,085,168/16,711,680 bytes, 18.46%
[00:00:00] [========================================]      13/13      0x0                                                                                                                               
[00:00:00] [========================================]       1/1       0x8000                                                                                                                            
[00:00:32] [========================================]    1716/1716    0x10000
[2023-05-03T15:37:20Z INFO ] Flashing has completed!
```

So my question is: what exactly does the `-s` option do? Is it just the total flash size like the `--help` info implies, or is it app partition size? And also, would the new version of the tool let/tell me if I was flashing a too-large app into an app partition that's not large enough?

jessebraham:
Closes #396 (I think..?)

@dragazo @MabezDev please confirm this fixes the issue, but as far as I can tell we just were never passing the command-line argument to the `Flasher` struct when it was provided.

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -158,9 +158,17 @@ fn flash(args: FlashArgs, config: &Config) -> Result<()> {
     let cargo_config = CargoConfig::load(&metadata.workspace_root, &metadata.package_root);
 
     let mut flasher = connect(&args.connect_args, config)?;
+
+    // If the user has provided a flash size via a command-line argument, we'll
+    // override the detected (or default) value with this.
+    if let Some(flash_size) = args.build_args.flash_config_args.flash_size {
+        flasher.set_flash_size(flash_size);
+    }
+
     let chip = flasher.chip();
     let target = chip.into_target();
     let target_xtal_freq = target.crystal_freq(flasher.connection())?;
+
     flasher.disable_watchdog()?;
 
     let build_ctx =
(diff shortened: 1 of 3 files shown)


SergioGasquez (approved):
LGTM, thanks for the fix!

MabezDev (approved):
LGTM!

PR #418, opened 2023-05-30
Downgrade `crossterm` to `0.25.0`

jessebraham:
Computers were a mistake

(using `0.26.x` somehow breaks things on Windows in weird circumstances)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -451,7 +451,7 @@ version = "6.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6e7b787b0dc42e8111badfdbe4c3059158ccb2db8780352fa1b01e8ccf45cc4d"
 dependencies = [
- "crossterm 0.25.0",
+ "crossterm",
  "strum",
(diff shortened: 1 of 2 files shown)


bjoernQ:
I tried 0.27.0 today and it still gives us problems on Windows.

For the records: One (or probably the) issue is replicated as follows
- create a batch file flashing and monitoring at least two ELF files
- run the batch file
- see the first application running
- press CTRL-C
- next ELF will get flashed fine but monitoring will immediately end

There is a workaround which probably would have even worked in 0.26.x

`src\cli\monitor\mod.rs:160`
```
                        KeyCode::Char('c') => {
                            poll(Duration::from_millis(100)).ok();
                            break;
                        },
```

Apparently since 0.26.0 the event somehow isn't consumed by `read()` and it reads the same again in the next run. Don't know the details t.b.h.

It's interesting that the event doesn't get picked up during flashing (but I think it's because at that stage it's handled by the `ctrlc` crate)

With the above change it just polls again and might or might not consume a queued event (CTRL-C)

PR #419, opened 2023-05-31
Update the CI workflow

jessebraham:
- Extracted an action to set up an ARM build environment to make the workflow simpler to read
- All existing actions now run on `ubuntu-22.04`
- We now additionally check the binary build on `macos-12` and `windows-22`
- The CI workflow files have been renamed to be consistent with other repos in the org

I like that the workflows now both follow the same structure and are fairly easy to grok. The downside is that we have a fair bit of duplicated YAML, but given how infrequently these files need changing IMO it's a worthwhile sacrifice.

@SergioGasquez if you have any feedback or opinions on this please let me know!

I will update the `package` workflows in a later PR.


.github/actions/setup-arm/action.yml
```diff
@@ -0,0 +1,70 @@
...
+      uses: dtolnay/rust-toolchain@v1
+      with:
+        toolchain: ${{ inputs.toolchain }}
+        target: ${{ inputs.target }}
+        components: ${{ inputs.components }}
+
+    - uses: Swatinem/rust-cache@v2
+
+    - name: Install Cross-Compile Support
+      uses: cyberjunk/gha-ubuntu-cross@v2
```
SergioGasquez, line 36:
```suggestion
      uses: cyberjunk/gha-ubuntu-cross@v3
```

SergioGasquez (approved):
LGTM! The CI now looks way better!

PR #420, opened 2023-06-01
Add missing import when using `linux` and `musl`

jessebraham:
Found this when updating the `release` workflow, will have to update our CI to check `musl` as well (I'll take care of that in another PR)


SergioGasquez (approved):
LGTM! Thanks also for adding `musl `to the CI, so we avoid this in the future!

PR #421, opened 2023-06-02
Update the release workflow

jessebraham:
Successful test run:  
https://github.com/jessebraham/espflash/actions/runs/5158194931/jobs/9291556933

Successful test release:  
https://github.com/jessebraham/espflash/releases/tag/test-112


.github/actions/package-arm/action.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Compress
+      shell: bash
+      run: |
+        zip -j cargo-espflash-${{ inputs.target }}.zip target/${{ inputs.target }}/release/cargo-espflash
+        zip -j espflash-${{ inputs.target }}.zip target/${{ inputs.target }}/release/espflash
+
+    - uses: svenstaro/upload-release-action@v2
+      with:
+        repo_token: ${{ inputs.github_token }}
+        file: cargo-espflash-${{ inputs.target }}.zip
```
SergioGasquez, line 33: Not sure if it's a good idea, but I'll throw it anyway, we can use `*.zip` here and avoid using the action twice.

SergioGasquez (approved):
LGTM! I think it is way more elegant now. Maybe we could merge the two packages actions by adding a bunch of ifs, but not sure if it's worth.

MabezDev (approved):

PR #422, opened 2023-06-05
Additionally check the `x86_64-unknown-linux-musl` target in CI

jessebraham:
This should avoid needing to apply fixes like #420 in the future.


SergioGasquez (approved):
LGTM!

PR #438, opened 2023-06-27
Evaluate merging CI files

SergioGasquez:


jessebraham:
Thanks for looking into this! At first glance I think this actually looks quite good, I was expecting `ci.yml` to be messier than it is, so that's great.

I will do a proper review when I get a chance but I expect it's likely fine as-is.

jnross:
I like the idea of this one.  I was recently bouncing between the two CI workflows while debugging a CI issue and wondering why they were split.

jessebraham:
Sorry it's been awhile 😅 I think this looks fine, if you wouldn't mind updating it (our workflow has changed since this PR was opened, so we need to be careful to reflect those changes) then I think we can merge this.

SergioGasquez:
Closed in favor of #470

PR #441, opened 2023-07-05
Add help text for all subcommands

jessebraham:
I've kept these fairly minimal for the time being, with the expectation that we will expand on the help text as needed. With that said, would still like to get some feedback on what I have included.

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -89,7 +89,7 @@ enum Commands {
     ///
     /// Allows for conversion between formats via the '--to-csv' and
     /// '--to-binary' options, plus the ability to print a partition table
-    /// in tabluar format.
+    /// in tabular format.
     PartitionTable(PartitionTableArgs),
     /// Generate a binary application image and save it to a local disk
     ///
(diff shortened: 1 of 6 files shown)


cargo-espflash/src/main.rs
```diff
@@ -49,11 +54,49 @@ enum CargoSubcommand {
...
+    /// Convert partition tables between CSV and binary format
+    ///
+    /// Uses the ESP-IDF format for partition tables; please refer to the
+    /// ESP-IDF documentation for more information on this format:
+    ///
+    /// https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html
+    ///
+    /// Allows for conversion between formats via the '--to-csv' and
+    /// '--to-binary' options, plus the ability to print a partition table
+    /// in tabluar format.
```
bjoernQ, line 92: tabular

bjoernQ (approved):
LGTM

I think it's helpful for users and I like there are links to further information

SergioGasquez (approved):
LGTM! Just one small detail that we could add, as some people are confused  with this, is that `cargo-espflash` is able to detect if the app uses `esp-idf-sys` and if using it, will reuse the partition table and bootloader esp-idf while `espflash` uses a "hardcoded" bootloader and partition table.  But, feel free to disregard it as it's already on the README.

PR #455, opened 2023-08-05
Add interactive flag

jneem:
Adds an `--interactive` (or `-i`) flag that prevents auto-connection to a recognized device, even if it is the only recognized one.

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -52,6 +52,9 @@ pub struct ConnectArgs {
     /// Serial port connected to target device
     #[arg(short = 'p', long, env = "ESPFLASH_PORT")]
     pub port: Option<String>,
+    /// Require confirmation before auto-connecting to a recognized device.
+    #[arg(short = 'i', long)]
+    pub interactive: bool,
     /// DTR pin to use for the internal UART hardware. Uses BCM numbering.
     #[cfg(feature = "raspberry")]
     #[cfg_attr(feature = "raspberry", clap(long))]
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -53,8 +53,8 @@ pub struct ConnectArgs {
     #[arg(short = 'p', long, env = "ESPFLASH_PORT")]
     pub port: Option<String>,
     /// Require confirmation before auto-connecting to a recognized device.
-    #[arg(short = 'i', long)]
-    pub interactive: bool,
+    #[arg(short = 'c', long)]
+    pub confirm_port: bool,
     /// DTR pin to use for the internal UART hardware. Uses BCM numbering.
     #[cfg(feature = "raspberry")]
     #[cfg_attr(feature = "raspberry", clap(long))]
--- a/espflash/src/cli/serial.rs
+++ b/espflash/src/cli/serial.rs
@@ -40,7 +40,7 @@ pub fn get_serial_port_info(
     } else if let Some(serial) = &config.connection.serial {
         find_serial_port(&ports, serial)
     } else {
-        let (port, matches) = select_serial_port(ports, config, matches.interactive)?;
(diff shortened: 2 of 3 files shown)


espflash/src/cli/serial.rs
```diff
@@ -220,8 +221,12 @@ fn select_serial_port(
...
     {
         // There is a unique recognized device.
-        Ok(((*port).to_owned(), true))
-    } else if ports.len() > 1 {
+        if !force_interactive {
+            return Ok(((*port).to_owned(), true));
+        }
+    }
+
+    if ports.len() > 1 {
```
SergioGasquez, line 229: Is there any reason for this `if` to not be and `esle if` as before?

espflash/src/cli/mod.rs
```diff
@@ -47,6 +47,9 @@ pub struct ConnectArgs {
     /// Serial port connected to target device
     #[arg(short = 'p', long, env = "ESPFLASH_PORT")]
     pub port: Option<String>,
+    /// Require confirmation before auto-connecting to a recognized device.
+    #[arg(short = 'i', long)]
+    pub interactive: bool,
```
SergioGasquez, line 52: Do you mind moving these arguments to espflash/src/bin/espflash.rs and cargo-espflash/main.rs? It would require some duplication but, changing the CLI mod would result in a breaking change requiring us to bump the version to 3.0

jneem: Ah, good point about the compatibility issue. Unfortunately, I don't see how to get around it. I can add an `interactive` flag in various places, but eventually I need to pass the `interactive` flag to the (public) `cli::connect` function. I can't add it to the `ConnectArgs` or `Config` struct because they're both public, and adding it as an extra parameter would also break semver :disappointed: 

I guess I could deprecate `cli::connect` and add a new version that takes an `interactive` argument... but that starts to feel like overkill. I think the better solution would be to add some `#[non_exhaustive]` annotations so that arguments are easier to add in the future, but of course adding those annotations will itself be a breaking change.

SergioGasquez: Yes, as you mentioned, there are plans for using `#[non_exhaustive]` (https://github.com/esp-rs/espflash/issues/473), not sure if we can work around this, or we just need to wait to #473 to be resolved and add this to `v3`

SergioGasquez: Did a bit more of investigation and couldn't find a workaround for this, I'll add the `v3` milestone, and we can merge it once we start working towards `v3`.

SergioGasquez:
Hi! We've just started working towards version 3.0, do you mind rebasing your changes so we can merge them?

jneem:
Rebased. I noticed while testing it that the `-f` flag is used for both `flash_freq` and `log_format`, triggering a debug_assertion in clap.

espflash/src/cli/mod.rs
```diff
@@ -52,6 +52,9 @@ pub struct ConnectArgs {
...
```
SergioGasquez, line 57: Changes LGTM! The only thing that I have some doubts is the name of the argument. Maybe `interactive` is fine, other options that came to my mind `confirm-device`, `confirm`, `select-device`. Any input on this?

SergioGasquez (approved):

espflash/src/cli/mod.rs
jneem: I don't have a strong opinion about the name, so I'm happy to implement whichever you prefer. I guess I like "confirm" better than "select," and maybe also `confirm-port` is a reasonable option :man_shrugging:

SergioGasquez: Awesome, let's go for `confirm-prot`!

jneem: Done!

SergioGasquez:
One last thing! Do you mind updating the CHANGELOG?

jneem:
Sure thing

PR #461, opened 2023-08-21
Add `--target-app-partition` argument to flash command

jnross:


jnross:
Whoops, I didn't mean to close this.  I landed the change in our fork, but shouldn't have deleted the branch.

jnross:
I rebased this on top of #465 to get the CI fixes.  That should land before this one.

espflash/src/cli/mod.rs
```diff
@@ -106,6 +106,9 @@ pub struct FlashArgs {
     /// Path to a CSV file containing partition table
     #[arg(long, value_name = "FILE")]
     pub partition_table: Option<PathBuf>,
+    /// Label of target app partition
+    #[arg(long, value_name = "LABEL")]
+    pub target_app_partition: Option<String>,
```
SergioGasquez, line 111: Do you mind moving these arguments to `espflash/src/bin/espflash.rs` and `cargo-espflash/main.rs`? It would require some duplication but, changing the CLI mod would result in a breaking change requiring us to bump the version to `3.0`

jnross:
@SergioGasquez:  It's easy to move the argument changes out as you requested.  But there are changes to `pub` function in `cli/mod.rs`, `flasher/mod.rs`, and `idf_bootloader.rs`.  I can't see how to avoid those changes and implement this feature without duplicating a large amount of code.  Does that mean this change will need to wait until the SemVer major can be bumped to 3?  Or is there an additive non-breaking way to add these changes that I'm not seeing?

SergioGasquez:
> @SergioGasquez: It's easy to move the argument changes out as you requested. But there are changes to `pub` function in `cli/mod.rs`, `flasher/mod.rs`, and `idf_bootloader.rs`. I can't see how to avoid those changes and implement this feature without duplicating a large amount of code. Does that mean this change will need to wait until the SemVer major can be bumped to 3? Or is there an additive non-breaking way to add these changes that I'm not seeing?

I don't think it's worth to duplicate such large amount of code, I'll mark it as `v3` and we can implement it once we work for such version

SergioGasquez:
Hi! We've just started working towards version 3.0, do you mind rebasing your changes so we can merge them?

jnross:
Yes, will do.

SergioGasquez (approved):
LGTM! Do you mind updating the `CHANGELOG.md`?

jnross:
> LGTM! Do you mind updating the `CHANGELOG.md`?

Done!

Issue #460, opened 2023-08-18
Add command to erase all flash


jnross:
In some of our workflows we're still using `esptool.py` for its `erase_flash` command.  Sometimes we can avoid this with the `--erase-parts` option on the `flash` subcommand, but it's awkward that we need to provide an ELF binary to flash when we're only trying to erase data.

I propose that an `erase` subcommand be added, with options to erase all flash, or to erase only named partitions.

PR #462, opened 2023-08-21
Add `erase-flash`, `erase-region`, and `erase-parts` subcommands

jnross:


jnross:
Whoops, I didn't mean to close this PR.  I landed the branch in our fork, but shouldn't have deleted the branch.

jnross:
I rebased this on top of #460.  That should land first before this one.

SergioGasquez (commented):
Same as #461, we want to keep the CLI as it is (for the moment) to avoid having to bump SemVer major. This will be later addressed in https://github.com/esp-rs/espflash/issues/473

jnross:
Thanks for the comments here and in #461.  I'll make these changes to avoid the need for the SemVer bump.

espflash/src/bin/espflash.rs
```diff
@@ -156,6 +204,36 @@ fn main() -> Result<()> {
     }
 }
 
+fn erase_flash(args: EraseFlashArgs, config: &Config) -> Result<()> {
+    info!("Erasing Flash...");
+    let mut flash = connect(&args.connect_args, config)?;
+    flash.erase_flash()?;
+
```
SergioGasquez, line 211: Is there any reason not to `flash.connection().reset()` here as we do in `erase_parts` and `erase_region`?

SergioGasquez (commented):
I left some minor comments and also have a few suggestions:
- Can we implement the subcommands also for `cargo-espflash`?
  - Note that we don't require `partition_table` for `erase-parts` in `cargo-espflash` as it can be obtained from the [package metadada](https://github.com/esp-rs/espflash/tree/main/cargo-espflash#package-metadata), or it will result in a similar issue to [#390](https://github.com/esp-rs/espflash/issues/390)
- [EraseFlash and EraseRegion are only available when using stub](https://github.com/espressif/esptool/blob/67a91cbfef54f281212951b8226583ba3c1d0a85/esptool/loader.py#L217)
    - It would be great if we could check if we are using a stub, if not, throw an informative error
        - esptool does not have such check or I haven't found it
        - This might not be possible to do without breaking semver. If that's the case we can just open an issue for the future
    -  At the moment, if you flash with no-stub and then do an `erase-flash` and you get `invalid header: 0xffffffff`

jnross:
Thanks @SergioGasquez for taking a detailed look and for these great suggestions.  I'll work them into my branch.

espflash/src/bin/espflash.rs
jnross: I do have a `reset()` in place after `erase_flash`, but it's in `flasher/mod.rs` instead of here.  Your comment here makes me realize that's inconsistent with the other operations.  I'll move this into `espflash.rs` as suggested.

jnross:
@SergioGasquez :  In this PR as originally authored, there were changes to `cli/mod.rs`, but they were additive only.  There were new `pub` structs and functions added, but none of the existing interface was altered or broken.  Isn't that permitted in a minor version bump in SemVer?  I'd working on adding these subcommands to `cargo-espflash`, but without making additive changes to `cli/mod.rs` there's quite a bit of duplicated code required.

jnross:
@SergioGasquez :  I adopted all of your suggestions.  Then I went about adding support for these subcommands to `cargo-espflash`, and found that I would have to duplicate several new structs and new functions in both `espflash/src/bin/espflash.rs` and `cargo-espflash/src/main.rs`.  Since these structs and functions are purely additive this shouldn't require a major SemVer bump, only a minor one.

I'd ask that you take a close look at the `erase-parts` implementation in `cargo-espflash`.  I wasn't sure if I should require a build there, or whether I should fail if a cargo package isn't found in the current directory or per the `--package` argument, as long as an alternative `--partition-table` is provided.

SergioGasquez (commented):
Just a small comment to avoid using `bail!` and use a defined Error that we can reuse

jnross:
@SergioGasquez: Thanks for catching that, you're quite right I should have defined a single error for these cases.  Fixed in https://github.com/esp-rs/espflash/pull/462/commits/22d792ada3795ae0ce284b7e530aba89856b9de3

espflash/src/error.rs
```diff
@@ -83,6 +83,13 @@ pub enum Error {
     )]
     NoSerial,
 
+    #[error("Cannot erase flash without the RAM stub")]
+    #[diagnostic(
+        code(espflash::stub_required_to_erase_flash),
+        help("Don't use the `--no-ram-stub` option with `erase` commands")
+    )]
+    StubRequiredToEraseFlash,
```
SergioGasquez, line 91:
```suggestion
    #[error("Erase commands require using the RAM stub")]
    #[diagnostic(
        code(espflash::stub_required_to_erase_flash),
        help("Don't use the `--no-stub` option with erase commands")
    )]
    StubRequiredToEraseFlash,
```

SergioGasquez (commented):
Thanks for the quick fixes! One last detail!

espflash/src/error.rs
jnross: 👍

SergioGasquez (approved):
LGTM! Thanks for the contribution and for addressing all the changes!

jnross:
Thanks for your stewardship and excellent suggestions here.  I'm happy to see these changes get integrated.

PR #465, opened 2023-08-25
Fix CI checks

jnross:
I investigated the CI failures and discovered it was a non-zero exit code from a call to `apt-get update` that was causing the workflow to fail.  In order to fix this, I had to fork the `gha-ubuntu-cross` repo and apply the fix there.  I'm working to get this fix upstreamed in https://github.com/cyberjunk/gha-ubuntu-cross/pull/6.  Once it is, we can revert this reference back to the original repo.

I also fixed a clippy warning that was added 1.72.0.  https://rust-lang.github.io/rust-clippy/master/index.html#/single_range_in_vec_init  The page says "this is almost always incorrect", but it seems clear that this is quite intentional in our case here.  So I silenced the warning by adding an `#[allow` annotation.

### DIFF
--- a/.github/actions/setup-arm/action.yml
+++ b/.github/actions/setup-arm/action.yml
@@ -33,7 +33,7 @@ runs:
     - uses: Swatinem/rust-cache@v2
 
     - name: Install Cross-Compile Support
-      uses: cyberjunk/gha-ubuntu-cross@v3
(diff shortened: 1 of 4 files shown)


jnross:
I updated this PR with an improved fix for `gha-ubuntu-cross` and included the fixes from #467 so checks will pass.

PR #466, opened 2023-08-26
Print esp-println generated defmt messages

### LINKED ISSUE #316: add defmt support
Adding a flag that parses defmt data would be nice for using with this tool, I currently just pipe the monitor to ``defmt-print``.
I think defmt could also be detected instead of adding a flag but idk which is preferred.

bugadani:
Closes #316 

Counterpart of https://github.com/esp-rs/esp-println/pull/48 as it relies on the injected framing bytes. Not sure how to best lift this curse yet.

This PR adds a new compile-time `defmt` flag that enables defmt support in the binaries and the library. For this, the PR raises MSRV to 1.70. By default, the `defmt` feature is enabled. defmt detection is done by parsing the elf file.

The PR refactors serial port input handling so library users can define their own processing if they so choose. The default processor is able to recognize and parse frames generated by esp-println. Library users can disable this processing by using the `Serial` input processor or by disabling the `defmt` feature (which internally does the same besides disabling the dependencies).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Read esp-println generated defmt messages (#466)
+
 ### Fixed
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,9 +30,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Update dependencies to their latest versions (#482)
-- **breaking** Bumped MSRV to 1.70.0
-
-### Removed
 
 ## [2.0.1] - 2023-07-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Update dependencies to their latest versions (#482)
+- **breaking** Bumped MSRV to 1.70.0
+
+### Removed
 
 ## [2.0.1] - 2023-07-13
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,10 +30,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
  "cpp_demangle",
- "fallible-iterator",
- "gimli",
+ "fallible-iterator 0.3.0",
+ "gimli 0.28.0",
  "memmap2",
- "object",
+ "object 0.32.1",
  "rustc-demangle",
  "smallvec",
 ]
@@ -163,7 +163,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "miniz_oxide",
- "object",
+ "object 0.32.1",
  "rustc-demangle",
 ]
 
@@ -543,6 +543,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "acbf1af155f9b9ef647e42cdc158db4b64a1b61f743629225fde6f3e0be2a7c7"
 
 [[package]]
+name = "colored"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2674ec482fbc38012cf31e6c42ba0177b431a0cb6f15fe40efa5aab1bda516f6"
+dependencies = [
+ "is-terminal",
+ "lazy_static",
+ "windows-sys 0.48.0",
+]
+
+[[package]]
 name = "comfy-table"
 version = "7.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


bugadani:
Current status: waiting for other PRs to land before moving forward, to give priority to fixes.

I'm undecided whether I should add a `--defmt-flavour=<raw|esp (default)>` switch to support defmt sources other than esp-println.

bugadani:
Oh come on... @jessebraham do you think we can bump MSRV for defmt, either selectively or otherwise?

MabezDev:
Dependency hell is real :D. Are these not breaking changes from the library(s)? Surely bumping `rust-version` indicates a breaking change? Idk but that's really annoying, not sure what to do here other than bump out MSRV too.

bugadani:
In this case, defmt is a new dependency so it's probably not a case of a patch-with-breaking-MSRV-bump. CI just wasn't passing for the last few weeks and I didn't notice this PR requires a bump.

jessebraham:
Strictly speaking, I don't think this situation is _technically_ covered by SemVer itself.

There is, however, some documentation in The Cargo Book which states that this is allowed if we bump the `minor` version:  
https://doc.rust-lang.org/cargo/reference/semver.html#possibly-breaking-changing-the-minimum-version-of-rust-required

I have no issue bumping the MSRV as long as it's not the current `stable` version. I'd prefer to have at least a few versions back, if possible.

MabezDev:
Oh yes, of course. In that case we either bump the MSRV and be done with it, or we say we don't make any MSRV guarentees behind a defmt feature (that we'd need to introduce). 

Imo bumping MSRV is okay, it realistically only affects library users which (publically) is just me: https://crates.io/crates/espflash/reverse_dependencies :D.

bugadani:
> We would need to do a minor release to include these changes anyway, so no harm done.

Oh hell please do not start following this madness.

jessebraham:
Okay, well we can't merge this then and can't update a number of dependencies ever. So I'm not sure that's preferable.

bugadani:
Have you guys a recommendation how I could featurize defmt? I think it would be nice to ship the binaries with it enabled, but I also don't want to force defmt on libraries. I believe I can extract the code itself to make parsing in monitor opt-in, hide that behind a feature flag and call the library part done. But for the binaries, I think it would be better to have the feature default on, as opposed to passing a flag in CI.

I also don't know if the feature should be controllable in the binaries, via a --defmt switch or something similar. I tend to think it'd just complicate the implementation without much value, but I'd love to hear your take.

bugadani:
I'd consider this done now. I've been using this PR for the last few days/weeks without any issue, though I'm a bit unsure if defmt-by-default is the polite way to handle the issue.

Because of the MSRV bump I'd suggest a major release (3.0.0) at least for the library, but considering the low number of public dependent maybe not doing this isn't the end of the world. For binaries, semver is a lot less relevant.

The defmt feature is enabled by default mainly because I wanted to avoid editing the CI files for this.
I considered adding a runtime switch but the implementation complexity doesn't seem to be worth it.

Let me know if you'd like to instead keep defmt off by default. Doing that would allow keeping MSRV low, which would avoid the semver-related headaches. From-source installations could still enable the feature, however one downside is that it's not very feasible to install from source on a Raspberry Pi.

CHANGELOG.md
```diff
@@ -28,6 +30,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Update dependencies to their latest versions (#482)
+- **breaking** Bumped MSRV to 1.70.0
```
SergioGasquez, line 33:
```suggestion
- **breaking** Bumped MSRV to 1.70.0
```
This is no longer needed, the version is `1.70.0` already

CHANGELOG.md
```diff
@@ -28,6 +30,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Removed
```
SergioGasquez, line 35:
```suggestion
```

SergioGasquez:
Just tried testing your branch and couldnt monitor a `esp-tempalte` project (tried c3 and s3)
```
cargo install --force espflash --git https://github.com/bugadani/espflash --branch defmt
cargo generate esp-rs/esp-template --name test -d mcu=esp32c3 -d advanced=false
cd test
cargo r -r
```
Also tried updating the `esp-println`: `esp-println = { git = "https://github.com/esp-rs/esp-println", features = ["esp32c3", "defmt"] }`

But always getting the following:
```
Hello world!
Error:   × Main thread panicked.
                                  ├─▶ at /rustc/bf9a1c8a193fc373897196321215794c8bebbeec/library/std/src/io/mod.rs:1630:36
                                                                                                                            ╰─▶ range start index 78 out of range for slice of length 77
                                                                                                                                                                                          help: set the `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Not sure if its failing, or I am doing something wrong?

bugadani:
Interesting find. I have experienced a few panics myself, but so far the impl has been stable enough for me. The last commit has resolved a similar issue for me but maybe newlines weren't the root cause. I'll investigate.

bugadani:
Okay so the panic happened like this:
 - `println` printed a non-defmt frame
 - The frame delimiter has correctly handled this frame as Raw
 - The contents were written correctly to `ResolvingPrinter`. The frame ended with a `\n` character.
 - The `\n` was expanded to `\r\n` using `Normalized`, increasing the length of the string by one
 - `ResolvingPrinter::write` returned the length of the normalized string instead of the input's, causing `write_all` to incorrectly index during the next iteration, causing the panic.

SergioGasquez:
I did some further testing and seems to be working fine! Just one question/discussion, shall we have the `defmt` feature enable by default? @jessebraham @MabezDev @bjoernQ

bjoernQ:
> I did some further testing and seems to be working fine! Just one question/discussion, shall we have the `defmt` feature enable by default? @jessebraham @MabezDev @bjoernQ

I guess it shouldn't be a default. While it's not very likely that someone tries to use the framing marker by accident it's not impossible. Additionally, I think that a user who wants to use defmt is probably a more advanced user and it should be okay if they need some extra steps in enabling it. But maybe I'm thinking too cautiously

MabezDev:
I agree with @bjoernQ's points. I think that if we get a lot of requests to enable it by default then we can re-assess later.

MabezDev (approved):

PR #467, opened 2023-08-27
Avoid updating clap to newer minor versions

bugadani:
`clap_complete` 4.4.0 raised MSRV to 1.70, breaking CI due to espflash only specifying 1.65.

As per https://github.com/clap-rs/clap/issues/5087 the clap project treats MSRV breaks as minor changes. While this is also what cargo suggests, it's simpler to lock clap to patch releases than to keep bumping MSRV every 6 weeks.

I've also resolved two clippy lint issues, at least one of them affecting CI.

Hopefully, along with #465 this will fix the current CI woes.


bugadani:
Closing in favour of #465

PR #477, opened 2023-09-18
Switch to binrw from unmaintained binread

tommy-gilligan:


jessebraham:
Thanks for the PR. Sorry, we had some CI issues which should now be resolved. If you wouldn't mind rebasing then CI should hopefully be green here after.

jessebraham (approved):
LGTM, thanks!

PR #482, opened 2023-09-29
Update dependencies

SergioGasquez:
- Update dependencies
- Update CHANGELOG
- Bump `espflash` and `cargo-espflash`

### DIFF
--- a/cargo-espflash/Cargo.toml
+++ b/cargo-espflash/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-espflash"
-version = "2.0.2-dev"
+version = "2.0.2"
 edition = "2021"
 rust-version = "1.65"
 description = "Cargo subcommand for flashing Espressif devices"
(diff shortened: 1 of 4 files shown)


MabezDev:
I forget which library it is, but cc: @bjoernQ for the windows terminal issue just in case.

bjoernQ:
The problematic crate on Windows is `crossterm` - since this leaves it on 0.25.0 it should be fine.

We could update it with a workaround like described in https://github.com/esp-rs/espflash/pull/418#issuecomment-1717302410 .... but that workaround probably needs to be tested on other OSs or only get applied for Windows builds

But maybe leave it out of this PR and have a separate one for that

cargo-espflash/Cargo.toml
```diff
@@ -26,15 +26,15 @@ pkg-fmt = "zip"
 
 [dependencies]
 cargo_metadata = "0.18.0"
-clap = { version = "4.3.*", features = ["derive", "wrap_help"] }
+clap = { version = "4.4.6", features = ["derive", "wrap_help"] }
 env_logger = "0.10.0"
 esp-idf-part = "0.4.1"
-espflash = { version = "=2.0.2", path = "../espflash" }
+espflash = { version = "=2.1.0", path = "../espflash" }
```
jessebraham, line 32:
```suggestion
espflash = { version = "2.1.0", path = "../espflash" }
```

Let's not pin this please, just in case we end up publishing a patch release of `espflash` that are lib-only changes.

(I think pinning this dependency is just a relic of the pre-2.0 days? I can't really remember why we're doing that now...)

jessebraham (commented):
Once little thing but otherwise LGTM, thanks for taking care of this!

jessebraham (approved):

MabezDev (approved):

PR #486, opened 2023-10-05
Add publish to Crates.io step in the release workflow.

SergioGasquez:
We would also need to set `CARGO_API_KEY` in the repository secrets with the Crates.io key

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,6 @@ jobs:
 
   publish-cratesio:
     name: Publishing to Crates.io
-    needs: release
     runs-on: ubuntu-20.04
     steps:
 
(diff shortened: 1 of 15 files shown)


.github/workflows/release.yml
```diff
@@ -53,3 +53,33 @@ jobs:
           features: ${{ matrix.platform.features }}
           target: ${{ matrix.platform.target }}
           runs_on: ${{ matrix.platform.os }}
+
+  publish-cratesio:
+    name: Publishing to Crates.io
+    needs: release
```
jessebraham, line 59: Don't think this is necessary

jessebraham (approved):
The commit history is a little messed up, so I think this needs a rebase or something. Changes look fine as far as I can tell, though.

SergioGasquez:
Closed in favor of #492

PR #489, opened 2023-10-11
Send SpiSetParams command

SergioGasquez:


espflash/src/flasher/mod.rs
```diff
@@ -229,6 +229,43 @@ impl FromStr for FlashSize {
     }
 }
 
+/// Parameters of the attached SPI flash chip (sizes, etc).
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct SpiSetParams {
```
jessebraham, line 235: Do we actually need to expose this publicly?

jessebraham (approved):
LGTM, thanks for fixing this!

PR #490, opened 2023-10-16
Add defmt feature to cargo-espflash

bugadani:
This was missed when I undefaulted the feature


SergioGasquez (approved):
LGTM! Thanks for the fix

PR #492, opened 2023-10-17
Add publish to crates.io job

SergioGasquez:


SergioGasquez:
> LGTM! Have we set the `CARGO_API_KEY` secret yet, or does that still need to be done?

Just added the secret!

PR #493, opened 2023-10-18
Add log-format argument

SergioGasquez:
~Add `-d/--defmt` flag for `monitor` and `flash` subcommands.~
Add `-f/--log-format` argument for `monitor` and `flash` subcommands.

cc @bugadani since you added this feature

~Draft until properly tested.~ Did some testing and seems to be working fine! 

We still need to figure out if we want opt-in or opt-out of defmt

### DIFF
--- a/espflash/Cargo.toml
+++ b/espflash/Cargo.toml
@@ -40,8 +40,8 @@ clap_complete = { version = "4.4.3", optional = true }
 comfy-table = { version = "7.0.1", optional = true }
 crossterm = { version = "0.25.0", optional = true } # 0.26.x causes issues on Windows
 ctrlc = { version = "3.4.0", optional = true }
-defmt-decoder = { version = "=0.3.8", features = ["unstable"], optional = true }
-defmt-parser = { version = "=0.3.3", features = ["unstable"], optional = true }
+defmt-decoder = { version = "0.3.9", features = ["unstable"], optional = true }
+defmt-parser = { version = "0.3.3", features = ["unstable"], optional = true }
 dialoguer = { version = "0.10.4", optional = true }
 directories = { version = "5.0.1", optional = true }
 env_logger = { version = "0.10.0", optional = true }
@@ -74,6 +74,8 @@ cli = [
     "dep:comfy-table",
     "dep:crossterm",
     "dep:ctrlc",
+    "dep:defmt-decoder",
+    "dep:defmt-parser",
     "dep:dialoguer",
     "dep:directories",
     "dep:env_logger",
@@ -84,8 +86,5 @@ cli = [
     "dep:regex",
     "dep:update-informer",
 ]
-defmt = [
-    "dep:defmt-decoder",
-    "dep:defmt-parser",
-]
+
 raspberry = ["dep:rppal"]
--- a/espflash/Cargo.toml
+++ b/espflash/Cargo.toml
@@ -40,7 +40,7 @@ clap_complete = { version = "4.4.3", optional = true }
 comfy-table = { version = "7.0.1", optional = true }
 crossterm = { version = "0.25.0", optional = true } # 0.26.x causes issues on Windows
 ctrlc = { version = "3.4.0", optional = true }
-# defmt dependencies are pinned since defmt does not guarantees MSRV even for patch releases
+# defmt dependencies are pinned since defmt does not guarantee MSRV even for patch releases
 defmt-decoder = { version = "=0.3.9", features = ["unstable"], optional = true }
 defmt-parser = { version = "=0.3.3", features = ["unstable"], optional = true }
 dialoguer = { version = "0.10.4", optional = true }
--- a/espflash/Cargo.toml
+++ b/espflash/Cargo.toml
@@ -40,8 +40,9 @@ clap_complete = { version = "4.4.3", optional = true }
 comfy-table = { version = "7.0.1", optional = true }
 crossterm = { version = "0.25.0", optional = true } # 0.26.x causes issues on Windows
 ctrlc = { version = "3.4.0", optional = true }
-defmt-decoder = { version = "0.3.9", features = ["unstable"], optional = true }
-defmt-parser = { version = "0.3.3", features = ["unstable"], optional = true }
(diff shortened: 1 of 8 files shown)


bugadani:
Since the framing should mostly be compatible with the classic string streams, I'd probably suggest this to be opt out instead of opt in. It's probably much more annoying to enable defmt constantly for those who want it, than to disable for the few who have some custom binary in their logs.

SergioGasquez:
To be honest, I don't have a preference if it should be opt-in or opt-out, any suggestion @esp-rs/espressif?

bugadani:
As an idea, maybe we could enable printing location info with `--log-format=defmt-verbose` or similar option, so I think I shall update my opinion and vote for that format :)

espflash/src/cli/monitor/mod.rs
```diff
@@ -33,6 +34,17 @@ pub mod parser;
...
 
+#[cfg_attr(feature = "cli", derive(clap::ValueEnum))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Display, EnumIter, EnumString, EnumVariantNames)]
+#[non_exhaustive]
+#[strum(serialize_all = "lowercase")]
+pub enum LogFormat {
+    /// defmt
+    Defmt,
+    /// serial
+    Serial,
```
MabezDev, line 45: We should make `Serial` the default option, not sure the best way to do this in clap so I'll leave it to you :D.

MabezDev (changes requested):
I think with the `--log-format` change we're ready to go. I much prefer this to have to build a whole different espflash, and we can support other formats in the future.

Just one comment about making serial the default :)

espflash/src/cli/monitor/mod.rs
bugadani: ^ I wrote this same comment and forgot to send it haha.

SergioGasquez: This should now be fixed with https://github.com/esp-rs/espflash/pull/493/commits/5273a4a8f1d14b6849029ea631f39115bb4f3ce4

MabezDev (approved):

espflash/Cargo.toml
```diff
@@ -40,8 +40,8 @@ clap_complete = { version = "4.4.3", optional = true }
 comfy-table = { version = "7.0.1", optional = true }
 crossterm = { version = "0.25.0", optional = true } # 0.26.x causes issues on Windows
 ctrlc = { version = "3.4.0", optional = true }
-defmt-decoder = { version = "=0.3.8", features = ["unstable"], optional = true }
-defmt-parser = { version = "=0.3.3", features = ["unstable"], optional = true }
+defmt-decoder = { version = "0.3.9", features = ["unstable"], optional = true }
+defmt-parser = { version = "0.3.3", features = ["unstable"], optional = true }
```
bugadani, line 44: One reason why these were pinned was that defmt does not guarantee MSRV even for patch releases. I would recommend keeping the dependencies pinned, otherwise you may find that CI can break without changing anything defmt-related.

bugadani (approved):

espflash/Cargo.toml
SergioGasquez: Oh! Thanks for the heads up, just fixed it!

espflash/Cargo.toml
```diff
@@ -40,8 +40,9 @@ clap_complete = { version = "4.4.3", optional = true }
 comfy-table = { version = "7.0.1", optional = true }
 crossterm = { version = "0.25.0", optional = true } # 0.26.x causes issues on Windows
 ctrlc = { version = "3.4.0", optional = true }
-defmt-decoder = { version = "0.3.9", features = ["unstable"], optional = true }
-defmt-parser = { version = "0.3.3", features = ["unstable"], optional = true }
+# defmt dependencies are pinned since defmt does not guarantees MSRV even for patch releases
```
bugadani, line 43:
```suggestion
# defmt dependencies are pinned since defmt does not guarantee MSRV even for patch releases
```

espflash/src/cli/mod.rs
```diff
@@ -134,6 +138,9 @@ pub struct FlashArgs {
     /// Load the application to RAM instead of Flash
     #[arg(long)]
     pub ram: bool,
+    /// Logging format.
+    #[arg(long, short = 'f', default_value = "serial", requires = "monitor")]
+    pub log_format: LogFormat,
```
SergioGasquez, line 143: As @jessebraham pointed, this is a major breaking change, tomorrow I will try to check if there is any workaround for this.

bugadani: This is breaking for library users because a cargo feature got removed, but other than that, do the version of the binary artifacts matter much? Am I missing some other breakage?

Is there much reason for keeping the version bump minor and spend time trying to work around the breakage? Version numbers are supposedly cheap.

SergioGasquez: I couldn't find a way to workaround this in a non-breaking change manner. 

> This is breaking for library users because a cargo feature got removed, but other than that, do the version of the binary artifacts matter much? Am I missing some other breakage?
> 
> Is there much reason for keeping the version bump minor and spend time trying to work around the breakage? Version numbers are supposedly cheap.

I guess we should try to be SemVer compatible to avoid breaking our dependents applications, that being said we could evaluate if we should include it for a minor release anyway. Cc @esp-rs/espressif

SergioGasquez: We decided to start working towards v3, so this PR can now be merged

jessebraham (approved):
LGTM

espflash/src/cli/monitor/mod.rs
```diff
@@ -102,7 +108,16 @@ pub fn monitor_with<L: InputParser>(
             err => err,
         }?;
 
-        parser.feed(&buff[0..read_count], &mut stdout);
+        match log_format {
```
bugadani, line 111: I'm not sure why I didn't notice this but this isn't correct. The defmt parser is stateful, we can't just create a new one every time we have read some bytes.

PR #499, opened 2023-11-06
Update log-format flag and merge monitor methods

SergioGasquez:
- The flag `f` was already being used by [`flash-freq`](https://github.com/esp-rs/espflash/blob/39fbdacebaaa74a88e5d447a118f188be7d2e9bb/espflash/src/cli/mod.rs#L104), updated log-format to use `L`
- Since the PR was intended to be released in a minor version, we created `monitor_with` method to avoid changing `monitor`, now it can be modified


jessebraham (approved):
LGTM, thanks!

PR #501, opened 2023-11-09
flasher: add write_bins_to_flash.

Dirbaio:
This makes flashing a bit faster vs calling `write_bin_to_flash` because it
does the "spi attach" only once.


SergioGasquez (approved):

PR #503, opened 2023-11-13
Fix defmt losing bytes

### LINKED ISSUE #502: `espflash` fails to parse some defmt frames
From espflash monitor:
image

With `defmt-print`
image

bugadani:
Fixes #502

cc https://github.com/esp-rs/espflash/pull/493#discussion_r1390694672


MabezDev (approved):

SergioGasquez (approved):
Thanks for the fix!

PR #512, opened 2023-11-17
API improvements

SergioGasquez:
- Created `non_exhaustinve` `FlashData` and `FlashSettings` structs to reduce number of input arguments in the following functions:
  - `save_elf_as_image`
  - `flash_elf_image`
  - `load_elf_to_flash`
  - `get_flash_image` for all targets
- Using `FlashData` avoids duplicating code of parsing the bootloader and partition table file that was present in several places.

Not completely convinced on the `FlashData` name.


jessebraham:
Thanks for taking care of this! I took a quick look and I think it looks pretty good overall, but I have a couple features I'd like to continue developing to see how they affect this.

Will try to get this taken care of sooner than later and update here when able.

SergioGasquez:
Rebased on top of #516! But now Clippy is back at screaming that [a function has too many args (8/7)](https://github.com/esp-rs/espflash/blob/515707ba8db3dd4ba5257f41e9a44631e2a19652/espflash/src/image_format/idf_bootloader.rs#L34-L43) :(

SergioGasquez:
Rebased on top of https://github.com/esp-rs/espflash/pull/525! Clippy is now screaming louder that [a function has too many args (9/7)](https://github.com/esp-rs/espflash/blob/9133c083e4c4601ca823045e65108ba1de899cb5/espflash/src/image_format/idf_bootloader.rs#L33-L44)

jessebraham (approved):
Sorry for taking so long to review this, but I think it's looking pretty good. Thanks again for taking care of this.

We should still do another review of the public API before the release, but no point holding things up any longer here.

PR #515, opened 2023-11-21
Require --elf arg for --log-format arg in monitor subcommand

SergioGasquez:
If we use `monitor --log-format defmt`  but don't provide the elf file, we won't be able to interpret the logs.


jessebraham (approved):
Good catch!

PR #516, opened 2023-11-21
Add handling for custom partition table offsets

### LINKED ISSUE #458: Support flashing partition table at custom location
In our use of the ESP32-S3, we are enabling the secure boot option in the bootloader which increases the size of the bootloader such that it overflows into default location of the partition table at 0x8000.  We've had to move our partition table to 0xA000 via the `sdkconfig`, but this appears to break our usage of `cargo-espflash flash`.

What would be the best way to support this change in the partition table location?  A few ideas:
1. Detect the partition table location via the `sdkconfig` or `esp-idf-sys` output
2. Allow the partition table location to be explicitly specified via command-line option to `cargo-espflash` and `espflash`
3. Add a partition type to the `.csv` file that allows the location and size of the partition table in flash to be define just like other partitions.  (Maybe the bootloader should be added to the csv as well.  We know the bootloader must be located at 0x0, but it might be useful to specify the size of the bootloader in the .csv)

thetek42:
- Add a `--partition-table-offset` flag that corresponds to esp-idf's `CONFIG_PARTITION_TABLE_OFFSET`
  - If the flag is specified, the partition table will be flashed to the given location.
  - If the flag is omitted, the partition table location will be determined automatically from the partition table by subtracting 0x1000 from the first partition, which should be the most common case.
- The location of the application is now automatically determined from the partition table instead of the hard-coded 0x10000.

Closes #458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add --target-app-partition argument to flash command (#461)
 - Add --confirm-port argument to flash command (#455)
 - Add --chip argument for flash and write-bin commands (#514)
+- Add --partition-table-offset argument for specifying the partition table offset (#516)
 
 ### Fixed
 
 - Fixed printing panic backtraces when using `esp-println` and `defmt` (#496)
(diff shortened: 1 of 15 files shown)


jessebraham (approved):
LGTM, thanks!

PR #517, opened 2023-11-22
Mark cli args as `non_exhaustive` and warn that `cli` mod doesnt provide SemVer guarantees.

SergioGasquez:
- Add warnings that `cli` mod does not provide SemVer guarantees
- Mark cli args as non_exhaustive


bjoernQ (approved):
LGTM

Pretty much straight forward. At first, I thought why not make the cli module non-public  ... but yes - there is still cargo-espflash

PR #518, opened 2023-11-22
Use `esp-flasher-stub` stubs

SergioGasquez:
- Replace the current C stubs for `esp-flasher-stub` generated stubs.
  - `esp-flasher-stub` does not generate ESP8266 stubs so I kept the C based stub for it


MabezDev:
Given how untested the stubs are, this has the potential to induce a bunch of bug reports. Do we want to add them behind a flag first?

SergioGasquez:
> Given how untested the stubs are, this has the potential to induce a bunch of bug reports. Do we want to add them behind a flag first?

I initially did that, but as Jesse suggested, not sure how many people will test it if it's behind a feature. Happy to revert the lats commit and keep it behind a flag, though.

MabezDev:
> I initially did that, but as Jesse suggested, not sure how many people will test it if it's behind a feature

True, presumably there will be a rc release for espflash 3.0, so I guess if many people run into issues with the rc we can revisit the decision.

MabezDev (approved):

jessebraham (approved):
LGTM, thanks!

Issue #450, opened 2023-07-24
Error: espflash::connection_failed  (ESP32 + ESP-Prog + JTAG)


systec-ms:
Good day, I got an `ESP32 D1 MINI`, which I can program via `espflash` over the serial (micro-usb). 
Additionally, I can flash the esp `blinky` example via the VSCode extension over the ESP-Prog JTAG connection (`⚡️ Flashed Successfully (JTag)`).

However, if I try to flash my rust code via `espflash` over the EPS-Prog JTAG connection, I run into `Error: espflash::connection_failed`.

```
     Running `espflash flash --monitor target/xtensa-esp32-espidf/debug/blinky`
[2023-07-24T14:45:47Z INFO ] Detected 3 serial ports
[2023-07-24T14:45:47Z INFO ] Ports which match a known common dev board are highlighted
[2023-07-24T14:45:47Z INFO ] Please select a port
✔ Remember this serial port for future use? · no
[2023-07-24T14:45:50Z INFO ] Serial port: '/dev/ttyUSB0'
[2023-07-24T14:45:50Z INFO ] Connecting...
[2023-07-24T14:45:51Z INFO ] Unable to connect, retrying with extra delay...
[2023-07-24T14:45:52Z INFO ] Unable to connect, retrying with default delay...
[2023-07-24T14:45:53Z INFO ] Unable to connect, retrying with extra delay...
[2023-07-24T14:45:54Z INFO ] Unable to connect, retrying with default delay...
[2023-07-24T14:45:54Z INFO ] Unable to connect, retrying with extra delay...
[2023-07-24T14:45:56Z INFO ] Unable to connect, retrying with default delay...
[2023-07-24T14:45:56Z INFO ] Unable to connect, retrying with extra delay...
Error: espflash::connection_failed
```

I also tried https://github.com/esp-rs/espflash/pull/387 as mentioned by https://github.com/esp-rs/espflash/issues/394 and https://github.com/esp-rs/espflash/issues/394 using
 `cargo install --git https://github.com/AVee/espflash --branch add_unix_tight_reset espflash` with the same result.

```
    Finished dev [optimized + debuginfo] target(s) in 31.24s
     Running `espflash flash --monitor target/xtensa-esp32-espidf/debug/blinky`
[2023-07-24T14:32:58Z INFO ] 🚀 A new version of espflash is available: v2.0.1
[2023-07-24T14:32:58Z INFO ] Detected 2 serial ports
[...]

PR #521, opened 2023-11-23
Flush buffer after writing command

bugadani:
Dropping `BufWriter` flushes the internal buffers, but doesn't flush the inner stream. Calling `flush` directly does both _and_ also propagates any IO errors back to the caller.

May fix (or may not fix) issues like #450


bugadani:
> Lets see if it fix any issues!

It fixes mine :)

SergioGasquez:
> It fixes mine :)

That's already a big thing! Thanks for the PR!

PR #524, opened 2023-11-28
Don't silently fail if defmt data isn't available/correct

### LINKED ISSUE #523: error out in some form when no table could be found parsing an elf for defmt
current behavior just silently does nothing if no table found when user runs espflash like `espflash monitor --elf=test  --log-format=defmt
`

offending part https://github.com/esp-rs/espflash/blob/main/espflash/src/cli/monitor/parser/esp_defmt.rs#L88

bugadani:
Closes #523

### DIFF
--- a/espflash/src/cli/monitor/parser/esp_defmt.rs
+++ b/espflash/src/cli/monitor/parser/esp_defmt.rs
@@ -5,6 +5,7 @@ use crossterm::{
     QueueableCommand,
 };
 use defmt_decoder::{Frame, Table};
+use miette::{bail, miette};
 
 use crate::cli::monitor::parser::InputParser;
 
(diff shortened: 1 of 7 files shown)


espflash/src/cli/monitor/parser/esp_defmt.rs
```diff
@@ -79,31 +80,37 @@ impl FrameDelimiter {
...
-                Some(table)
-            } else {
-                log::warn!("Unsupported defmt encoding: {:?}", encoding);
-                None
-            }
-        })
+    /// Loads symbols from the ELF file (if provided) and initializes the context.
+    fn load_table(elf: Option<&[u8]>) -> miette::Result<Table> {
+        let Some(elf) = elf else {
+            bail!("The defmt log format can not be used without an .elf file");
```
SergioGasquez, line 90: I would create a custom error for this in the `Error` enum. Same applies for others `bail!`s

espflash/src/cli/monitor/parser/esp_defmt.rs
```diff
@@ -79,31 +80,37 @@ impl FrameDelimiter {
...
-            // a defmt stream and an ASCII log stream over the same serial port.
-            if encoding == defmt_decoder::Encoding::Rzcobs {
-                Some(table)
-            } else {
-                log::warn!("Unsupported defmt encoding: {:?}", encoding);
-                None
-            }
-        })
+    /// Loads symbols from the ELF file (if provided) and initializes the context.
+    fn load_table(elf: Option<&[u8]>) -> miette::Result<Table> {
```
SergioGasquez, line 88: I would recommend including `miette::Result` just to be consistent with the rest of the codebase where we use the include and only use `Result`  and not `miette::Result`

SergioGasquez (commented):
Thanks for the improvements!  Just left a few nitpicks in comments.

Also, mind updating the CHANGELOG?

SergioGasquez (approved):
LGTM! Thanks!

Vollbrecht:
Ty

PR #525, opened 2023-11-30
Add `min_chip_rev` argument

SergioGasquez:
With these changes, when using the argument and specifying a `--min-chip-rev ` greater that the chip revision you get:
```
     Running `/home/sergio/Documents/Espressif/forks/espflash/target/release/espflash flash --min-chip-rev 5 esp32c3`
[2023-11-30T15:26:40Z INFO ] Serial port: '/dev/ttyACM0'
[2023-11-30T15:26:40Z INFO ] Connecting...
[2023-11-30T15:26:40Z INFO ] Using flash stub
[2023-11-30T15:26:40Z WARN ] Setting baud rate higher than 115,200 can cause issues
Error: espflash::unsupported_chip_revision

  × Minimum supported revision is 5, connected device's revision is 3

```

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -133,6 +133,9 @@ pub struct FlashArgs {
     /// Logging format.
     #[arg(long, short = 'L', default_value = "serial", requires = "monitor")]
     pub log_format: LogFormat,
+    /// Minimum chip revision supported by image, in format: major * 100 + minor
+    #[arg(long, default_value = "0")]
+    pub min_chip_rev: u16,
     /// Open a serial monitor after flashing
     #[arg(short = 'M', long)]
     pub monitor: bool,
@@ -182,6 +185,9 @@ pub struct SaveImageArgs {
     pub chip: Chip,
     /// File name to save the generated image to
     pub file: PathBuf,
+    /// Minimum chip revision supported by image, in format: major * 100 + minor
+    #[arg(long, default_value = "0")]
+    pub min_chip_rev: u16,
     /// Boolean flag to merge binaries into single binary
     #[arg(long)]
     pub merge: bool,
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -134,7 +134,7 @@ pub struct FlashArgs {
     #[arg(long, short = 'L', default_value = "serial", requires = "monitor")]
     pub log_format: LogFormat,
     /// Minimum chip revision supported by image, in format: major * 100 + minor
-    #[arg(long, default_value = "0")]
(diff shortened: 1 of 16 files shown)


espflash/src/cli/mod.rs
```diff
@@ -133,6 +133,9 @@ pub struct FlashArgs {
     /// Logging format.
     #[arg(long, short = 'L', default_value = "serial", requires = "monitor")]
     pub log_format: LogFormat,
+    /// Minimum chip revision supported by image, in format: major * 100 + minor
+    #[arg(long, default_value = "0")]
+    pub min_chip_rev: u16,
```
bjoernQ, line 138: I wonder if it would be worth it to parse the `major.minor` format here (and I don't even know exactly how to do that with clap) - but might not be worth it

SergioGasquez: Good idea! Just pushed changes to implement this in [0570570](https://github.com/esp-rs/espflash/pull/525/commits/05705702b4272003d9f28ea84ddc3866bcfc076d)

espflash/src/cli/mod.rs
```diff
@@ -133,6 +133,9 @@ pub struct FlashArgs {
     /// Logging format.
     #[arg(long, short = 'L', default_value = "serial", requires = "monitor")]
     pub log_format: LogFormat,
+    /// Minimum chip revision supported by image, in format: major * 100 + minor
+    #[arg(long, default_value = "0", value_parser = parse_chip_rev)]
```
bjoernQ, line 137: great we have the parse_chip_rev now 👍 we need to change the default to conform to the format

espflash/src/cli/mod.rs
```diff
@@ -133,7 +133,7 @@ pub struct FlashArgs {
     /// Logging format.
     #[arg(long, short = 'L', default_value = "serial", requires = "monitor")]
     pub log_format: LogFormat,
-    /// Minimum chip revision supported by image, in format: major * 100 + minor
+    /// Minimum chip revision supported by image, in format: major.minor
     #[arg(long, default_value = "0", value_parser = parse_chip_rev)]
```
bjoernQ, line 137: ah sorry - what I meant was `default_value = "0.0"` since otherwise when not providing the argument we get

```
error: invalid value '0' for '--min-chip-rev <MIN_CHIP_REV>': Failed to parse chip revision: 0. Chip revision must be in the format `major.minor`
```

SergioGasquez: Oh! Didn't notice it, updated.

bjoernQ (approved):
LGTM - works fine

PR #528, opened 2023-12-12
Add serde derives to flash params

bugadani:
This may be useful for probe-rs ([example where I would have used it](https://github.com/probe-rs/probe-rs/pull/1952))


SergioGasquez (approved):
LGTM! Thanks

PR #530, opened 2023-12-14
Make `FlasherSize::from_detected` public

MabezDev:
We'd like to use this in probe-rs' flash size detection.


SergioGasquez (approved):
LGTM!

PR #533, opened 2023-12-20
Initial support for P4

SergioGasquez:
There are still a few TODOs to resolve but we match the state of P4 in `esptool`.

@georgik did some testing and was able to flash and monitor:
```
espflash flash --no-stub build/hello_world.elf --monitor
...
I (414) main_task: Calling app_main()
Hello world!
This is esp32p4 chip with 2 CPU core(s), , silicon revision v0.0, 4MB external flash
Minimum free heap size: 618808 bytes
```


jessebraham (approved):
LGTM!

Issue #500, opened 2023-11-07
espflash always pulls in libudev for UART operation, even if one only decodes the IDF format.


Yatekii:
I would love if this could be put behind a feature flag so other crates don't need the udev dependency :)

PR #535, opened 2023-12-27
Add serialport feature

SergioGasquez:
This tries to avoid always pulling `libudev`, see #500. Not sure if it is the best approach, though.


bjoernQ (approved):
LGTM

bjoernQ:
I think this is for probe-rs but I'm wondering what is probe-rs actually using from espflash? Just the image creation? Would it make sense (long term) to have it in its own crate?

SergioGasquez:
> what is probe-rs actually using from espflash? 
In the issue #500, they mention “even if one only decodes the IDF format”, so that's one thing, not sure if they are using anything else.

Issue #479, opened 2023-09-20
Improve flash speed by skipping bootloader and partition tables when possible


MabezDev:
See: https://github.com/esp-rs/esp-hal/issues/803#issuecomment-1727958673

But essentially, we could make espflash smart enough to flash the things that have changed, which is usually just the main app.

bugadani:
If it is possible to read the hash of a particular flash region, it should also be possible I believe to implement selective reflashing: for each flash block, read the hash, compare, and only reflash if differs.

Also, cc https://github.com/esp-rs/espflash/issues/259 as this issue is basically the meat of flash content verification.

bjoernQ:
There is  ` ESP_FLASH_VERIFY_MD5` command ( https://github.com/espressif/esptool/blob/8538b8983e5e04ad5d6f63620d609bfa63412eb3/flasher_stub/include/stub_flasher.h#L53 )

I think it's not yet supported by the Rust esp-flasherstub but by the esp-tool stub and in ROM (not for ESP-8266)

bjoernQ:
Started to look into `ESP_FLASH_VERIFY_MD5` ... as mentioned, we don't have it supported by the Rust esp-flasherstub which is not that good

Tried without a flasher stub and it needs quite a few changes to espflash internals because of the way we currently parse responses into `CommandResponse` (and how we currently cannot return something that is not u32)

Additionally, there are at least two different response formats for that command

Issue #259, opened 2022-10-17
Add an option to verify the flash contents after flashing


jessebraham:
See:  
https://github.com/espressif/esptool/blob/590c2c68381b32e1d296d0b338b9d50a231169a5/esptool/cmds.py#L1135-L1173

bugadani:
It would also be nice if we had the option to save a partition to a file. If verification requires reading back from flash, it probably shouldn't be very hard to add either function if the other exists :)

SergioGasquez:
Related esptool docs: https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/serial-protocol.html#verifying-uploaded-data

Issue #526, opened 2023-12-06
ESP32-C2 26MHz logs don't work properly


JurajSadel:
Log from `embassy_hello_world"

```rust
Chip type:         esp32c2 (revision v1.0)
Crystal frequency: 26MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       08:3a:8d:48:7d:6c
App/part. size:    89,760/4,128,768 bytes, 2.17%
[00:00:01] [========================================]      11/11      0x0
[00:00:00] [========================================]       1/1       0x8000
[00:00:02] [========================================]      22/22      0x10000
[2023-12-06T10:30:24Z INFO ] Flashing has completed!
Commands:
    CTRL+R    Reset chip
    CTRL+C    Exit

r�bl~�lr��
�nnr�
��
l��l
�nn�l�n�쎌
��l`�r�b



b�l����lllp��nn��
```

MabezDev:
I don't have a 26mhz board, but to be sure, you're passing the `26mhz_xtal` feature too?

SergioGasquez:
Id suggest building a bootloader in a esp-idf project and using it,see https://esp-rs.github.io/book/troubleshooting/espflash.html#rtc_clk_init-possibly-invalid-config_xtal_freq-setting.

I don't have a 26 MHz board but I can elaborate more on Monday if needed.

beeb:
Thanks @SergioGasquez , I had not found that resource in the book. Unfortunately, I have no `esp-idf-sys-*` folder in my `target/xtensa-esp32-none-elf/release/build` folder. Can't find any bootloader binary in fact...

beeb:
I'm trying to compile the project generated by `esp-idf-template` but it complains that `ERROR: C:\tb\.embuild\espressif\espidf.constraints.v5.1.txt doesn't exist. Perhaps you've forgotten to run the install scripts. Please check the installation guide for more information.`

But I don't see any install script. I used `espup` to install everything.

Vollbrecht:
we are currently revisiting how the bootloader is exposed in esp-idf-sys. This is already done on gh but not released. Currently you can find the bootloader in the build artifacts of esp-idf-sys. It might look like the following inside your target dir 
`target/riscv32imc-<or-your-target>/debug/build/esp-idf-sys-<random-number>/out/build/bootloader`

beeb:
Cool thanks, so I am now able to build and flash my project by specifying the bootloader that was built in the idf template project.

[...]

PR #536, opened 2023-12-27
Add `checksum-md5` command

bjoernQ:
This adds a `checksum-md5` command to espflash.

e.g.
```
❯ espflash checksum-md5 --address 0x10000 --length 0x1000 --no-stub
[2023-12-27T11:39:27Z INFO ] Serial port: 'COM22'
[2023-12-27T11:39:27Z INFO ] Connecting...
[2023-12-27T11:39:27Z WARN ] Setting baud rate higher than 115,200 can cause issues
0x95d0df6456de4ac114251a67a5f85
```

While this is not too exciting on its own, it's the ground work for #259 and #479

It's very unfortunate that the Rust flasher-stub currently doesn't implement the required command while the ROM code and the C based flasher stub does. (Verified by temporarily reverting the stubs locally). Maybe we should revert the change to use the Rust flasher-stub (but use it based on a feature at build time) until it's supported there

I checked the result with reading a flash region with `esptool` and running `md5sum` on it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add --chip argument for flash and write-bin commands (#514)
 - Add --partition-table-offset argument for specifying the partition table offset (#516)
 - Add `Serialize` and `Deserialize` to `FlashFrequency`, `FlashMode` and `FlashSize`. (#528)
+- Add `checksum-md5` command (#536)
 
 ### Fixed
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -126,12 +126,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
-name = "array-init"
-version = "2.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3d62b7694a562cdf5a74227903507c56ab2cc8bdd1f781ed5cb4cf9c9f810bfc"
-
-[[package]]
 name = "arrayvec"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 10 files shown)


SergioGasquez (changes requested):
Thanks for adding this feature! We should also add the subcommand for `cargo-espflash`

bjoernQ:
> > It's very unfortunate that the Rust flasher-stub currently doesn't implement the required command while the ROM code and the C based flasher stub does. (Verified by temporarily reverting the stubs locally). Maybe we should revert the change to use the Rust flasher-stub (but use it based on a feature at build time) until it's supported there:
> 
> The initial idea of the Rust stub, was to put them behind a feature, but we decided not to do that because nobody will use the feature. That being said, we now have an argument to do so or to go back to the C stubs or just force to use `--no-stub` for this subcommand until [esp-rs/esp-flasher-stub#43](https://github.com/esp-rs/esp-flasher-stub/issues/43) is implemented

BTW I also currently cannot flash the H2 with the current stub - we really need to be very confident that the stub works for us (including MD5) before we force every user to use it (while I agree that no average user will use such a feature we could agree to use it in our team - it's kind of a chicken and egg problem)

SergioGasquez:
Current `main` of `esp-flasher-stub` should be able to flash all targets, we should probably cut a release and update them in `espflash`. But I agree, current state of `espflash` is not good as you can't flash many targets, I'm happy to add the Rust stubs behind a feature.

SergioGasquez:
In any case, that behind the scope of this PR, I'll do some testing tomorrow morning and update with the results!

bjoernQ:
I locally updated the flasher stubs ( commit 2b0b0f0de5020e6d5e2bf39e8950d49737fd5edb ) and it works for all chips - also flashing H2 works now for me

I'm still not sure if we should add a feature to choose the old vs the new stub. e.g. the new stub doesn't support 26 MHz xtals - maybe having at least a feature to force using the old stubs might be helpful for those users (but certainly not scope of this PR)

SergioGasquez (approved):
LGTM! Tested it on all the targets but S2 and ESP32 and worked fine with the new stubs, will create a PR updating them!

Edit: PR updating the stubs: https://github.com/esp-rs/espflash/pull/537

SergioGasquez:
> I'm still not sure if we should add a feature to choose the old vs the new stub. e.g. the new stub doesn't support 26 MHz xtals - maybe having at least a feature to force using the old stubs might be helpful for those users (but certainly not scope of this PR)

Is the 26 MHz a stub issue? I don't have any 26 MHz board so I cant test it. We have https://github.com/esp-rs/espflash/issues/526, but that's a bootloader issue. This definitely needs some investigation, so I'll try to get some 26 MHz devkit

bjoernQ:
> > I'm still not sure if we should add a feature to choose the old vs the new stub. e.g. the new stub doesn't support 26 MHz xtals - maybe having at least a feature to force using the old stubs might be helpful for those users (but certainly not scope of this PR)
> 
> Is the 26 MHz a stub issue? I don't have any 26 MHz board so I cant test it. We have #526, but that's a bootloader issue. This definitely needs some investigation, so I'll try to get some 26 MHz devkit

Yes it works with `--no-stub` ... so probably we don't need that feature ... I forgot there is `--no-stub` 😆

PR #538, opened 2023-12-28
Implement skip and verify

### LINKED ISSUE #259: Add an option to verify the flash contents after flashing

### LINKED ISSUE #479: Improve flash speed by skipping bootloader and partition tables when possible

bjoernQ:
Implements verify, closes #259 
Implements skip flashing of unchanged flash regions, closes #479


SergioGasquez (approved):
LGTM! Tested it a bit and re-flashing the same app is now way faster, thanks for the feature!

PR #542, opened 2024-01-09
Include license texts in packaged crates

davide125:
Ensure the license files are included in the crates when they're published.


MabezDev (approved):
Thanks!

Issue #469, opened 2023-08-27
Attempt to read espflash config from project_dir/.config/espflash.toml


svenstaro:
The config is currently only read system-wide (`$HOME/.config/espflash/espflash.toml`) but this is annoying if you work on multiple unrelated projects. I think cargo-espflash should first attempt to read `project_dir/.config/espflash.toml` and fall back to the system config if the former config doesn't exist.

galah92:
Would also like a way to have a local config file (that take precedence over the global one). That would help multiple developers working on the same project were each has it's own serial mapping.
If there's another way to achieve that, I'll be happy to hear about it.
Also, our team mostly use the Cargo Runner (`.cargo/config.toml`), to run `espflash`, so ideally it would be a solution that supports this.

PR #551, opened 2024-01-17
Unify configuration methods

SergioGasquez:
- Check for config file in:
  - <current_folder>/.config/espflash.toml
  - [<config_dir>](https://docs.rs/directories/5.0.1/directories/struct.BaseDirs.html#method.config_dir)/espflash/espflash.toml
- Add the following configurations:
  - Bootloader
  - Partition table
  - Baudrate
  - Port
- [cargo-espflash]: Remove using partition table and bootloader path from cargo metadata, use the config file instead.


MabezDev:
> <current_folder>/.config/espflash.toml

What's the rationale around having it inside `<current_dir>/.config`, instead of `<current_dir>`? It doesn't matter much to me, but I just wonder if there is a particular reason. Most other tools with configs, namely probe-rs (cargo embed) place there configuration files in the root of the project.

cargo-espflash/src/package_metadata.rs
```diff
@@ -84,17 +69,10 @@ impl PackageMetadata {
                 Some(meta) if meta.is_table() => {
                     let meta = meta.as_table().unwrap();
 
-                    espflash_meta.bootloader = meta
-                        .get("bootloader")
-                        .map(|bl| package.root().join(bl.as_str().unwrap()));
-
+                    // TO BE REMOVED?
```
MabezDev, line 72: Could you remind me again what this is? I don't use cargo-espflash very often 😅.

SergioGasquez:
> What's the rationale around having it inside `<current_dir>/.config`, instead of `<current_dir>`? It doesn't matter much to me, but I just wonder if there is a particular reason. Most other tools with configs, namely probe-rs (cargo embed) place there configuration files in the root of the project.

Tbh, I just followed the suggestion in https://github.com/esp-rs/espflash/issues/469, I can update the path to be `<current_dir>`

cargo-espflash/src/package_metadata.rs
SergioGasquez: You could set the `partition-table`, `bootloader` and `format` in the package metadata, see https://github.com/esp-rs/espflash/tree/main/cargo-espflash#package-metadata. At the moment, I removed `baudrate` and `partition-table` properties. IMHO, we can also remove the format

MabezDev: Ah, I see, thanks! Yes lets remove format too.

SergioGasquez:
> imo it should probably be in the current directory, but I'll let @jessebraham @bjoernQ and @JurajSadel weigh in too.

Awesome, let's hear their opinion, I am happy to change it with whatever we decide

bjoernQ:
I think I would prefer the current directory

jessebraham:
Sorry if I've missed something, why was the Cargo package metadata loading in `cargo-espflash` crippled? What's the point of even having it in its current rendition?

SergioGasquez:
> Sorry if I've missed something, why was the Cargo package metadata loading in `cargo-espflash` crippled? What's the point of even having it in its current rendition?

I'd say we can remove the Cargo metadata, as it is now in the config file and its the same for `espflash/cargo-espflash`

jessebraham:
That's disappointing, I use it in most of my projects :/ Guess I will switch over, thanks.

SergioGasquez:
> That's disappointing, I use it in most of my projects :/ Guess I will switch over, thanks.

I mean, we can keep it. No hard opinion there, but then we would have multiple ways of defining those configs.

SergioGasquez:
I updated the local config to live under `<current_dir>/espflash.toml>, removed the `format` param from cargo metadata for `cargo-espflash` and udpated the docs

bjoernQ (approved):
Looks good to me

JurajSadel (approved):
LGTM

PR #553, opened 2024-01-18
Support different crystal frequencies bootloader

SergioGasquez:
- Add bootloarder for ESP32 26MHz and ESP32-C2 26MHz
- When using `flash` use the target xtal_freq to choose the bootloader
- When using `save-image` add a `-x/--xtal-freq` arg to choose the bootloader.
- Add a `XtalFrequency` enum to avoid hardcoded values.


MabezDev (approved):
This looks good to me! Annoying that we have to lug around some more bootloaders, but given that we've only seen 26mhz xtals on the esp32 and esp32c2, its not too bad.

Thanks for working on this!

MabezDev:
> As usual, I forgot to update the changelog, so I just did it and also formatted other changelog entries

Maybe you can steal our CI config on esp-hal and esp-wifi to make sure we add entries. That can wait for another PR though, if you wish!

MabezDev (approved):

SergioGasquez:
> > As usual, I forgot to update the changelog, so I just did it and also formatted other changelog entries
> 
> Maybe you can steal our CI config on esp-hal and esp-wifi to make sure we add entries. That can wait for another PR though, if you wish!

This was discussed when we added the changelog, https://github.com/esp-rs/espflash/pull/416#discussion_r1210376372, but we decided not to add it as it also has drawbacks

MabezDev:
I think that was _before_ we fixed the fork issue around `skip-changelog`, @jessebraham will be able to confirm. Either way just a suggestion, feel free to ignore! Let's get this merged :).

PR #554, opened 2024-01-19
Windows USB serial workaround: set DTR after RTS

haileys:
When setting DTR before RTS, espflash can't connect to a USB serial device on Windows:

image

After changing the order, it can:

image

Strange but true!

esptool itself [features a similar workaround](https://github.com/espressif/esptool/blob/16e4faeeaa3f95c6b24dfdcc498ffc33924d5f5f/esptool/reset.py#L59-L62) - it additionally sets DTR to its last known state whenever RTS is set. For our purposes though, it seems good enough to just switch the order we set the pins.

### DIFF
--- a/espflash/src/connection/reset.rs
+++ b/espflash/src/connection/reset.rs
@@ -92,24 +92,24 @@ impl ResetStrategy for ClassicReset {
             "Using Classic reset strategy with delay of {}ms",
             self.delay
         );
-        self.set_dtr(interface, false)?;
         self.set_rts(interface, false)?;
+        self.set_dtr(interface, false)?;
 
-        self.set_dtr(interface, true)?;
         self.set_rts(interface, true)?;
+        self.set_dtr(interface, true)?;
 
-        self.set_dtr(interface, false)?; // IO0 = HIGH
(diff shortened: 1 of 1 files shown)


SergioGasquez (commented):
Thanks for the fix! I wonder if we should also update the [`UsbJtagSerialReset` order](https://github.com/esp-rs/espflash/blob/main/espflash/src/connection/reset.rs#L169). 

> esptool itself [features a similar workaround](https://github.com/espressif/esptool/blob/16e4faeeaa3f95c6b24dfdcc498ffc33924d5f5f/esptool/reset.py#L59-L62) - it additionally sets DTR to its last known state whenever RTS is set.

I wonder if we could easily store the value of `dtr` and `rts` and do like `esptool` does, when setting `rst` set the current value of `dtr`, so we can keep the same order as `esptool` while have this Windows workaround

SergioGasquez:
I've also added the changes for UsbJtagSerialReset in https://github.com/SergioGasquez/espflash/tree/fix/windows-reset. Do you have by any chance a devkit with jtab port available by any chance to confirm if the issue is also present there and my changes fix it?

Issue #346, opened 2023-01-17
espflash monitor without "connecting"?


MabezDev:
Example: I have a debugger session running but I want to view UART output. Right now running the monitor subcommand will reset the chip to detect what it is. It will also load the stub, which seems unnecessary for the monitor subcommand.

Perhaps it should be possible to optionally pass the chip via the CLI, if it's passed then espflash won't try and autodetect the chip and instead just go straight to monitoring.

jessebraham:
Looking at the monitor implementation I'm not sure we even need to pass the chip? Seems to be agnostic in that regard, unless I'm missing something.

I agree the stub isn't really necessary for the monitor so that could easily be disabled I'm sure. This shouldn't require too many changes I don't think, I'll look into it soon.

PR #561, opened 2024-01-25
Add before and after reset arguments

SergioGasquez:
- Adds the `-b/--before` argument
- Adds the `-a/--after` argument

This also solves #346 as `espflash monitor --before no-reset-no-sync --chip esp32c3` can monitor a target without resetting it.

### DIFF
--- a/espflash/src/connection/mod.rs
+++ b/espflash/src/connection/mod.rs
@@ -202,20 +202,18 @@ impl Connection {
             ResetAfterOperation::SoftReset => {
                 soft_reset(self, false, is_stub, chip)?;
                 println!("Soft resetting");
-                return Ok(());
+                Ok(())
             }
             ResetAfterOperation::NoReset => {
                 soft_reset(self, true, is_stub, chip)?;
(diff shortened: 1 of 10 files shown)


espflash/src/connection/mod.rs
```diff
@@ -163,6 +232,28 @@ impl Connection {
         Ok(())
     }
 
+    // Reset the device taking into account the reset after argument
+    pub fn reset_after(&mut self, is_stub: bool, chip: Chip) -> Result<(), Error> {
+        match self.after_operation {
+            ResetAfterOperation::HardReset => HardReset.reset(&mut self.serial),
+            ResetAfterOperation::SoftReset => {
+                println!("Soft resetting");
```
MabezDev, line 240: These should probably be log statements, right?

MabezDev (approved):
Just a small nit about the println statements, other than that this LGTM! Thanks!

espflash/src/connection/mod.rs
SergioGasquez: Updated to info messages

PR #562, opened 2024-01-25
Windows: Set RTS before DTR to avoid issues

SergioGasquez:
Supersedes #554 

I couldn't verify that this actually works as I have not encountered this issue, see #554 and [`esptool` workaround ](https://github.com/espressif/esptool/blob/16e4faeeaa3f95c6b24dfdcc498ffc33924d5f5f/esptool/reset.py#L59-L62)


MabezDev:
@bjoernQ would you mind testing this just to ensure it still works on windows after this change?

bjoernQ (approved):
Works exactly the same as before for me. Tried with a few different boards (using different Windows drivers)

MabezDev:
Thank you for testing @bjoernQ! LGTM!

MabezDev (approved):

PR #566, opened 2024-01-30
Add FlashDataBuilder

bugadani:
This PR implements my suggestions from https://github.com/esp-rs/espflash/pull/512#issuecomment-1916650260 except for the error handling which is a more fundamental question it looks like.


SergioGasquez (approved):

PR #567, opened 2024-01-31
Re-calculate the bootloader's SHA256

### LINKED ISSUE #206: SHA-256 comparison failed: [...] attempting to boot anyway...
When the app image header is updated with new flash configuration values, output like the following can be seen in the bootloader messages:

```text
SHA-256 comparison failed:
Calculated: bf6e7aa7738b60d6b15de2bea35979ca40a8a94224f3c78ef00f19a96becf599
Expected: c279089098b4d648c53538bb1dfba7caa16140817d4be0ee4394fe6967e6813d
Attempting to boot anyway...
```

While the application still executes as expected, we will need to address this. This is not an issue unless you plan to use Secure Boot, which we do not presently support anyways. I have a local branch in-progress but it still needs some work.

bjoernQ:
Fixes #206 

We modify the ESP-IDF booloader's header - then we also need to re-calculate it's SHA256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed `defmt` parsing when data is read in parts (#503)
 - Use partition table instead of hard-coded values for the location of partitions (#516)
 - Fixed a missed `flush` call that may be causing communication errors (#521)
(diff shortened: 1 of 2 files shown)


SergioGasquez (approved):
Sweet! Thanks for the fix, I didn't realize it was the bootloader hash which was giving troubles.

Issue #570, opened 2024-02-02
Flashing a chip with different xtal freq resutls in a panic


SergioGasquez:
In 2.1, you could flash applications for wrong targets, this resulted in a bootloop. Now, trying to do so results in:
```
[2024-02-02T10:07:06Z INFO ] Serial port: '/dev/ttyUSB0'
[2024-02-02T10:07:06Z INFO ] Connecting...
[2024-02-02T10:07:06Z DEBUG] Using UnixTight reset strategy with delay of 50ms
Error:   × Main thread panicked.
  ├─▶ at espflash/src/connection/mod.rs:171:80
  ╰─▶ called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 0, error_len: Some(1) }
  help: set the `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

This happens when we try to decode in which bootmode the target is, [we do so by reading the serial port](https://github.com/esp-rs/espflash/blob/main/espflash/src/connection/mod.rs#L156-L186), and if the xtal freq is different we see noise and we cant decode it, resulting in this error.

PR #572, opened 2024-02-03
Add error for cases when we cant read the serial monitor trying to decode the bootmode

SergioGasquez:
In some (weird cases), like flashing a 26MHz C2 instead of 40Mhz target, https://github.com/esp-rs/espflash/issues/570 would happen. It should be an error, but the error should mean something, so I just added a custom error for the case.

### DIFF
--- a/espflash/src/error.rs
+++ b/espflash/src/error.rs
@@ -99,7 +99,7 @@ pub enum Error {
     #[error("Invalid byte sequence read from the serial port while trying to detect Boot Mode")]
     #[diagnostic(
         code(espflash::invalid_serial_read),
-        help("This migth be caused by a xtal frequency missmatch")
+        help("This might be caused by a xtal frequency mismatch")
     )]
     InvalidSerialRead,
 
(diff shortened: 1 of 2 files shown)


espflash/src/error.rs
```diff
@@ -96,6 +96,10 @@ pub enum Error {
     #[error("The provided bootloader binary is invalid")]
     InvalidBootloader,
 
+    #[error("Invalid byte sequence read from the serial port while trying to detect Boot Mode")]
```
MabezDev, line 99: Do you think we should mention the possible failure case? I.e "this be caused by a xtal frequency mismatch" or something?

SergioGasquez: Good idea, just added it on the help message

espflash/src/error.rs
```diff
@@ -97,7 +97,10 @@ pub enum Error {
     InvalidBootloader,
 
     #[error("Invalid byte sequence read from the serial port while trying to detect Boot Mode")]
-    #[diagnostic(code(espflash::invalid_serial_read))]
+    #[diagnostic(
+        code(espflash::invalid_serial_read),
+        help("This migth be caused by a xtal frequency missmatch")
```
MabezDev, line 102:
```suggestion
        help("This might be caused by a xtal frequency mismatch")
```

MabezDev (approved):
LGTM, thanks!

PR #573, opened 2024-02-03
Tolerate some non-utf8 data when detecting boot mode over serial

kcking:
My ESP32-WROOM is sending some non-utf8 data just after reset, causing boot mode detection to fail.

This commit uses `String::from_utf8_lossy` to ignore these bytes.

Tested on macOS, boards: ESP32-WROOM-32, ESP32-WROOM-32D, ESP32-C3-MINI-1, ESP32-C3-12F.

### DIFF
--- a/espflash/src/connection/mod.rs
+++ b/espflash/src/connection/mod.rs
@@ -174,6 +174,9 @@ impl Connection {
             }
 
             let read_slice = String::from_utf8_lossy(&buff[..read_bytes as usize]).into_owned();
+            if !read_slice.contains("boot") {
+                return Err(Error::InvalidSerialRead);
+            }
 
             let pattern = Regex::new(r"boot:(0x[0-9a-fA-F]+)(.*waiting for download)?").unwrap();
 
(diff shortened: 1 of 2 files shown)


espflash/src/connection/mod.rs
```diff
@@ -174,18 +173,19 @@ impl Connection {
                 )));
             }
 
-            let read_slice = from_utf8(&buff[..read_bytes as usize]).map_err(|err| {
-                debug!("Error: {}", err);
-                Error::InvalidSerialRead
-            })?;
+            let read_slice = String::from_utf8_lossy(&buff[..read_bytes as usize]).into_owned();
```
SergioGasquez, line 176: I was thinking of checking if the `read_slice` only contained the [REPLACEMENT_CHARACTER](https://doc.rust-lang.org/std/char/constant.REPLACEMENT_CHARACTER.html), but maybe it would be better to see if the `read_slice` contains `boot`:

```suggestion
            let read_slice = String::from_utf8_lossy(&buff[..read_bytes as usize]).into_owned();
             //Check if the read_slice contains other characters than REPLACEMENT_CHARACTER
            if !read_slice.contains("boot") {
                return Err(Error::InvalidSerialRead);
            }
```

SergioGasquez (approved):

PR #577, opened 2024-02-05
Remove direct boot

jessebraham:
Removes the direct boot image format. Also refactored away the `ImageFormat` trait since we only have one format now.


espflash/tests/README.md
SergioGasquez, line 1: Is this any useful now? I think it can be removed with the `test/resources/` bins and elfs, I think with the changes they are not being used.

SergioGasquez (commented):
Changes LGTM! Just left a note regarding the test resources!

I've also done some testing, and I was able to flash all the targets but encountered #573 a few times, so it might be worth fixing it before the rc.

MabezDev (approved):
Whilst I can see a world where having it here is useful, I'm okay with removing it as we don't have any current use. We can always add it back in the future.

LGTM!

PR #584, opened 2024-02-15
Fix flash/monitoring of 26mhz targets

SergioGasquez:
- Remove `InvalidSerialRead` error
- Fix Clippy lints
- Update monitoring baudrate when using stub
- Update ESP32-C2 stub to support 26MHz targets. Stubs have been generated using  `esp-flasher-stub`, which uses `esp-hal` with the changes of this branch: https://github.com/esp-rs/esp-hal/compare/main...SergioGasquez:esp-hal:feat/clk-src

Thanks @bjoernQ for the help resolving this issue

I've reverted the changes for ESP32, as I dont have a 26MHz to test it, I dont know if anyone from the team has one, and ESP32 cant choose `xtal_clock` as UART source. From ESP32 TRM:
> The UART block can derive its clock from two
sources: the 80-MHz APB_CLK, or the reference clock REF_TICK (please refer to Chapter Reset and Clock for
more details). These two clock sources can be selected by configuring UART_TICK_REF_ALWAYS_ON.

> UART_TICK_REF_ALWAYS_ON This register is used to select the clock; 1: APB clock; 0: REF_TICK

## Tests
- I've build and flash the `rng` example of esp-hal to a 26MHz C2 and a 40MHz C2 with no issues
  - Monitoring also worked fine
- Board info shows the proper freq for both 26 and 40 MHz


bjoernQ (approved):
LGTM - works fine for me with both 26 and 40 MHz ESP32-C2

PR #585, opened 2024-02-16
Remove support for Raspberry Pi's internal UART peripherals

SergioGasquez:
- Remove `raspberry` feature
- Remove `interface` module
- Use `Port` types instead of `Interface`


MabezDev (approved):
LGTM, we can always add it back if it really is used a lot.

PR #589, opened 2024-02-19
Address minor feedback things of rc.1

SergioGasquez:
- Prepare CHANGELOG for next dev-cycle
- Update log message when we skip writing a segment. Here is how it looks, reflashing an app where only the app changed:
  ```
  ...
  App/part. size:    223,776/4,128,768 bytes, 5.42%
  [2024-02-19T09:16:12Z INFO ] Segment at address '0x0' did not change, skipping it
  [2024-02-19T09:16:12Z INFO ] Segment at address '0x8000' did not change, skipping it
  [00:00:01] [========================================]      74/74      0x10000                                                                                                                                                                                    [2024-02-19T09:16:14Z INFO ] Flashing has completed!
  ```
- Include the removal of Cargo package metadata configuration in changelog.

cc: @jessebraham @Vollbrecht

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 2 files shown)


MabezDev (approved):
I think this makes it a bit more clear to users what's going on now. Thanks!

Vollbrecht:
Thanks, LGTM

PR #590, opened 2024-02-20
Add list-all-ports flag and filter the default list

### LINKED ISSUE #575: Serial port selection provides too many options, some of which are invalid
Previously we filtered the available serial ports to only those which looked like ESP devices. This was changed to display all devices, however, due to reports from some users that not all ports were listed on some Linux distributions. However, as a result of this change there are too many ports listed on other OSs (like macOS), many of which are not flashable in any way.

I would like to see us go back to filtering serial ports, and maybe we can have an option to list everything using a flag and/or environment variable instead.

SergioGasquez:
- Add `--list-all-ports` connection argument to avoid serial port filtering
- In all the targets, except for linux-musl, only list the available USB Ports by default (#590)

~~Draft PR until it can be properly tested and we verify that it closes #575~~
@jessebraham verified that it resolves #575, thanks for testing it!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Add `--list-all-ports` connection argument to avoid serial port filtering (#590)
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


MabezDev (approved):
Thanks! LGTM!

Issue #587, opened 2024-02-16
Library should not rely on `miette`


bugadani:
Relying on miette makes using espflash (the library) in application like probe-rs less friendly. We can no longer determine what exactly happened, we can only stringify the report and pass that to the user.

I absolutely love miette, but I think it has place only in applications, not libraries.

PR #591, opened 2024-02-20
Return `crate::Error` instead of a miette Report

bugadani:
cc #587


SergioGasquez (approved):
LGTM! Thanks for your help on testing and fixing this!

PR #157, opened 2022-04-12
bugfix: fix reset sequence for usb-serial-jtag chips

JurajSadel:
Small refactor of @bjoernQ's workaround in #153. 

Closes #153

Issue #153, opened 2022-04-07
Windows/JTAG-SERIAL Issues


bjoernQ:
Flashing on Windows (11) via JTAG-SERIAL has some issues
- the chip is not reset after flashing
- monitoring doesn't start after flashing if invoked with the `--monitor` option (most probably because it "hangs" in an attempt to reset the chip?)

Manually pushing the reset button and using "espmonitor" works however.

Windows Version: Microsoft Windows [Version 10.0.22000.556]

Used the ESP-Rust board, also a vanilla ESP32C3 shows the same behavior

MabezDev:
@JurajSadel I believe you ran into this at one point? Any ideas?

bjoernQ:
I can make it locally work for me with replacing `reset` in `connection.rs` with this
```rust
pub fn reset(&mut self) -> Result<(), Error> {
        sleep(Duration::from_millis(100));

        if self.port_info.pid == USB_SERIAL_JTAG_PID {
            self.serial.write_data_terminal_ready(false)?;
            self.serial.write_request_to_send(false)?;

            sleep(Duration::from_millis(100));

            self.serial.write_data_terminal_ready(false)?;
            self.serial.write_request_to_send(false)?;

            sleep(Duration::from_millis(100));

            self.serial.write_request_to_send(true)?;
            self.serial.write_data_terminal_ready(false)?;
            self.serial.write_request_to_send(true)?;

            sleep(Duration::from_millis(100));

            self.serial.write_data_terminal_ready(false)?;
            self.serial.write_request_to_send(false)?; 
    
        } else {
            self.serial.write_data_terminal_ready(false)?;
            self.serial.write_request_to_send(true)?;
    
            sleep(Duration::from_millis(100));
    
            self.serial.write_request_to_send(false)?;    
        }

        Ok(())
    }
``` 

Probably even less code would do ... just copied and adapted the code from `reset_to_flash` - CTRL+R still doesn't work then since there we have similar code duplicated (so changing it there should do the trick) - however I'm wondering why it's apparently just me and why it's only on Windows

JurajSadel:
@MabezDev For some reason, the `SW reset` or `write_requet_to_send()` is not working correctly on Windows, when the chip (with USB-serial-JTAG) is in download mode. Unfortunately, I don't know the root cause of this, yet - if it's in `serialport` or in a deeper layer like the driver, for example, and I'm not sure, how to debug it.  I'm sorry I didn't bring anything new to this issue.

[...]

PR #594, opened 2024-02-21
Fix USB JTAG Serial HardReset sequence in Windows

SergioGasquez:
- Reuse `reset_after_flash` sequence in `hard_reset`
  - We dont know where this sequence came from, but it seems to work fine in all platforms. See #157 and #153
  - Using [`esptool` HardReset sequence](https://github.com/espressif/esptool/blob/3301d0ff4638d4db1760a22540dbd9d07c55ec37/esptool/reset.py#L132-L153) does not work in Windows.
- Move `reset_after_flash` method to `reset` module  

Thanks @bjoernQ for the help debugging and testing this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


bjoernQ (approved):
LGTM! This fixes my issues. Thanks!

PR #595, opened 2024-02-21
Allow Cargo workspace config file

SergioGasquez:
- Local `espflash.toml` config file can also be stored in parent folder to simplify Cargo workspaces


MabezDev (approved):
LGTM, thanks!

PR #605, opened 2024-03-01
Update dependencies section: libudev

kubo39:


cargo-espflash/README.md
```diff
@@ -27,26 +27,24 @@ Supports the **ESP32**, **ESP32-C2/C3/C6**, **ESP32-H2**, **ESP32-P4**, and **ES
 
 If you are installing `cargo-espflash` from source (ie. using `cargo install`) then you must have `rustc>=1.74.0` installed on your system.
 
-If you are running **macOS** or **Linux** then [libuv] must also be installed; this is available via most popular package managers. If you are running **Windows** you can ignore this step.
+If you are running **Linux** then [libudev] must also be installed; this is available via most popular package managers. If you are running **Windows** you can ignore this step.
```
SergioGasquez, line 30: Does macOS already includes `libudev`? If thats the case we should also update the last sentence to include macOs

cargo-espflash/README.md
```diff
@@ -27,26 +27,24 @@ Supports the **ESP32**, **ESP32-C2/C3/C6**, **ESP32-H2**, **ESP32-P4**, and **ES
...
 
-```bash
-# macOS
-brew install libuv
+```console
```
SergioGasquez, line 32: Is there any reason to change this to `console`?

SergioGasquez (commented):
Thanks for your contribution! Forgot to update the `libudev` dep here, good catch! Left 2 comments

cargo-espflash/README.md
kubo39: macOS doesn't require lubudev, please see [serialport-rs dependencies](https://github.com/serialport/serialport-rs?tab=readme-ov-file#dependencies).
macOS uses CoreFoundation and IOKit, which is built-in.

cargo-espflash/README.md
kubo39: OK, back to using bash here.

cargo-espflash/README.md
SergioGasquez:
```suggestion
If you are running **Linux** then [libudev] must also be installed; this is available via most popular package managers. If you are running **Windows** or **macOS** you can ignore this step.
```

SergioGasquez (approved):
LGTM! Thanks again!

PR #610, opened 2024-03-12
Update stubs

SergioGasquez:
Stubs were generated using https://github.com/esp-rs/esp-flasher-stub/compare/main...SergioGasquez:esp-flasher-stub:feat/xtal-clock which patch my [`esp-hal` fork](https://github.com/esp-rs/esp-hal/compare/main...SergioGasquez:esp-hal:feat/xtal-clock), that uses `xtal-clock` as source for `uart`.


MabezDev (approved):
Thanks!

PR #623, opened 2024-04-07
Fixed docs for read-flash size parameter

irwineffect:
The size parameter of the read-flash subcommand indicated it was erasing flash.


SergioGasquez (approved):
LGTM! Good catch!

PR #625, opened 2024-04-18
HIL: Allow manually running tests on other repos/branches

SergioGasquez:
- Update log message to debug
-  Allow manually running HIL tests on other repos/branches

### DIFF
--- a/espflash/src/targets/esp32c2.rs
+++ b/espflash/src/targets/esp32c2.rs
@@ -1,5 +1,7 @@
 use std::{collections::HashMap, ops::Range};
 
+use log::debug;
+
 #[cfg(feature = "serialport")]
 use crate::{connection::Connection, targets::bytes_to_mac_addr};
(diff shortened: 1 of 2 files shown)


espflash/src/targets/esp32c2.rs
```diff
@@ -95,11 +97,11 @@ impl Target for Esp32c2 {
     ) -> Result<IdfBootloaderFormat<'a>, Error> {
         let booloader: &'static [u8] = match xtal_freq {
             XtalFrequency::_40Mhz => {
-                println!("Using 40MHz bootloader");
+                debug!("Using 40MHz bootloader");
```
MabezDev, line 100: Nice, I've been meaning to make an issue on that :D

SergioGasquez: It was never meant to be released this way, but it slipped in

PR #627, opened 2024-04-23
Support loading flash settings from the config file

simpkins:
This supports loading the flash size, frequency, and mode from the config file.

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -10,7 +10,13 @@
 //! [cargo-espflash]: https://crates.io/crates/cargo-espflash
 //! [espflash]: https://crates.io/crates/espflash
 
-use std::{collections::HashMap, fs, io::Write, num::ParseIntError, path::PathBuf};
+use std::{
+    collections::HashMap,
+    fs,
+    io::Write,
+    num::ParseIntError,
+    path::{Path, PathBuf},
+};
 
 use clap::Args;
 use clap_complete::Shell;
@@ -31,8 +37,8 @@ use crate::{
     elf::ElfFirmwareImage,
     error::{Error, MissingPartition, MissingPartitionTable},
     flasher::{
-        parse_partition_table, FlashData, FlashFrequency, FlashMode, FlashSize, Flasher,
-        ProgressCallbacks,
+        parse_partition_table, FlashData, FlashFrequency, FlashMode, FlashSettings, FlashSize,
+        Flasher, ProgressCallbacks,
     },
     targets::{Chip, XtalFrequency},
 };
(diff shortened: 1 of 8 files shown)


espflash/src/cli/mod.rs
```diff
@@ -214,35 +207,43 @@ pub struct ReadFlashArgs {
...
     /// File name to save the generated image to
     pub file: PathBuf,
-    /// Minimum chip revision supported by image, in format: major.minor
-    #[arg(long, default_value = "0.0", value_parser = parse_chip_rev)]
-    pub min_chip_rev: u16,
     /// Boolean flag to merge binaries into single binary
     #[arg(long)]
     pub merge: bool,
-    /// Custom partition table for merging
-    #[arg(long, short = 'T', requires = "merge", value_name = "FILE")]
```
simpkins, line 232: This change does remove the `requires = "merge"` setting from the `partition_table` argument for `save-image`, since the `ImageArgs` can't see the `merge` flag.  This attribute was previously different between the `save-image` and `flash` commands.

We could add back our own manual check to reject the `--partition-table` flag if `--merge` is not specified.  However, it does look like the `partition_table` argument is in fact used to print the partition size even when the `--merge` flag is not specified.

espflash/src/flasher/mod.rs
```diff
@@ -60,27 +60,38 @@ pub(crate) mod stubs;
 #[repr(u8)]
 pub enum FlashFrequency {
     /// 12 MHz
+    #[serde(rename = "12MHz")]
```
simpkins, line 63: I mentioned this in the commit comments, but just to explicitly call it out: these `rename` attributes will break anything that tries to deserialize FlashFrequency/FlashMode/FlashSize values that were serialized by old versions of the code.

However, I didn't see anything that appeared to be using this serialization.  The serialization support was added in #528, for https://github.com/probe-rs/probe-rs/pull/1952, but it seems like that functionality wasn't ultimately used in probe-rs as best as I can tell.

SergioGasquez: @bugadani Will this cause issues in probe-rs?

SergioGasquez (commented):
Thanks for adding this feature! Changes LGTM! Could you to please document it on both readmes under the config section (eg: https://github.com/esp-rs/espflash/tree/main/espflash#configuration-file)???

Also, have you conducted any testing?

espflash/src/flasher/mod.rs
bugadani: I don't think so. I think what happened here is that #528 was merged, then I relaised I can simplify something and didn't end up needing serde.

simpkins:
I've updated espflash/README.md with some brief documentation of the `[flash]` section.

> Also, have you conducted any testing?

I've been using it for my own development, and I have confirmed that the image header that is written contains the correct flash mode, size, and speed settings taken from the config file.  I've also confirmed that passing in flash mode CLI arguments takes precedence over values stored in the config file.

SergioGasquez:
> I've updated espflash/README.md with some brief documentation of the [flash] section.

I think the changes are not pushed, also, mind also updating the cargo-espflash readme? 

Thanks a lot! This is a very nice addition.

simpkins:
> I think the changes are not pushed, also, mind also updating the cargo-espflash readme?

Whoops you're right, my local repo had the wrong tracking branch so the push didn't update this branch.  I've updated the cargo-espflash readme too, and pushed to the right branch this time.

SergioGasquez (approved):

PR #632, opened 2024-04-29
Always resolve package_id from metadata when finding bootloader and partition table

kyrias:
The package ID format was officially documented as an opaque identifier and the only thing we were allowed to do with them is look them up in the cargo metadata.  Rust 1.77 switched the package ID field to use the publicly documented "Package ID Specification" format instead, which means that the old logic no longer works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed help text for size parameter of read-flash subcommand
+- [cargo-espflash]: Always resolve package_id from metadata when finding bootloader and partition table (#632)
 
 ### Changed
 
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -4,7 +4,7 @@ use std::{
     process::{exit, Command, ExitStatus, Stdio},
 };
 
-use cargo_metadata::Message;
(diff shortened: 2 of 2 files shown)


kyrias:
Turns out that it's only things built by `-Zbuild-std` which end up not existing in the metadata, so I'm going to rework this to always look it up there instead.

SergioGasquez (commented):
Thanks for the contribution! Changes LGTM! Just left one nitpick suggestion (sorry). I was able to build both a std and a `no_std` with your changes. Mind elaborating which further testing you did?

kyrias:
> I was able to build both a std and a `no_std` with your changes. Mind elaborating which further testing you did?

Compile-testing isn't enough because they _build_ fine, but they silently fail to find the bootloader and partition table and silently fall back to the ones built into the binary.  To test it properly the simplest way is to flash it and look that it's using the correct partition table when it boots.  The bootloader is a bit trickier, though if one was found the other would've been found as well, but to explicitly test it the simplest way would be to add an extra log line to the bootloader and see that it's printed, I guess.

Sadly there's no way to see from the `espflash` side whether it used the built-in ones instead.  It would actually be nice if it was possible to configure it to _never_ use the built-in ones and instead return an error if it can't find a custom one, because if you _have_ a custom partition table it would definitely be an error condition to suddenly build without one.

SergioGasquez (approved):
LGTM! Thanks

Issue #617, opened 2024-03-20
`image_too_big` error when using espflash v3.0.0


armandas:
I tried to upgrade to v3.0.0, but it produces the following error:

```
[2024-03-20T08:04:58Z INFO ] Serial port: '/dev/ttyACM0'
[2024-03-20T08:04:58Z INFO ] Connecting...
[2024-03-20T08:04:58Z INFO ] Using flash stub
Chip type:         esp32s3 (revision v0.2)
Crystal frequency: 40 MHz
Flash size:        8MB
Features:          WiFi, BLE
MAC address:       48:27:e2:82:7d:58
Error: espflash::image_too_big (link)

  × Supplied ELF image of 1114592B is too big, and doesn't fit configured app partition of 1048576B
  help: Reduce the size of the binary or increase the size of the app partition.
```

V2.1.0 works fine:

```
[2024-03-20T07:58:18Z INFO ] 🚀 A new version of espflash is available: v3.0.0
[2024-03-20T07:58:18Z INFO ] Serial port: '/dev/ttyACM0'
[2024-03-20T07:58:18Z INFO ] Connecting...
[2024-03-20T07:58:18Z INFO ] Using flash stub
Chip type:         esp32s3 (revision v0.2)
Crystal frequency: 40MHz
Flash size:        8MB
Features:          WiFi, BLE
MAC address:       48:27:e2:82:7d:58
App/part. size:    1,114,592/8,323,072 bytes, 13.39%
[00:00:00] [========================================]      14/14      0x0                                                                                                                                                                    [00:00:00] [========================================]       1/1       0x8000                                                                                                                                                                 [00:00:08] [========================================]     650/650     0x10000                                                                                                                                                                [2024-03-20T07:58:28Z INFO ] Flashing has completed!
```

armandas:
Sure. I am using the `espflash` command, as described [here](https://github.com/esp-rs/esp-idf-template?tab=readme-ov-file#flash). My project is using `std`, but how does that affect the flashing tool?

`--no-stub` didn't help, but it did work with `--flash-size 8mb`. From the output in my previous post, however, we can see that the tool already knows there's 8MB of flash.

armandas:
Also just want to add that I haven't explicitly configured the partitions, but V2 had no problem with it, so from usability perspective, the new tool should not break either.

empirephoenix:
Hi, I'm running into the same issue since the last update of espflash

```
[...]

PR #633, opened 2024-05-02
#617 save-image with partition-table and without merge

empirephoenix:
allow specify partition table without merge, to create a larger than default partition ota image.
For more details see #617


SergioGasquez (approved):
LGTM! Thanks

empirephoenix:
Ok it seems to be fixed with the latest changes anyway, if I clone and install the 3.0.1-dev it just works, thanks

PR #634, opened 2024-05-04
Write the image to the correct target app partition

simpkins:
Perhaps I am misunderstanding what the --target-app-partition flag is supposed to do, but the code previously always wrote the image to the "factory" app partition if one was present, regardless of the partition specified with the `--target-app-partition` flag.

This change updates the code to write the image to the partition specified with the `--target-app-partition` flag.

I confirmed that using `espflash flash --target-app-partition ota_0` now causes espflash to write the image to the `ota_0` partition.  Note that this leaves the `factory` partition and any `ota_data` partition unchanged, so the bootloader will still boot from the factory partition by default, but this makes it possible to write new app images to the OTA partitions for testing purposes.


simpkins:
It looks like the `--target-app-partition` code was working correctly when added in #461, but it looks like it was probably broken shortly afterwards in d886d33a69 which updated the code to always write to the `factory` partition again.

SergioGasquez (commented):
Hi! Thanks for your contribution! Changes LGTM!

Mind solving the conflicts so we can merge it?

PR #639, opened 2024-05-24
Version 3.1.0

SergioGasquez:
- Update changelog
- Bump `espflash` and `cargo-espflash` version
- Update dependencies
- Bump MSRV to 1.76.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,21 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [3.1.0] - 2024-05-24
 
 ### Added
 - Support loading flash size, frequency, and mode from the config file (#627)
 
 ### Fixed
-- Fixed help text for size parameter of read-flash subcommand
+- Fixed help text for `size` parameter of `read-flash` subcommand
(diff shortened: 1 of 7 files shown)


MabezDev:
Ah I see you've done it in this PR - was there a reason to do so? If not IMO we should revert.

SergioGasquez:
> Does the MSRV bump affect the library, or just the binaries? If it affects the library then its a breaking change :(

I think it should affect `cargo-espflash` only, as the failing crate was `cargo`, see https://github.com/esp-rs/espflash/actions/runs/9221103611/job/25369434066

MabezDev:
Maybe we should make the MSRV check only check for the library then, and make no guarantee about the binaries (they can install from binstall or gh releases if they want stability)?

CHANGELOG.md
```diff
@@ -5,21 +5,20 @@ All notable changes to this project will be documented in this file.
...
 - Fixed port detection on `musl` when detection returns paths starting with `/dev/`
 - [cargo-espflash]: Always resolve package_id from metadata when finding bootloader and partition table (#632)
-- Fixed behavior of the --target-app-partition flag (#634)
+- Fixed behavior of the `--target-app-partition` flag (#634)
 
 ### Changed
 - Update ESP32, ESP32-C2, ESP32-C3, ESP32-C6, ESP32-H2, ESP32-S2, ESP32-S3 stub (#638)
-
-### Removed
+- MSRV bumped to `1.76.0` (#639)
```
MabezDev, line 21: I guess we can remove this too

MabezDev (approved):
LGTM, thanks!

PR #644, opened 2024-05-28
Add commands `reset` and `hold-in-reset`

jnross:
We found these commands to be useful for our team.


jnross:
Sorry for the noise here.  I cherry-picked this from a branch that was based on 2.1.0.  I'll clean it up shortly.

jnross:
Ok, the mess is cleaned up now.  Please let me know what you think.  I considered whether to add these commands to cargo-espflash as well but I can't think of a use case for it.  We generally use `cargo-espflash` for specialized purposes (like build & flash), and `espflash` is our more general-purpose tool to react to any given scenario.

SergioGasquez (commented):
Thanks for the contribution! The commands can be handy for some users! I would also add the commands for `cargo-espflash` as some user might not use both tools as you do.

jnross:
> Thanks for the contribution! The commands can be handy for some users! I would also add the commands for `cargo-espflash` as some user might not use both tools as you do.

Sounds good, I will do so.

jnross:
@SergioGasquez:  Thanks for your review!  I've updated this set of changes to also add these commands to `cargo-espflash`.

SergioGasquez (approved):

SergioGasquez:
Ups! I just noticed that we forgot to update the usage section of [cargo-espflash](https://github.com/esp-rs/espflash/blob/main/cargo-espflash/README.md#usage) and [espflash](https://github.com/esp-rs/espflash/blob/main/espflash/README.md#usage), mind updating it? Sorry

jnross:
@SergioGasquez:  I just pushed this fix.  Thanks for catching that!

SergioGasquez (approved):
LGTM! Thanks for the features!

PR #652, opened 2024-07-18
Add note about current development state

SergioGasquez:


README.md
```diff
@@ -8,6 +8,10 @@ Serial flasher utilities for Espressif devices, based loosely on [esptool.py](ht
 
 Supports the **ESP32**, **ESP32-C2/C3/C6**, **ESP32-H2**, **ESP32-P4**, and **ESP32-S2/S3**.
 
+> [!IMPORTANT]
+> Espressif considers espflash and cargo-espflash to be feature-complete at this time. This does not mean we will stop maintaining them; contributions are still welcome. However, we do not plan to pursue further development.
```
jessebraham, line 12:
```suggestion
> Espressif considers espflash and cargo-espflash to be feature-complete at this time. This does not mean we will stop maintaining them; contributions are still welcome. However, we do not plan to pursue further development at this point in time.
```

Sorry small nitpick, but IMO we should not make any strong commitments here either way. Perhaps it will make sense to continue development here again in the future, hard to say right now.

jessebraham (approved):
Thanks!

PR #661, opened 2024-08-02
Make monitor only send key on press, not on release

### LINKED ISSUE #660: Serial Monitor : Data is sent both for key press and release events
When trying out a simple serial echo with my new esp32, I found out it received twice the letters I typed when using espflash monitor.
However, when using PuTTY, it only received them once.

Using the original windows console allowed me to pinpoint the issue, keys are sent for press and for release event.
Adding a simple `dbg!(key)` here https://github.com/esp-rs/espflash/blob/45946f84e0b54043f4ac63bd520adccd36bf63a0/espflash/src/cli/monitor/mod.rs#L117-L124 showed me this (the "Read" are my echo program running on the esp)
```
[espflash\src\cli\monitor\mod.rs:119:5] &key = KeyEvent {
    code: Char(
        's',
    ),
    modifiers: KeyModifiers(
        0x0,
    ),
    kind: Press,
    state: KeyEventState(
        0x0,
    ),
}
Read: 's'
[espflash\src\cli\monitor\mod.rs:119:5] &key = KeyEvent {
    code: Char(
        's',
    ),
    modifiers: KeyModifiers(
        0x0,
    ),
    kind: Release,
    state: KeyEventState(
        0x0,
    ),
}
Read: 's'
```

I am not sure if this is intended, this can be fixed simply by checking if it is a key press or a key release.

I also want to add that this was difficult to find because inside some terminals, like the RustRover terminal or the new windows terminal, key presses were sending press and release instantaneously, even if the key was not yet released. The old windows terminal however correctly waited for me to release the key before sending the release event. Maybe this is due to a bug in an upstream library ?

LelouBil:
Fixes #660

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Downgrade crossterm and update time crates (#659)
+- Monitor now only sends key presses on key down events
 
 ### Changed
 
--- a/espflash/src/cli/monitor/mod.rs
+++ b/espflash/src/cli/monitor/mod.rs
@@ -15,6 +15,7 @@ use std::{
     time::Duration,
 };
 
+use crossterm::event::KeyEventKind;
 use crossterm::{
     event::{poll, read, Event, KeyCode, KeyEvent, KeyModifiers},
     terminal::{disable_raw_mode, enable_raw_mode},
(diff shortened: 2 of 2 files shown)


SergioGasquez (commented):
Hi! Thanks for your contribution! Do you mind rebasing/fixing the rustfmt error and sharing your testing results? 

This might be an improvement but, as you mentioned, there are differences between os/terminals, so I would like to test it properly.

LelouBil:
Test results, works for :
- IntelliJ/RustRover terminal
- PuTTY Serial
- Old Windows terminal (conhost.exe)
- New Windows terminal
Works is : key press sends character, release does not. Key repetition works if held for a bit

I think it's fine, I can't test it on Linux now, but since it works with PuTTY I assume it will work with anything. PuTTY is the most basic serial client you can find.

I might also try tracking the issue from some kind of upstream dependency I think, that added difficulty debugging this : the fact that on the new Windows terminal and rustrover terminal key press and release events were sent only on key press.

But for the espflash issue, I think we're good

SergioGasquez (approved):
LGTM! Thanks for the PR!

PR #666, opened 2024-08-05
Update bytemuck dependency

SergioGasquez:
- Update `bytemuck` as `1.16.0` was yanked
- Update `cargo` and update code accordingly as we now should use [`GlobalContext::default`](https://docs.rs/cargo/0.81.0/cargo/util/context/struct.GlobalContext.html) instead of  [`Config::default`](https://docs.rs/cargo/0.78.0/cargo/util/config/struct.Config.html#method.default)


MabezDev (approved):
LGTM!

PR #672, opened 2024-08-20
fix: do not contact crates.io to check for updates

ROMemories:
This reduces the number of dependencies and improves privacy of users, who may not expect a flashing tool to directly make HTTP requests. It also makes flashing faster when network is not available, as it is then not needed to wait for the HTTP connection to timeout.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1219,7 +1219,6 @@ dependencies = [
  "strum",
  "thiserror",
  "toml",
- "update-informer",
  "xmas-elf",
 ]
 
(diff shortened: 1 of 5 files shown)


MabezDev:
Thanks for the PR! Whilst I appreciate your concerns, we do this to reduce the issue reports from people using older versions. You are welcome to use espflash as a library and create your own CLI tool that meets your requirements.

A PR we would accept would be to add an `--offline` flag which disables this check to get the speed improvement on non-internet connected devices.

PR #680, opened 2024-09-12
Display timestamp, useful to find timing problem

davidqge:
I found timestamp in defmt documents, but following their instruction still don't show timestamp.

Spent a week figured out it's because espflash didn't display it.

This change made following work too

in Cargo.toml dependency:
embassy-time        ={ version = "0.3.1" , features = ["defmt-timestamp-uptime"] }

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Fixed
+- espflash defmt log didn't display timestamp, according to [defmt doc](https://defmt.ferrous-systems.com/timestamps).
 
 ### Changed
 
--- a/espflash/src/cli/monitor/parser/esp_defmt.rs
+++ b/espflash/src/cli/monitor/parser/esp_defmt.rs
@@ -1,9 +1,6 @@
 use std::io::Write;
 
-use crossterm::{
-    style::{Color, Print, PrintStyledContent, Stylize},
-    QueueableCommand,
-};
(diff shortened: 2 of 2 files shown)


SergioGasquez (commented):
Does this show the log level and line? Also, what is the format when not using embassy and the selected feature for timestamps?

davidqge:
> Does this show the log level and line?

it shows the log level. Using same display code from defmt.
Doesn't show the source line info. Probably need to call another function, that I didn't look into.


> Also, what is the format when not using embassy and the selected feature for timestamps?
Display rest of info without timestamp. 
You can also manually add timestamp according to [defmt doc](https://defmt.ferrous-systems.com/timestamps). Embassy feature just their way of implementation.

SergioGasquez:
Did some testing in the PR and changes the `defmt` format from:
image
To: 
image

IMHO, the new format lacks of a separator between the level and the message. Is that the default `dfmt` format that other platform are using?

davidqge:
> IMHO, the new format lacks of a separator between the level and the message. Is that the default `dfmt` format that other platform are using?

Yes, here is an image of [defmt](https://ferrous-systems.com/blog/knurling-rs/)
You can customize output if you want to, I just like to get timestamp to debug timing problems.

SergioGasquez (commented):
Mind resolving the conflicts?

jessebraham (approved):
I've resolved the conflict, changes seem fine to me. @SergioGasquez will give you final say on this and let you merge it if you approve :)

SergioGasquez (approved):
Thanks for the contribution! And thanks for resolving the conflict @jessebraham!

PR #682, opened 2024-09-17
Add value_parser so `partition-table-offset` accepts offset expressed in hexadecimal.

jnross:


SergioGasquez (approved):
Thanks for the contribution! Changes lgtm and did a small test where everything worked as expected!

PR #689, opened 2024-10-17
Add skip update check option

### LINKED ISSUE #688: Add an option to prevent checking for new version
Hi, I'm working behind a proxy and it seems the programs tries to connect to crates.io in order to find updates.
As the requests times out, it takes a lot of time to even scan a board.
Is it possible to add an option to prevent such connections?

Thanks in advance!

Thrathra:
Add a new global option to prevent espflash from checking updates.
It will make it work faster behind a proxy.

Closes #688


jessebraham (approved):
Thanks!

darconeous:
It would be great if this could be added to espflash.toml, or perhaps even an environment variable.

PR #693, opened 2024-10-24
If a port was specified via CLI or config, search *all* serial ports, not just the filtered list

### LINKED ISSUE #692: -p option does not work as expected
Without `--list-all-ports`, `detect_usb_serial_ports()` returns only `SerialPortType::UsbPort`. When the device is not recognized as `SerialPortType::UsbPort` for some reasons, `--port` will not work.

The help says "`--list-all-ports` List all available ports" but it is confusing. It does not "list ports". Instead, the option does different things. When a port is specified in `espflash.toml` or with `-p` and the port is found, it uses the found port without asking users. If not specified, it asks users to choose ports found by an underlying library. 

What I expect :

```console
> espflash board-info -p /dev/cuaU0     
```

should work.

The actual result:

```console
> espflash board-info -p /dev/cuaU0                 

  × The serial port '/dev/cuaU0' could not be found

> ls /dev/cuaU0
/dev/cuaU0
```

The workaround is, always use `--list-all-ports`.

If you cannot change the behavior, there should be an entry for it in `espflash.toml`.

```
[connection]
serial = "/dev/cuaU0"
list_all_ports = true
```

jessebraham:
Since the list of available ports were being filtered down by default, when specifying a port via CLI option or configuration file which had been filtered out, the port was reported as unavailable. This should resolve the issue by only filtering the list of ports if no port was specified.

@trombik are you able to test this branch out and confirm that it indeed resolves the issue for you, please?

Closes #692

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Only filter the list of available serial ports if a port has not been specified via CLI option or configuration file (#693)
+
 ### Removed
 
 ## [3.2.0]
--- a/espflash/src/cli/serial.rs
+++ b/espflash/src/cli/serial.rs
@@ -33,13 +33,14 @@ pub fn get_serial_port_info(
     // doesn't work (on Windows) with "dummy" device paths like `COM4`. That's
     // the reason we need to handle Windows/Posix differently.
 
-    let ports = detect_usb_serial_ports(matches.list_all_ports).unwrap_or_default();
-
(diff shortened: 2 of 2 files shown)


SergioGasquez (approved):
Changes LGTM! Let's wait for the confirmation that it fixes the issue. Thanks!

trombik:
Confirmed that the patch fixed the issue. Tested with `-p` and `espflash.toml`.

By the way, it is always impressive how quickly espressif responds to issues and reports. Other vendors should follow the same practice.

jessebraham:
Thank you for confirming the fix, and for the kind words! There are still some additional small changes I'd like to make, but I hope to publish new releases next week.

Issue #699, opened 2024-10-31
Allow partition table offset to be specified in `espflash.toml`, Cargo.toml `[package.metadata.espflash.flash]` or in the partition table itself


jnross:
For different devices, different partition table offsets may be used.  To allow this to be more explicitly configured, I propose that we allow the partition table to be specified in the `espflash.toml` file, or even in the partition table itself.  
Proposed `espflash.toml`:
```
[flash]
mode = "qio"
size = "8MB"
frequency = "80MHz"
partition_table_offset = 0x8000
```
Proposed `partition_table.csv`:
```
# ESP-IDF Partition Table
#----------------------------------------------------------------
# Name,         Type, SubType, Offset,   Size,     Flags
# bootloader                   0x0,      0xA000
partition_table                0xA000,   0x1000
otadata,        data, ota,     0xB000,   0x2000,
emul_efuse,     data, 5,       0xD000,   0x2000,
phy_init,       data, phy,     0xF000,   0x4000,
cps,            data, nvs,     0x13000,  0x8000,
nvs,            data, nvs,     0x1B000,  0x25000,
ota_0,          app,  ota_0,   0x40000,  0x3e0000,
ota_1,          app,  ota_1,   0x420000, 0x3e0000,

```
Proposed `Cargo.toml`:
```
[package.metadata.espflash]
partition_table = "../../partitions-dev.csv"
partition_table_offset = 0x8000
```

SergioGasquez:
Closed by #700

PR #700, opened 2024-11-01
Allow `partition_table_offset` to be specified in the config file. (for #699)

jnross:
Change to implement #699 

I tested this with `espflash.toml` in the local directory:
```
partition_table_offset = 0xA000

[flash]
size = "8MB"
```

It works as expected, I can see that when I change the partition_table_offset address in the config file that the tool's behavior honors that change.


jessebraham (approved):
LGTM, thanks!

PR #705, opened 2024-11-26
Add external processor support

bjoernQ:
This adds an option `--processors` which enables external executables to pre-process data received from the target in monitoring mode.

I have a couple of things in mind I want to build upon this since adding those functionalities to `espflash` itself would be problematic (in terms of growing code size, maintainability, and additional dependencies)

### DIFF
--- a/espflash/src/bin/espflash.rs
+++ b/espflash/src/bin/espflash.rs
@@ -30,7 +30,7 @@ pub struct Cli {
     subcommand: Commands,
 
     /// Do not check for updates
-    #[clap(short, long, global = true, action)]
+    #[clap(long, global = true, action)]
     skip_update_check: bool,
 }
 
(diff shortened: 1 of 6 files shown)


espflash/src/bin/espflash.rs
```diff
@@ -30,7 +30,7 @@ pub struct Cli {
     subcommand: Commands,
 
     /// Do not check for updates
-    #[clap(short, long, global = true, action)]
+    #[clap(long, global = true, action)]
```
bjoernQ, line 33: `short` clashed with some other option's short-name

SergioGasquez, line 33: I guess it clashed with `flash_size`, maybe we could use another short instead of removing it
```suggestion
    #[clap(short = 'S', long, global = true, action)]
```

SergioGasquez (commented):
LGTM! Is there any reason not to add this feature to `cargo-espflash` too?

bjoernQ:
> LGTM! Is there any reason not to add this feature to `cargo-espflash` too?

No real reason other than I never used it - can add it

SergioGasquez:
> No real reason other than I never used it - can add it

I think it would be good to add it to maintain feature parity as long as we can.

SergioGasquez (approved):
Thanks!

PR #716, opened 2024-12-29
Calculate end of bootloader to update SHA256

### LINKED ISSUE #715: `espflash save-image` does not update the App Image SHA256 in the correct location.
When generating a project with secure boot v2, the bootloader is padded: https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32c3/security/secure-boot-v2.html#secure-padding

Using a padded image with `espflash save-image --bootloader path-to-bootloader ...` will read the file into a buffer:

`flasher/mod.rs`:
```
impl FlashData {
    pub fn new(...) -> Result<Self, Error> {
        // If the '--bootloader' option is provided, load the binary file at the
        // specified path.
        let bootloader = if let Some(path) = bootloader {
            let data = fs::canonicalize(path)
                .and_then(fs::read)
                .map_err(|e| Error::FileOpenError(path.display().to_string(), e))?;

            Some(data)
        } else {
            None
        };
```

Then that bootloader buffer gets passed to `image_format.rs` `IdfBootloaderFormat`. From there, the header is updated to match the args so the sha256 needs updating. However, the location of the sha256 is naively calculated as being the last 32 bytes of the bootloader file. However, it's actually the last 32 bytes of the bootloader without padding. 

`image_format.rs`:
```
        // re-calculate hash of the bootloader - needed since we modified the header
        let bootloader_len = bootloader.len(); // <- size of the file including padding
        let mut hasher = Sha256::new();
        hasher.update(&bootloader[..bootloader_len - 32]); //  <-- wrong location of sha256
        let hash = hasher.finalize();
        bootloader.to_mut()[bootloader_len - 32..].copy_from_slice(&hash);
```

One approach, which I'm not sure is the correct one due to all the different environments but worked in my specific case, was to walk through the bootloader segments while adding up the length. Then adding width the 16-byte aligned 1-byte checksum at the end. Maybe a better approach can be derived by referencing `esptool/bin_image.py` directly.

https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32c3/api-reference/system/app_image_format.html#application-image-structures

> The image has a single checksum byte after the last segment. This byte is written on a sixteen byte padded boundary, so the application image might need padding.

Without this, secure-boot-enabled ESPs will not boot because the first-stage bootloader will detect the mismatch between the actual and old unchanged SHA256.

This appears to be related but not the same as #713

chris-subtlebytes:
The current implementation uses the last 32 bytes of the bootloader file. When Secure Boot V2 is enabled, the bootloader is padded. The new implementation walks through the segments to find the end and adds the 16-byte aligned 1-byte checksum to update the SHA256 instead of incorrectly updating the padding.

Closes #715

### DIFF
--- a/espflash/src/image_format.rs
+++ b/espflash/src/image_format.rs
@@ -4,6 +4,7 @@ use std::{borrow::Cow, io::Write, iter::once, mem::size_of};
 
 use bytemuck::{bytes_of, from_bytes, Pod, Zeroable};
 use esp_idf_part::{Partition, PartitionTable, Type};
+use log::debug;
 use sha2::{Digest, Sha256};
 
 use crate::{
@@ -135,9 +136,20 @@ impl<'a> IdfBootloaderFormat<'a> {
         };
 
         // fetch the generated header from the bootloader
-        let mut header: ImageHeader = *from_bytes(&bootloader[0..size_of::<ImageHeader>()]);
+        let mut calc_bootloader_size = 0;
+        let bootloader_header_size = size_of::<ImageHeader>();
+        calc_bootloader_size += bootloader_header_size;
+        let mut header: ImageHeader = *from_bytes(&bootloader[0..bootloader_header_size]);
         if header.magic != ESP_MAGIC {
             return Err(Error::InvalidBootloader);
+        }
+
+        for _ in 0..header.segment_count {
+            let segment: SegmentHeader = *from_bytes(
+                &bootloader
+                    [calc_bootloader_size..calc_bootloader_size + size_of::<SegmentHeader>()],
+            );
+            calc_bootloader_size += segment.length as usize + size_of::<SegmentHeader>();
         }
 
         // update the header if a user has specified any custom arguments
@@ -156,12 +168,26 @@ impl<'a> IdfBootloaderFormat<'a> {
             bytes_of(&header).iter().copied(),
         );
 
-        // re-calculate hash of the bootloader - needed since we modified the header
-        let bootloader_len = bootloader.len();
+        // The header was modified so we need to recalculate the hash of the
+        // bootloader. The hash is at the end of the bootloader segments and
+        // 1-byte checksum at the end of a 16-byte padded boundary.
+        //
+        // Source: Point 3 of https://docs.espressif.com/projects/esp-idf/en/v5.4/esp32c3/api-reference/system/app_image_format.html
+        calc_bootloader_size += 1; // add checksum size
+        calc_bootloader_size = calc_bootloader_size + ((16 - (calc_bootloader_size % 16)) % 16);
+        let bootloader_sha_start = calc_bootloader_size;
+        calc_bootloader_size += 32; // add sha256 size
+        let bootloader_sha_end = calc_bootloader_size;
+
         let mut hasher = Sha256::new();
-        hasher.update(&bootloader[..bootloader_len - 32]);
+        hasher.update(&bootloader[..bootloader_sha_start]);
         let hash = hasher.finalize();
-        bootloader.to_mut()[bootloader_len - 32..].copy_from_slice(&hash);
+        debug!(
+            "Updating bootloader SHA256 from {} to {}",
+            hex::encode(&bootloader[bootloader_sha_start..bootloader_sha_end]),
+            hex::encode(hash)
+        );
+        bootloader.to_mut()[bootloader_sha_start..bootloader_sha_end].copy_from_slice(&hash);
 
         // write the header of the app
         // use the same settings as the bootloader
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Update the app image SHA in the correct location for padded images (#715)
+
 ### Removed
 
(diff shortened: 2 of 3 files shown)


SergioGasquez (commented):
Briefly looking at your changes and argumentation, they look good to me! Ill wait until the PR is ready for review and Ill do a more thorough review and some testing.

Thanks for contributting to espflash!

chris-subtlebytes:
I left some print statements in there mostly to illustrate the issue. Do you want those removed? Are there any plans to add logging crates?

I briefly looked into creating some tests. Are there any existing test `Target`s floating around so that I can plug in a bunch of differently sized images?

jessebraham:
> Are there any plans to add logging crates?

Not sure what you mean, `log` and `env_logger` are already dependencies, we do have logging.

chris-subtlebytes:
log line added:

```
[2025-01-20T21:34:02Z INFO ] Updating bootloader SHA256 from 2034f85843702b555013b639087cbf4bd5d0ab800a8a15a7fe45715f5519d97f to 573853af4720e60f022158fae5b1a4aebc2279cec379c57cd64687708a00759e
```

espflash/src/image_format.rs
```diff
@@ -157,11 +169,24 @@ impl<'a> IdfBootloaderFormat<'a> {
...
+        let bootloader_sha_start = calc_bootloader_size;
+        calc_bootloader_size += 32; // add sha256 size
+        let bootloader_sha_end = calc_bootloader_size;
+
         let mut hasher = Sha256::new();
-        hasher.update(&bootloader[..bootloader_len - 32]);
+        hasher.update(&bootloader[..bootloader_sha_start]);
         let hash = hasher.finalize();
-        bootloader.to_mut()[bootloader_len - 32..].copy_from_slice(&hash);
+        log::info!(
```
SergioGasquez, line 184: This is more a debug log to me
```suggestion
        debug!(
```

espflash/src/image_format.rs
```diff
@@ -4,6 +4,7 @@ use std::{borrow::Cow, io::Write, iter::once, mem::size_of};
 
 use bytemuck::{bytes_of, from_bytes, Pod, Zeroable};
 use esp_idf_part::{Partition, PartitionTable, Type};
+use log;
```
SergioGasquez, line 7:
```suggestion
use log::debug;
```

To be consisten with other places of the code.

espflash/src/image_format.rs
```diff
@@ -157,11 +169,24 @@ impl<'a> IdfBootloaderFormat<'a> {
         );
 
         // re-calculate hash of the bootloader - needed since we modified the header
-        let bootloader_len = bootloader.len();
+        // the hash is at the end of the bootloader, but the bootloader bytes are padded.
+        // the real end of the bootloader is the end of the segments plus the 16-byte aligned 1-byte checksum.
+        // the checksum is stored in the last byte so that the file is a multiple of 16 bytes.
```
SergioGasquez, line 174: Is there any documentation/esptool source code links that we could add here?

SergioGasquez (commented):
Thanks for working on this! Did some testing and review and everything seems to be working, left some minor comments!

espflash/src/image_format.rs
chris-subtlebytes: https://docs.espressif.com/projects/esp-idf/en/v5.4/esp32c3/api-reference/system/app_image_format.html

> The image has a single checksum byte after the last segment. This byte is written on a sixteen byte padded boundary, so the application image might need padding.

How exactly do you want this added? I was struggling a bit with the current comment format. I prefer a capitalized sentence with punctuation so it's easier to write multiline comments but I'm trying to keep with the existing format here though. Also how should I like to the docs? I think the page is the same for all the devices but which one should I link to? Also finding the exact source can be hard if I just have the link, especially if you're searching for "16" instead of "sixteen". Do you want the relevant section pasted in as a comment too?

I think the esptool source wasn't so easily quotable.

SergioGasquez: > Something like this should work: 

```suggestion
        // the hash is at the end of the bootloader, but the bootloader bytes are padded.
        // the real end of the bootloader is the end of the segments plus the 16-byte aligned 1-byte checksum.
        // the checksum is stored in the last byte so that the file is a multiple of 16 bytes.
        // Source: Point 3 of https://docs.espressif.com/projects/esp-idf/en/v5.4/esp32c3/api-reference/system/app_image_format.html 
```

> I prefer a capitalized sentence with punctuation so it's easier to write multiline comments

Feel free to use this style, its used in other parts of espflash

> Also how should I like to the docs? I think the page is the same for all the devices but which one should I link to? 

I would recommend using C6, also, I suggest you take a look at https://github.com/esp-rs/esp-hal/blob/main/documentation/API-GUIDELINES.md#maintainability

> Do you want the relevant section pasted in as a comment too?

What do you mean? Copying what the docs say? If so, no, we don't want to duplicate information.

>  I think the esptool source wasn't so easily quotable.

Doesnt need to be quotable, we just need something to understand why we are doing it this way

chris-subtlebytes:
Ok it should be good now. I used esp32c3 because that's what the links point to in the header docs in the beginning of image_format.rs.

I also reworded the comment to be more clear and aligned with how the docs define it.

SergioGasquez (approved):
LGTM! Thanks for sticking with me through the review!

Issue #70, opened 2021-10-06
Support flash encryption


jessebraham:
Flash encryption is supported by all chips from the **ESP32** and newer. We can encrypt data on the host and flash it to the target, assuming we first burn the encryption key into the target's eFuse.

For more details:  
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/security/flash-encryption.html

jessebraham:
After some further discussion, it seems there's not really anything actionable here. I have written a short guide for using flash encryption in this gist:  
https://gist.github.com/jessebraham/0c7d86f9866614dc602e5d8e0e97c37a

If there is any specific functionality we need in the future we can open issues for that specifically.

ostenning:
@jessebraham in your gist, you need to manually build a secure flash bootloader with esp-idf idf.py, and then manually pass that bootloader binary to espflash. 

That seems fairly incomplete to me. Shouldn't `espflash` be able to prepare the bootloader for encryption?

PR #718, opened 2025-01-06
Adds support for flash encryption

4rzael:
# Introduction
This PR adds support for flash encryption, by adding a `--encrypt` flag to `flash` and `write-bin` commands.
 Requested by #70 

As written below, I still need to perform a few tests (waiting for new chips, I hope to have it done in a few days).

# Implementation

When performing encrypted writes, espflash will use non-compressed flashing instructions to the bootloader, as compression + encryption is not supported. Similarly, skip & verify features are disabled on encrypted partitions.

When passed to the `write-bin` command, espflash perform an encrypted flash directly.

The `encrypt` flag is stored in `WriteBinArgs`, and then directly passed to the created `RomSegment`.

When passed to the `flash` command, espflash will perform encrypted flashes on bootloader, partition, and application partitions.

The parameter flow is a bit more complex there: we start in `FlashArgs`, which is used to construct `FlashData`, which is used to construct `IdfBootloaderFormat`, which finally populates the appropriate `RomSegment`s.

# Tests:
- [x] write-bin with --encrypt
- [ ] write-bin without --encrypt
- [x] flash with --encrypt
- [ ] flash without --encrypt
- [ ] cargo-espflash flash with --encrypt
- [ ] cargo-espflash flash without --encrypt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support external log-processors (#705)
 - Make the `libudev` dependency optional with a new - enabled by default - feature: `libudev` (#709)
 - Address Clippy lints (#710)
+- Add support for flash-encryption (#718)
 
 ### Changed
 


jessebraham:
Thanks for the PR. Unfortunately, this include breaking changes and will require a new major release, however we are currently discussing a v4 release anyway, so probably this is not too much of an issue. If I can just ask you to be patient, once we come up with a game plan I can get back to you with regards to merging this. Hopefully within a week or two we will have a better idea of what our plans are moving forward.

4rzael:
Hi ! Thanks for the quick response.
Yes, I understand. I'll wait for your input as to how to get it merged to v4.

jessebraham:
We did some planning this week for the v4 release. I do not think we will start working on this quite yet, probably not for a few weeks still, but in the meantime I will publish one final v3 release on Monday, and then we can at least unblock this PR (and others).

ivmarkov:
@4rzael Please excuse my ignorance... I do not understand how `espflash` can write encrypted payload given that there is no encryption key around?

ivmarkov:
@4rzael Also I'm a bit lost as to what exactly is requested in #70 ?

My understanding as to what @jessebraham demonstrates in #70, is just one way to do encryption, which is - "first flash plaintext images, and then the bootloader will encrypt them on-the-fly" (that is, given that the bootloader is compiled with encryption support in the first place, which I doubt to be the case with the bootloaders provided by `espflash` itself and embedded inside it with `include_bytes!` or whatever). By the way, this is done by the bootloader by it doing a lot of (dangerous!) e-fuse operations, like generating a random encryption key, burning it, then read+write protecting it and a few others.

This is by the way (and to my understanding - I'm still debugging the bootloader so I might very well be wrong!) a **one time** operation? If you can confirm my understanding is correct?

Like, once it is done, you can't just burn plaintext images and hope the bootloader would encrypt them? Or am I missing something here? My understanding is, once the initial bootloader-driven encryption is done, that's it end of story. The bootloader does not repeat the process anymore, and will assume all new images you upload to be pre-encrypted. and since you don't know the encryption key, you can't pre-encrypt them in the first place? That is, given that the UART downloader was left enabled (which depends on whether you are in DEVELOPMENT or RELEASE mode w.r.t. encryption).

I'm myself following the "other" process, where the encryption key is "externally" provided, and the payload (bootloader, part-table, app image, nvs keys etc.) are pre-encrypted. Seems to be what Espressif recommends as well?: 
https://docs.espressif.com/projects/esp-idf/en/v5.4/esp32s3/security/security-features-enablement-workflows.html

If you have an opinion on this stuff - please share. I'm about to enter production with my approach...

4rzael:
@ivmarkov 
Hey. Sorry about the slow answer there.

> My understanding as to what @jessebraham demonstrates in https://github.com/esp-rs/espflash/issues/70, is just one way to do encryption, which is - "first flash plaintext images, and then the bootloader will encrypt them on-the-fly"

Yes. IIRC, this used to be the recommended/documented workflow if you didn't want to keep the keys.
The process you linked is equivalent (as long as you destroy the key afterwards), and a is bit faster in the production line (the other approach requires the esp encrypting the flash at first boot, which can easily take like 30s).

The goal of this PR, however, is to replicate esptool's encrypt feature, which is indeed only useful in development mode, where the bootloader's ability to perform encrypted-flashes is not disabled.
You're completely right about it being impossible to re-flash once all security options are enabled in release mode.

Basically, after encryption has been enabled (by whichever of the 2 methods you prefer), you can make the bootloader reflash a plaintext image, and encrypt it on the fly, allowing developers to flash firmwares on devices with encryption enabled (and thus as similar as possible to the end product).

4rzael:
@ivmarkov Yeah, sorry, wasn't super clear. What I'm talking about is the `--encrypt` argument in the `write_flash` command, which tells the bootloader to encrypt the supplied plaintext blob during the transfer.
While I haven't experimented with the workflow where you directly encrypt on the host, I agree with you that, AFAIK, it should be completely transparent to esptool.

SergioGasquez:
Hi! Do you have any updates on your tests? Would you like me to help testing?

PR #719, opened 2025-01-10
Version 3.3.0

jessebraham:
Opening as a draft for now, not going to merge/release until Monday.

Just bumped the dependencies, the point of this release is mostly just to get the current changes in `main` published while we begin work on the next major release. This unfortunately resulted in a pretty hefty bump in the MSRV, but I don't really feel like fighting with different versions to find some middle ground, so it is what it is.

A couple dependencies are still outdated, as we have had issues when trying to update them (`addr2line` and `crossterm`). This is something I'd like to address in v4, but again not going to bother with it right now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,6 @@ jobs:
     runs-on: ${{ matrix.platform.os }}
 
     steps:
-
       - uses: actions/checkout@v4
 
       - uses: ./.github/actions/setup-target
@@ -98,14 +97,13 @@ jobs:
     runs-on: ubuntu-22.04
 
     steps:
-
       - uses: actions/checkout@v4
 
       - uses: ./.github/actions/setup-target
         with:
           arch: ${{ matrix.platform.arch }}
           target: ${{ matrix.platform.target }}
-          toolchain: "1.74"
+          toolchain: "1.82"
 
       - run: cargo check -p espflash --lib
 
@@ -127,7 +125,6 @@ jobs:
     runs-on: ubuntu-22.04
 
     steps:
-
       - uses: actions/checkout@v4
 
       - uses: ./.github/actions/setup-target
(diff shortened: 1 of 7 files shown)


SergioGasquez (approved):
LGTM! Thanks for taking care of this

kevin-june:
On the subject of MSRV, though a bit of a tangent / some foreshadowing:

Yesterday I encountered this error when installing `espflash` onto a system that was running Rust 1.82:
```
Caused by:
  failed to parse manifest at `/Users/jenkins/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck_derive-1.9.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.82.0 (8f40fc59f 2024-08-21)).
```

After some digging around, it seems that `bytemuck_derive` had bumped its [MSRV to 1.84](https://github.com/Lokathor/bytemuck/blob/c254af30fcb57ca21269963b129f42574924f7c5/derive/Cargo.toml#L12) (with resolver set to 3) and released this in version 1.9.0. `espflash` pulls in this requirement through its dependency on `bytemuck` and its `derive` feature.

This version was yanked shortly after it was released, so the problem has gone away for now. For context, see the [bytemuck_derive's version history](https://crates.io/crates/bytemuck_derive/versions) and [bytemuck/issues/294](https://github.com/Lokathor/bytemuck/issues/294).

I still find it a bit confusing that the compiler suggested setting `edition2024` rather than noting that the MSRV of Rust 1.84 was not met, so I might be missing something....

In any case, this this failure will likely reappear when `bytemuck_derive` deploys 1.9.0 again. If/when this happens, I hope that this information will help!

jessebraham:
Thanks for the heads up. I have added a note to the installation instructions to suggest using `--locked` when installing, which I guess is about as much as we can do about that right now 😅

Issue #650, opened 2024-07-02
Add a flag to wait for serial port (Feature request)


mikhail-m1:
It would be nice to have a flag (something like `--wait`) to wait for serial port to appear in the system. I can make a PR if you are ready to merge it.

SergioGasquez:
Hi! Do you mind elaborating on your use case for this flag? I guess you want to run the command and then plug the board? Will the command ever timeout?

SergioGasquez:
Hi! Sorry for the late reply, if this is something that would help your scenario we would be happy to accept your contribution

PR #721, opened 2025-01-12
feat(`espflash`)  Add a flag to wait for serial port

0xb-s:
closes https://github.com/esp-rs/espflash/issues/650


SergioGasquez (commented):
Hi! Sorry for the delay. Is there any reason why this only applies to `monitor`, and not to all commands that need to connect to a target (eg: Add the flag in `ConnectArgs`)

Did some testing and `cargo r -r --bin espflash -- monitor --wait` fails with:
```
Error: espflash::no_serial

  × No serial ports could be detected
  help: Make sure you have connected a device to the host system. If the device is connected but not listed, try using the `--list-all-ports` flag.
```

Am I doing something wrong?

Thanks for your contribution!

PR #722, opened 2025-01-13
Begin next release cycle

jessebraham:
Will hold off on merging this for a little bit, just to be sure that we don't need a patch release.


SergioGasquez (approved):
LGTM!

Issue #730, opened 2025-01-28
Revisit HIL


SergioGasquez:
Revisit HIL CI and see how we can improve it. I would like it to cover more corner cases and more arguments. Maybe we need to add some `xtask` integration to make things easier?

Shall we also cover cargo-espflash with some ESP-IDF application?

We could also add a `checksum-md5` test

There are also some known issues (Already fixed):
- `flash test` needs to use `--no-skip` otherwise it won't flash anything:
```
[2025-01-28T11:22:09Z INFO ] Segment at address '0x0' has not changed, skipping write
22
[2025-01-28T11:22:09Z INFO ] Segment at address '0x8000' has not changed, skipping write
23
[2025-01-28T11:22:10Z INFO ] Segment at address '0x10000' has not changed, skipping write
```
- Some HIL runs are cancelled: https://github.com/esp-rs/espflash/actions/runs/13010293728

PR #732, opened 2025-01-29
Avoid running  hil twice and add `--no-skip` to espflash flash command

SergioGasquez:
Addresses the known issues reported in https://github.com/esp-rs/espflash/issues/730

Not sure if it will avoid the cancelled HIL run, but I copied the workflow triggers from esp-hal, where it doesnt happen

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,11 @@ name: CI
 on:
   pull_request:
   push:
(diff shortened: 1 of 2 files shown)


jessebraham (approved):
> Not sure if it will avoid the cancelled HIL run, but I copied the workflow triggers from esp-hal, where it doesnt happen

Only one way to find out, I guess 😁

PR #734, opened 2025-01-29
Update license files and top-level `README.md`, move test binaries to `tests/` directory

jessebraham:
This is mostly just housekeeping:

- Updated the copyright date in the license, reformatted Apache 2.0 license (since it was weird before, now matches what we have in `esp-hal`)
- Removed noticed from top-level `README.md` stating that we are not developing any more features (because we are, now)
- Moved test binaries to the `tests/` directory, removed old binaries which are no longer in use
    - I would like to keep resources required by the application and test artifacts separate :)


espflash/tests/README.md
```diff
@@ -1,6 +1,6 @@
 # Test Resources
 
-This document describes how the test files under `tests/resources` were generated, so that they can be re-generated in the future if needed.
+This document describes how the test files under `tests/data` were generated, so that they can be re-generated in the future if needed.
```
jessebraham, line 3: As a note, this document needs updating. However, we no longer have a `blinky` example in the HAL, so have just left it as-is for now :)

SergioGasquez: > As a note, this document needs updating

Shall we open an issue about this?

SergioGasquez (approved):
LGTM! Thanks

PR #742, opened 2025-02-05
Remove the dependency on `libudev` and its corresponding feature

### LINKED ISSUE #711: Investigate if we need to depend on libudev
`serialport` can now work without `libudev`, see https://github.com/serialport/serialport-rs/pull/220, this can lead to weird ports being shown but it might be worth to explore how this performs, some initial tests show good signs, see https://github.com/esp-rs/espflash/pull/709

jessebraham:
Opening as draft as I guess we will need to do some testing, but at least it builds fine. We'll see what happens with HIL, I don't have a Linux box handy so that will be the first check I guess.

Not sure how we want to go about testing/verifying this.

If we merge this, closes #711. If not, we can close the issue regardless I guess.


SergioGasquez:
> Not sure how we want to go about testing/verifying this.

Not a clean solution, but since this requires large testing in different OSs, what if we merge it (after some initial testing, Ill test it tomorrow morning) and we get input from more people. If it doesnt work, we can always revert this before publishing a new version.

jessebraham:
> > Not sure how we want to go about testing/verifying this.
> 
> Not a clean solution, but since this requires large testing in different OSs, what if we merge it (after some initial testing, Ill test it tomorrow morning) and we get input from more people. If it doesnt work, we can always revert this before publishing a new version.

Yeah I was thinking the same thing honestly. Let me sleep on it, will make a decision tomorrow.

SergioGasquez (approved):
LGTM! Thanks

PR #743, opened 2025-02-05
Add Taplo config file and format Cargo manifests

jessebraham:


jessebraham:
> LGTM, I wonder if we can integrate this into the esp-hal xtask to format the tomls there 👀

Possibly, something I (or somebody else) can definitely look in to.

PR #746, opened 2025-02-06
Check for conflicting command-line option names in CI

jessebraham:
Using the suggestion from https://github.com/clap-rs/clap/issues/3133, we can take advantage of `clap_complete` to make sure that none of the subcommands have conflicting option names. Slight increase in runtime for the workflow but worth the peace of mind IMO.

(Also re-ordered a few lines in the workflow definition while I was in there)


SergioGasquez (approved):
LGTM! Thanks

PR #749, opened 2025-02-07
Remove special handling of MUSL Linux now that we no longer use libudev

jessebraham:
Now that `serialport` enumerates more ports without the use of `libudev` (see https://github.com/serialport/serialport-rs/pull/238), perhaps port detection will work as expected on Linux systems using MUSL without the specialized function for these platforms.

Opening as a draft as I currently have no way to test this, if somebody from the community is able to verify port detection still works on Linux systems using MUSL that would very much be appreciated!

If things continue to work as expected, would be nice to get rid of this extra code. If not, no real loss I guess, and this PR can just be closed.

### DIFF
--- a/espflash/src/cli/serial.rs
+++ b/espflash/src/cli/serial.rs
@@ -105,73 +105,7 @@ fn find_serial_port(ports: &[SerialPortInfo], name: &str) -> Result<SerialPortIn
     }
 }
 
-/// Serialport's auto-detect doesn't provide any port information when using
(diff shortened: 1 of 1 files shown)


jessebraham:
@SergioGasquez what should we do here? Should we just merge it and deal with the fallout if there is any? I don't have a system to test this on currently.

SergioGasquez:
> Should we just merge it and deal with the fallout if there is any? 

I think its the best way to properly test this!

SergioGasquez (approved):

PR #750, opened 2025-02-07
Additional improvements to CI workflow

jessebraham:
This has pretty drastically reduced the runtime on my fork at least. Some additional re-org in here too, sorry for the noise.

Additional considerations:

- The matrices in the `check`, `check-lib`, and `msrv` jobs are different
  - Should we check all platforms/archs for each of these jobs, or is a limited subset adequate?
  - If the latter, which platforms should be checked? Is x86_64 macOS/Linux/Windows enough for the `check-lib` and `msrv` jobs? Something else?

### DIFF
--- a/.github/actions/setup-target/action.yml
+++ b/.github/actions/setup-target/action.yml
@@ -3,9 +3,11 @@ description: Setup an ARM or x86_64 build environment
 
 inputs:
   arch:
-    required: true
+    default: x86_64
+    required: false
   target:
-    required: true
+    default: x86_64-unknown-linux-gnu
+    required: false
   toolchain:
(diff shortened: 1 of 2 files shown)


jessebraham:
Guess we need to update the merge queue settings prior to merging this, was a little confused why it was waiting for jobs which don't exist for a minute there 😅

SergioGasquez:
> Should we check all platforms/archs for each of these jobs, or is a limited subset adequate?

I think we should be good with current tests! If not, we can revert this situation in the future

SergioGasquez (approved):
LGTM! Thanks

PR #755, opened 2025-02-10
Remove the dependency on `lazy_static` in favour of built-in `LazyLock`

jessebraham:
There has been a built-in solution for this problem in Rust for quite some time now, so no reason to depend on this package anymore.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1268,7 +1268,6 @@ dependencies = [
  "flate2",
  "hex",
  "indicatif",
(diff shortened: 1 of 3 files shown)


SergioGasquez (approved):
LGTM! Thanks, shall we add a changelog entry about removing the `lazy_static` dependency?

jessebraham:
The change is invisible to the user, it was behind the `cli` feature anyway which is not stable.

PR #756, opened 2025-02-10
Remove the dependency on `parse_int` in favour of `from_str_radix`

jessebraham:


jessebraham:
Not sure what's up with HIL 😅

SergioGasquez (approved):
LGTM! Just one comment, since we changed the function a few times in the last week I manually tested it twice, it would be good to add a test for it, although is simple and I dont expect it to change too much in the future

espflash/src/cli/mod.rs
```diff
@@ -290,9 +290,15 @@ pub struct ChecksumMd5Args {
     connect_args: ConnectArgs,
 }
 
-/// Parses an integer, in base-10 or hexadecimal format, into a [u32].
+/// Parses an integer, in base-10 or hexadecimal format, into a [u32]
 pub fn parse_u32(input: &str) -> Result<u32, ParseIntError> {
```
SergioGasquez, line 294: We could add something like: 
```rust

    #[test]
    fn test_parse_u32() {
        // Hex
        assert_eq!(parse_u32("0x1234").unwrap(), 0x1234);
        assert_eq!(parse_u32("0X1234").unwrap(), 0x1234);
        assert_eq!(parse_u32("0xabcd").unwrap(), 0xabcd);
        // Decimal
        assert_eq!(parse_u32("1234").unwrap(), 1234);
        assert_eq!(parse_u32("0").unwrap(), 0);
        // Errors
        assert!(parse_u32("").is_err());
        assert!(parse_u32("0x").is_err());
        assert!(parse_u32("0xg").is_err());
        assert!(parse_u32("-123").is_err());
        assert!(parse_u32("12.34").is_err());
    }
```

Issue #704, opened 2024-11-12
Add functionality similar to `get_security_info` command in `esptool`


jessebraham:

ivmarkov:
Nice to have, but I think we need the full-blown capability of `espefuse.py` instead. Otherwise, there is actually no way to burn the SecureBoot V2 public key signature into efuse in the first place, and this is the one single thing the bootloader cannot do automatically "on first boot". But also in general, relying on the bootloader-compiled-with-encryption-enabled to auto-generate encryption XTS key, burn it, encrypt all plaintext bootloader/part table/factory image on first boot is super dangerous (IMO).

Espressif recommends [this workflow instead](https://docs.espressif.com/projects/esp-idf/en/v5.4/esp32s3/security/security-features-enablement-workflows.html), which relies on explicit efuse burns to enable both secure boot and encryption. Much more predictable. And allows for sub-sequent re-flashing of the encrypted/secure-booted device, as long as you keep the encryption key, the secure boot V2 private key around, and the ROM downloader enabled (in secure mode or not).

jessebraham:
This was opened because there have been change with how chip detection is performed with newer devices (e.g. [for the P4](https://github.com/espressif/esptool/blob/master/esptool/targets/esp32p4.py#L42)) and detecting via magic value does not work reliably with all devices. I did not consider any other application of this functionality when opening the issue.

ivmarkov:
> This was opened because there have been change with how chip detection is performed with newer devices (e.g. [for the P4](https://github.com/espressif/esptool/blob/master/esptool/targets/esp32p4.py#L42)) and detecting via magic value does not work reliably with all devices. I did not consider any other application of this functionality when opening the issue.

Fair enough!

jessebraham:
I probably should have mentioned that in the original comment 😅

PR #758, opened 2025-02-10
Add get-security-info command

JurajSadel:
Closes https://github.com/esp-rs/espflash/issues/704

I've tried to follow esptool's implementation and output formatting as close as possible here.

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -372,8 +372,13 @@ pub fn connect(
 pub fn board_info(args: &ConnectArgs, config: &Config) -> Result<()> {
     let mut flasher = connect(args, config, true, true)?;
     print_board_info(&mut flasher)?;
-    let si = flasher.get_security_info()?;
-    println!("{si}");
+
+    if flasher.chip() != Chip::Esp32 {
+        let security_info = flasher.get_security_info()?;
+        println!("{security_info}");
+    } else {
+        println!("Security features: None");
+    }
 
     Ok(())
 }
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -372,6 +372,8 @@ pub fn connect(
 pub fn board_info(args: &ConnectArgs, config: &Config) -> Result<()> {
     let mut flasher = connect(args, config, true, true)?;
     print_board_info(&mut flasher)?;
+    let si = flasher.get_security_info()?;
+    println!("{si}");
 
     Ok(())
 }
--- a/espflash/src/connection/mod.rs
+++ b/espflash/src/connection/mod.rs
@@ -35,6 +35,7 @@ use crate::{
     command::{Command, CommandType},
     connection::reset::soft_reset,
     error::{ConnectionError, Error, ResultExt, RomError, RomErrorKind},
+    flasher::SecurityInfo,
 };
 
 pub mod reset;
@@ -88,6 +89,35 @@ impl TryInto<Vec<u8>> for CommandResponseValue {
             CommandResponseValue::ValueU128(_) => Err(crate::error::Error::InternalError),
             CommandResponseValue::Vector(value) => Ok(value),
         }
+    }
+}
+
+impl TryFrom<Vec<u8>> for SecurityInfo {
+    type Error = crate::error::Error;
+
+    fn try_from(res: Vec<u8>) -> Result<Self, Self::Error> {
+        let esp32s2 = res.len() == 12 + 4; // +4 means header
+
+        // Parse response bytes
+        let flags = u32::from_le_bytes(res[0..4].try_into().unwrap());
+        let flash_crypt_cnt = res[4];
+        let key_purposes: [u8; 7] = res[5..12].try_into().unwrap();
+
+        let (chip_id, eco_version) = if esp32s2 {
+            (None, None) // ESP32-S2 doesn't have these values
+        } else {
+            let chip_id = u32::from_le_bytes(res[12..16].try_into().unwrap());
+            let eco_version = u32::from_le_bytes(res[16..20].try_into().unwrap());
+            (Some(chip_id), Some(eco_version))
+        };
+
+        Ok(SecurityInfo {
+            flags,
+            flash_crypt_cnt,
+            key_purposes,
+            chip_id,
+            eco_version,
+        })
     }
 }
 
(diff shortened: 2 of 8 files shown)


jessebraham:
Thanks for this! Will give it a proper review soon, but just some initial thoughts:

- I had not considered creating a subcommand for this initially, sorry for not communicating this clearly! With that said, I don't think it's a bad thing to have (`esptool.py` has one like this), however maybe this overlaps a bit with the existing `board-info` command, and they should be folded into a single `get-security-info` command or something? @SergioGasquez what are your thoughts on this?
- Any subcommands added for `espflash` should also be added for `cargo-espflash` please, sorry again for not clearly communicating this beforehand 😅 

Anyway, thanks again and I will try to review this ASAP!

SergioGasquez:
I was also thinking of this to be an extra subcommand, but now that you mention it, I agree on the overlap. Maybe we could have a `--security-info` flag on `board-info` or just dump the security info in `board-info` by default (not sure if all users will find the flag for it). If we merge the two subcommands, I think `board-info` has a more intuitive name for it than `get-security-info`.

Now my question is if we should have two commands or merge them into one… I don't have hard opinion, but maybe I'd lean towards merging it with `board-info` and printing the current information and security info by default.

SergioGasquez:
Whatever we end up doing, we should add a check for it on HIL

jessebraham:
I guess if we decide we want to try and mirror their CLI then we have some refactoring to do

JurajSadel:
> I dont think we should imitate `esptool` command names, also I dont think this command will be used that often by end-users, the main functionality is to detect new chips. What do you guys think about merging it into `board-info`?

I'm okay with it.

espflash/src/error.rs
```diff
@@ -210,6 +210,9 @@ pub enum Error {
 
     #[error("Failed to parse partition table")]
     Partition(#[from] esp_idf_part::Error),
+
+    #[error("Invalid response length")]
```
jessebraham, line 214: Maybe it would be nice to include some additional information here, i.e. what about the response was invalid

SergioGasquez (commented):
Looks pretty good to me, thanks for working on this!

espflash/src/cli/mod.rs
```diff
@@ -363,6 +363,8 @@ pub fn connect(
 pub fn board_info(args: &ConnectArgs, config: &Config) -> Result<()> {
     let mut flasher = connect(args, config, true, true)?;
     print_board_info(&mut flasher)?;
+    let si = flasher.get_security_info()?;
+    println!("{si}");
```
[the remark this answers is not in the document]
JurajSadel: It does return an error because `get-security-info` got merged and that is not implemented for ESP32. This is the current log
```rust
Chip type:         esp32 (revision v3.0)
Crystal frequency: 40 MHz
Flash size:        4MB
Features:          WiFi, BT, Dual Core, 240MHz, Embedded Flash, Coding Scheme None
MAC address:       34:86:5d:dc:2c:40
Error: espflash::unsupported_feature

  × The esp32 does not support
  │ get-security-info
```

espflash/src/flasher/mod.rs
```diff
@@ -47,6 +47,132 @@ use crate::{
 #[cfg(feature = "serialport")]
 pub(crate) mod stubs;
 
+/// Security Info Response containing
+/// 32 bits flags, 1 byte flash_crypt_cnt, 7x1 byte key_purposes, 32-bit word
+/// chip_id, 32-bit word eco_version
```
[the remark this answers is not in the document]
JurajSadel: I didn't find more information about it.

espflash/src/error.rs
JurajSadel: I'm not sure what else could be added here from reading esptool docs/code

espflash/src/connection/mod.rs
```diff
@@ -91,6 +92,61 @@ impl TryInto<Vec<u8>> for CommandResponseValue {
...
+            .try_into()
+            .map_err(|_| Error::InvalidResponse)?;
+
+        let (chip_id, eco_version) = if esp32s2 {
+            (None, None) // ESP32-S2 doesn't have these values
+        } else {
+            if res.len() < 20 {
+                return Err(Error::InvalidResponse);
+            }
+            let chip_id = u32::from_le_bytes(
```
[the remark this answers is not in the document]
JurajSadel: I would prefer to leave it for future PR.

espflash/src/error.rs
jessebraham: Well, a lot of the response parsing can probably be simplified. Looking at the code I'm not really sure this error is really necessary. e.g. currently we have a bunch of instances of code like this:

```rust
res.get(0..4)
  .ok_or(Error::InvalidResponse)?
  .try_into()
  .map_err(|_| Error::InvalidResponse)?,
```

We have already verified the length of the response and know that we have enough bytes, so we can simply index into the bytes and remove the first error check altogether:

```rust
res.get[0..4]
  .try_into()
  .map_err(|_| Error::InvalidResponse)?,
```

From here, it's not clear to my how `try_into` can actually fail, if we already know we have the correct number of bytes 🤔

espflash/src/connection/mod.rs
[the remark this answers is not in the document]
jessebraham: Well, the entire purpose of adding this functionality in the first place was intended to make chip detection work correctly for all devices 😅

espflash/src/error.rs
JurajSadel: > the expected length and actual length would probably be useful pieces of information to include in the error.

It seems like different revisions could have different size of response (at least it's the case with S2), I don't see a good way how to do this.

jessebraham: Yeah not sure what's going on here. I have a v0.0 and running `board-info` works as expected but your script is failing for me. Will look into it tomorrow.

jessebraham: I stumbled across [this comment](https://github.com/espressif/esptool/blob/master/esptool/loader.py#L232) in `esptool.py`, adding `--no-stub` to your script it's now working for me. Not sure how it works for other chips without this (I'm obviously missing something here), but FYI.

espflash/src/connection/mod.rs
```diff
@@ -441,12 +484,20 @@ impl Connection {
                             RomErrorKind::from(response.error),
                         )))
                     } else {
-                        Ok(response.value)
-                    }
-                }
-                _ => {
-                    continue;
+                        // Check if the response is a Vector and strip header (first 8 bytes)
+                        // https://github.com/espressif/esptool/blob/release/v4/esptool/loader.py#L481: data = p[8:]
```
jessebraham, line 488: Please include a shorthash in any URLs you include linking to specific lines. No need to reproduce the linked line in the comment, either.

espflash/src/flasher/mod.rs
```diff
@@ -47,6 +47,135 @@ use crate::{
...
+                    revoked_keys.join("\n  ")
+                )?;
+            }
+        } else {
+            writeln!(f, "Secure Boot: Disabled")?;
+        }
+
+        // Flash Encryption
+        let flash_crypt_cnt_binary = format!("{:b}", self.flash_crypt_cnt);
+        if flash_crypt_cnt_binary.chars().filter(|&c| c == '1').count() % 2 != 0 {
```
jessebraham, line 146: You should be able to just use [count_ones](https://doc.rust-lang.org/std/primitive.u8.html#method.count_ones) here instead of converting to a string first, AFACT.

espflash/src/error.rs
JurajSadel: I think I got it - the reason here was, I didn't rebase my branch to use C stubs so that's why it worked for me locally and failed in CI. C stub gives different length of response for S2 compared to Rut stub.

jessebraham (changes requested):
Again, can you please upgrade `cargo-espflash` as well?

jessebraham (approved):

Issue #736, opened 2025-01-30
Flash and monitor commands do not use the same baudrate source


SergioGasquez:
The `flash` command has a `--monitor-baud` option and does not use the `env:ESPFLASH_BAUD`. The monitor command has a `--baud` option that does use the `env:ESPFLASH_BAUD`.

In `esptool`/`esp-idf-monitor` they use [`ESPTOOL_BAUD` for setting the flashing baudrate](https://docs.espressif.com/projects/esptool/en/latest/esp32/esptool/basic-options.html#baud-rate) and then [`IDF_MONITOR_BAUD`/`MONITORBAUD` for setting the monitoring baudrate](https://github.com/espressif/esp-idf-monitor/blob/d2c0d9cebafd28c3d30c9b1ba119e28adad67b48/esp_idf_monitor/base/argument_parser.py#L42C58-L42C69)

We should probably do something about this, either we make `ESPFLASH_BAUD` set the monitoring baudrate in all the commands or have another environment variable to set the monitoring baudrate. Id go for the later one.

bjoernQ:
> We should probably do something about this, either we make ESPFLASH_BAUD set the monitoring baudrate in all the commands or have another environment variable to set the monitoring baudrate. Id go for the later one.

Definitely should be separate 👍

PR #759, opened 2025-02-11
Normalize similar arguments

SergioGasquez:
I also removed some short arguments that didnt make much sense to me (I can revert those)

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -267,9 +267,9 @@ pub struct MonitorArgs {
 #[derive(Debug, Args)]
 #[non_exhaustive]
 pub struct MonitorConfigArgs {
-    /// Baud rate at which to communicate with target device
+    /// Baud rate at which to monitor the target device
     #[arg(short = 'r', long, env = "MONITOR_BAUD", default_value = "115_200", value_parser = parse_u32)]
-    pub baudrate: u32,
+    pub monitor_baud: u32,
     /// ELF image to load the symbols from
     #[arg(long, value_name = "FILE")]
     pub elf: Option<PathBuf>,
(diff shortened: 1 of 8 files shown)


SergioGasquez:
Shall I also edit the other places of the codebase where we use `offset`/`addr` instead of `address` and so on..?

SergioGasquez:
> Id like to come up with better names for monitoring args:
> 
> * `MonitorConfigArgs` has `baudrate` which can be set with `MONITOR_BAUD` environment variable
> * `ConnectArgs` has `baud` which can be set `ESPFLASH_BAUD` environment variable
> 
> Here is what esptool does: #736

For the moment, I'm using `baud` for `ConnectArgs` and `monitor_baud` for `MonitorConfigArgs`. But open to suggestions here

JurajSadel (approved):
LGTM

espflash/src/cli/mod.rs
```diff
@@ -277,7 +280,7 @@ pub struct MonitorConfigArgs {
     #[arg(long, requires = "non_interactive")]
     no_reset: bool,
     /// Logging format.
-    #[arg(long, short = 'L', default_value = "serial", requires = "elf")]
+    #[arg(long, short = 'L', default_value = "serial")]
```
SergioGasquez, line 283: I removed the `requires` because it was failing when using it as a custom runner, since you have the elf from the image, but dont have the `elf` arg.

You still get a pretty decent error when you do `espflash monitor --log-format defmt` and dont provide and elf:
```
❯ cargo r -r --bin espflash -- monitor --log-format defmt
...
Commands:
    CTRL+R    Reset chip
    CTRL+C    Exit

Error: espflash::monitor::defmt::no_elf

  × No elf data available
  help: Please provide an ELF file with the `--elf` argument
```
But you get that after selecting the port

SergioGasquez: Maybe we could investigate improving this in https://github.com/esp-rs/espflash/issues/764, since is somewhat similar

espflash/src/cli/monitor/mod.rs
```diff
@@ -87,7 +87,7 @@ pub fn monitor(
         reset_after_flash(&mut serial, pid).into_diagnostic()?;
     }
 
-    let baud = monitor_args.baud_rate;
+    let baud = monitor_args.monitor_baud;
```
bugadani, line 90: I think you can remove the `too_many_arguments` allow from this function.

espflash/src/cli/mod.rs
bugadani: It's also an issue in that `espflash flash <image> --elf <image>` is meaningless, but it's allowed.

jessebraham (approved):
Thanks!

PR #762, opened 2025-02-12
Update workflows to use newer Ubuntu release

jessebraham:
As of April 1st the `ubuntu-20.04` image will no longer be supported. We are already using `ubuntu-22.04` in the CI workflow, so I have updated the `hil` and `release` workflows to match.


jessebraham:
Ahh yes, forgot about those awful `GLIBC` errors 😅

PR #763, opened 2025-02-12
Swap out the flasher stubs with the legacy C stubs

### LINKED ISSUE #720: Update flasher stubs for all devices
Will update these once https://github.com/esp-rs/esp-flasher-stub/pull/74 is merged, just opening an issue so I don't forget.

jessebraham:
Opening as draft for now because:

1. Still not 100% sure we're going to do this (pretty sure we will, though)
2. Maybe adding a dependency on `serde_json` isn't worth it, and we should convert these to TOML instead, but it's easier to update them this way at least

Mostly just making sure HIL is still working by opening this now.

Closes #720


jessebraham:
Okay from the sounds of things we will move forward with these changes, so un-drafting. @SergioGasquez any strong opinions on the serialization format of the stub?

SergioGasquez:
> any strong opinions on the serialization format of the stub?
 
 Both solutions work for me, I think we are ok manually converting them from json to toml as we used to do before (IIRC), also, we dont expect those to change so its basically doing the conversion only once

SergioGasquez (approved):

PR #768, opened 2025-02-13
Minor docs improvement

JurajSadel:


jessebraham:
Going to be a nightmare to rebase my documentation branch on top of this 😂

jessebraham:
I think probably a lot of these things don't need to be public and therefore don't need to be documented. Was working on cleaning that up already, so not sure what we want to do with this.

jessebraham:
I think we will just close this for now, thanks anyway though!

Issue #753, opened 2025-02-07
erase-region should check the provided length


SergioGasquez:
`erase-region` should check that the provided length is multiple of the SPI flash erase sector size. In [esptool docs:](https://docs.espressif.com/projects/esptool/en/latest/esp32c3/esptool/basic-commands.html#erase-flash-erase-flash-erase-region)
> The address and length must both be multiples of the SPI flash erase sector size. This is 0x1000 (4096) bytes for supported flash chips.

```
❯ esptool.py --chip esp32c6 erase_region 0x1000 0x50
Serial port /dev/ttyACM0
Connecting...
Chip is ESP32-C6 (QFN40) (revision v0.0)
Features: WiFi 6, BT 5, IEEE802.15.4
Crystal is 40MHz
MAC: 60:55:f9:00:00:f7:1b:6c
BASE MAC: 60:55:f9:f7:1b:6c
MAC_EXT: 00:00
Uploading stub...
Running stub...
Stub running...
Erasing region (may be slow depending on size)...

A fatal error occurred: Size of data to erase must be a multiple of 4096
Status Ok(ExitStatus(unix_wait_status(512)))
```

While in `espflash`:
```
❯ espflash erase-region 0x1000 0x5
[2025-02-07T10:48:49Z INFO ] Serial port: '/dev/ttyACM0'
[2025-02-07T10:48:49Z INFO ] Connecting...
[2025-02-07T10:48:49Z INFO ] Using flash stub
[2025-02-07T10:48:49Z WARN ] Setting baud rate higher than 115,200 can cause issues
[2025-02-07T10:48:49Z INFO ] Erasing region at 0x00001000 (5 bytes)
```

We dont throw any error and we delete the next multiple of 4096 of the provided length, in the case of my example, 4096 (0x1000)

- [ ] Once this is resolved we should create a test for it on HIL

PR #771, opened 2025-02-14
erase-region: Check if the address and size is multiple of 4096 and add HIL test

JurajSadel:
closes https://github.com/esp-rs/espflash/issues/753

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -718,11 +718,11 @@ pub fn erase_region(args: EraseRegionArgs, config: &Config) -> Result<()> {
     }
 
     if args.address % FLASH_SECTOR_SIZE as u32 != 0 || args.size % FLASH_SECTOR_SIZE as u32 != 0 {
-        return Err(Error::InvalidEraseRegionArgument(format!(
-            "Address ({}) and size ({}) must be multiples of {}",
-            args.address, args.size, FLASH_SECTOR_SIZE
-        ))
-        .into());
+        return Err(Error::InvalidEraseRegionArgument {
+            address: args.address,
+            size: args.size,
+        })
+        .into_diagnostic();
     }
 
     let mut flasher = connect(&args.connect_args, config, true, true)?;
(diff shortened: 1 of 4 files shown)


espflash/tests/scripts/erase-region.sh
```diff
@@ -1,5 +1,33 @@
...
+    if [[ $result =~ "Invalid erase region argument" ]]; then
+        echo "Unaligned erase correctly rejected: address=$address, size=$size"
+    else
+        echo "Test failed: unaligned erase was not rejected!"
+        exit 1
+    fi
+}
+
+# Test cases for unaligned address/size
+SECTOR_SIZE=0x1000  # FLASH_SECTOR_SIZE is 4096 (0x1000)
```
SergioGasquez, line 19: This line is not being used (?)

```suggestion
```

espflash/src/cli/mod.rs
```diff
@@ -716,6 +717,14 @@ pub fn erase_region(args: EraseRegionArgs, config: &Config) -> Result<()> {
         return Err(Error::StubRequired).into_diagnostic();
     }
 
+    if args.address % FLASH_SECTOR_SIZE as u32 != 0 || args.size % FLASH_SECTOR_SIZE as u32 != 0 {
+        return Err(Error::InvalidEraseRegionArgument(format!(
+            "Address ({}) and size ({}) must be multiples of {}",
+            args.address, args.size, FLASH_SECTOR_SIZE
+        ))
+        .into());
+    }
```
SergioGasquez, line 726: I think we should have an error that takes two args (address and size), something like: 
```rust
    #[error("Invalid `address`({address})  and/or `size`({size}) argument(s)")]
    #[diagnostic(code(espflash::erase_region::invalid_argument)),
        help("`address` and `size`must be multiples of 0x1000 (4096)")]
    InvalidEraseRegionArgument {
        address: u32,
        size: u32,
    },

    #[error("Invalid erase region argument: {0}")]
    InvalidEraseRegionArgument(String),
```

SergioGasquez (commented):
Thanks for working on this! Left some comments

SergioGasquez (approved):
LGTM! Thanks!

PR #772, opened 2025-02-17
Refactor: make more modules private

jessebraham:
This will be the first of a handful of refactoring PRs. The goal is to minimize our API surface and only export what needs to be exported. Hopefully none of these changes are too controversial, I think it should be quite straight-forward.

Starting with some modules: the `command` and `elf` modules contain implementation details which are used internally, and don't need to be exposed to users. Additionally, the `error` modules only contains one type which needs to be public, `Error`, which is now exported at the root of the package.

In addition to these changes, I have done some general cleanup and removed some unused code.

I have also decided that we do not need to document the modules/types gated behind the `cli` feature. This is all intended for use only for the `cargo-espflash` and `espflash` binaries, it's really of no concern to people using `espflash` as a library.


espflash/src/lib.rs
```diff
@@ -14,7 +14,7 @@
 //! [espflash] can also be used as a library:
 //!
 //! ```toml
-//! espflash = { version = "2.1", default-features = false }
+//! espflash = { version = "4.0.0-dev", default-features = false }
 //! ```
```
SergioGasquez, line 18: We could suggest using `cargo add espflash --no-default-features` to avoid having to bump the version or have the version outdated

CHANGELOG.md
```diff
@@ -8,17 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- Add list-ports command to list available serial ports.
+- Add list-ports command to list available serial ports. (#761)
 
 ### Changed
+
 - Split the baudrate for connecting and monitorinig in `flash` subcommand (#737)
 - Normalized arguments of the CLI commands (#759)
-
 - `board-info` now prints `Security information`. (#758)
+- The `command`, `elf` and `error` modules are no longer public (#772)
```
SergioGasquez, line 22: Shall we mention that `command` module is now under `connection`?

SergioGasquez (approved):
LGTM! 2 nitpicks that can be ignored

espflash/src/lib.rs
jessebraham: Probably yeah, I'm planning on rewriting the module documentation anyway so can be addressed in the future.

CHANGELOG.md
jessebraham: Don't think it matters if nobody can access it anyway :)

Issue #697, opened 2024-10-25
Cannot read flash partition


DaneSlattery:
I cannot read flash partitions with espflash on an esp32-s3 devkit over UART. 

```
[2024-10-25T10:47:14Z INFO ] 🚀 A new version of espflash is available: v3.2.0
[2024-10-25T10:47:14Z INFO ] Serial port: '/dev/ttyUSB0'
[2024-10-25T10:47:14Z INFO ] Connecting...
[2024-10-25T10:47:14Z INFO ] Using flash stub
Chip type:         esp32s3 (revision v0.2)
Crystal frequency: 40 MHz
Flash size:        8MB
Features:          WiFi, BLE
MAC address:       30:30:f9:55:19:54
Error: espflash::read_flash::corrupt_data

  × Corrupt data, expected 1000 bytes but receved fd1 bytes
```

I am able to do the same with esptool.py:
```
esptool.py read_flash 0x9000 0x6000 ./nvs_dump.bin
esptool.py v4.8.1
Found 33 serial ports
Serial port /dev/ttyUSB0
Connecting....
Detecting chip type... ESP32-S3
Chip is ESP32-S3 (QFN56) (revision v0.2)
Features: WiFi, BLE, Embedded PSRAM 8MB (AP_3v3)
Crystal is 40MHz
MAC: 30:30:f9:55:19:54
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
24576 (100 %)
24576 (100 %)
Read 24576 bytes at 0x00009000 in 2.2 seconds (90.0 kbit/s)...
Hard resetting via RTS pin...

```

I don't know what is missing?

MabezDev:
Can you try again with `--no-stub`?

zakimoo:
same hier

```
espflash read-flash 0x0 0x0400000 test_3.bin

[2024-11-21T14:48:38Z INFO ] Serial port: '/dev/ttyUSB0'
[2024-11-21T14:48:38Z INFO ] Connecting...
[2024-11-21T14:48:38Z INFO ] Using flash stub
Chip type:         esp32c3 (revision v0.3)
Crystal frequency: 40 MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       a0:76:4e:5a:41:70
Error: espflash::read_flash::corrupt_data

  × Corrupt data, expected 1000 bytes but receved fcf bytes
```
but with esptool.py it works

```
[...]

PR #773, opened 2025-02-17
Improve HIL tests

SergioGasquez:
- Added a test to check `log-format defmt` arg
- Added a `list-ports` subcommand check
- Minor improvements to the workflow to avoid duplications
- Resolved the TODOs that were pending on https://github.com/esp-rs/espflash/issues/697


JurajSadel (approved):
LGTM, thank you for working on this!
Not related to the PR but we should add some timeouts to all tests.

SergioGasquez:
> Not related to the PR but we should add some timeouts to all tests.

Agree, created https://github.com/esp-rs/espflash/issues/774 to avoid 4 hours long flash erases

Issue #728, opened 2025-01-20
Document use via WSL2


jessebraham:
See this comment for details: https://github.com/esp-rs/espflash/issues/641#issuecomment-2408771592

PR #775, opened 2025-02-18
Mention WSL2 in READMEs

JurajSadel:
closes https://github.com/esp-rs/espflash/issues/728

### DIFF
--- a/cargo-espflash/README.md
+++ b/cargo-espflash/README.md
@@ -93,16 +93,16 @@ It is _not_ currently possible to use `cargo-espflash` from within WSL1. There a
 
 It is also _not_ possible to flash chips using the built-in `USB_SERIAL_JTAG` peripheral when using WSL2, because resetting also resets `USB_SERIAL_JTAG` peripheral, which then disconnects the chip from WSL2. Chips _can_ be flashed via UART using WSL2, however.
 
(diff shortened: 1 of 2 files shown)


cargo-espflash/README.md
```diff
@@ -93,6 +93,21 @@ It is _not_ currently possible to use `cargo-espflash` from within WSL1. There a
 
 It is also _not_ possible to flash chips using the built-in `USB_SERIAL_JTAG` peripheral when using WSL2, because resetting also resets `USB_SERIAL_JTAG` peripheral, which then disconnects the chip from WSL2. Chips _can_ be flashed via UART using WSL2, however.
 
+To be able to flash within WSL2, `systemd` should be enabled. To do so, create or edit `/etc/wsl.conf` and add the following:
```
SergioGasquez, line 96:
```suggestion
To be able to flash within WSL2, [`systemd` should be enabled](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#systemd-support). To do so, create or edit `/etc/wsl.conf` and add the following:
```

I think it's important to link the Windows documentation of how to do it. Also, they mention `sudo` might be required.

We could also mention that we need to restart the WSL instance(s), hence the shutdown command from bellow.

JurajSadel: Good idea!

jessebraham (approved):
Thanks!

Issue #774, opened 2025-02-17
Add default timeouts for HIL tests


SergioGasquez:
Test should fail if certain time is exceeded

PR #776, opened 2025-02-18
Add timeout to HIL tests

JurajSadel:
Not sure this is the best way but I like it more compared to editing all scripts and wrapping them in a timeout helpers. Maybe I'm missing other simpler option. 

closes https://github.com/esp-rs/espflash/issues/774


SergioGasquez (approved):
LGTM! My idea was to set [`timeout-minutes`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes) for the job, but the minimum timeout would be 1 min

PR #777, opened 2025-02-18
Add serial ports config file

SergioGasquez:
Don't really like the way I solved this as it's basically a bunch of duplicated code, but not sure if there's any better way. 

Draft until I do some more testing and/or I find a better solution

### DIFF
--- a/espflash/src/cli/config.rs
+++ b/espflash/src/cli/config.rs
@@ -80,18 +80,12 @@ pub struct Config {
     /// Bootloader path
     #[serde(default)]
     pub bootloader: Option<PathBuf>,
-    /// Preferred serial port connection information
-    #[serde(default)]
-    pub connection: Connection,
     /// Partition table path
     #[serde(default)]
     pub partition_table: Option<PathBuf>,
     /// Partition table offset
     #[serde(default)]
     pub partition_table_offset: Option<u32>,
-    /// Preferred USB devices
-    #[serde(default)]
-    pub usb_device: Vec<UsbDevice>,
     /// Flash settings
     #[serde(default)]
     pub flash: FlashSettings,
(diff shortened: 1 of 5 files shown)


espflash/src/cli/config.rs
```diff
@@ -164,6 +158,72 @@ impl Config {
     }
 }
 
+/// Deserialized contents of a serial port configuration file
+#[derive(Debug, Deserialize, Serialize, Default, Clone)]
+pub struct PortConfig {
+    /// Preferred serial port connection information
+    #[serde(default)]
+    pub connection: Connection,
+    /// Partition table path
```
JurajSadel, line 167:
```suggestion
```
seems redundant?

jessebraham:
Not sure this is the path we want to take to be honest. Had some other ideas I've been meaning to explore, so let's sit on this for awhile please.

espflash/README.md
```diff
@@ -131,49 +134,56 @@ or `cargo add espflash --no-default-features`
...
-    serial = "/dev/ttyUSB0"
-    ```
-  - By USB VID/PID values:
-    ```toml
-    [[usb_device]]
-    vid = "303a"
-    pid = "1001"
-    ```
-- Baudrate:
+## Configuration Files
```
SergioGasquez, line 137: Before merging, we would need to document this also in cargo-espflash

espflash/README.md
```diff
@@ -131,49 +134,56 @@ or `cargo add espflash --no-default-features`
...
+
+The are configuration files allowing you to define various parameters for your application:
```
JurajSadel, line 139: This sentence doesn't make sense, probably you meant something like
```suggestion
There are configuration files allowing you to define various parameters for your application:
```

JurajSadel:
I wonder if we can test this somehow in HIL

MabezDev:
It seems to me that we can avoid the code duplication by loading the two things separately then combining them again in code. We just need to be careful when saving the config again to write out the config back to the two separate files. After that, I think this becomes a relatively simple change?

Here is a small snippet to explain what I mean:

```rust
// package both together for code locality reasons, but they are seperate configs which are saved seperately
struct Config {
  config: ProjectConfig, // currently this is Config, let's rename it
  port_config: PortConfiguration // user port config
} 
```

What do you think?

espflash/src/cli/config.rs
```diff
@@ -75,91 +75,133 @@ impl UsbDevice {
...
+pub struct PortConfig {
+    /// Preferred serial port connection information
+    #[serde(default)]
+    pub connection: Connection,
+    /// Preferred USB devices
+    #[serde(default)]
+    pub usb_device: Vec<UsbDevice>,
     /// Path of the file to save the configuration to
     #[serde(skip)]
     save_path: PathBuf,
```
SergioGasquez, line 113: Only PortConfig requires `save_path`, as its the only config file that can be saved, `save_with` is only called here: https://github.com/esp-rs/espflash/blob/main/espflash/src/cli/serial.rs#L53

SergioGasquez:
> What do you think?

Definitely better, thanks for the suggestion @MabezDev, just updated it!

PR #778, opened 2025-02-18
Update `addr2line` and `crossterm` dependencies

### LINKED ISSUE #729: Update `addr2line` and `crossterm` dependencies
We have held off on updating these due to A) API changes in `addr2line` and B) issues with updating `crossterm` on Windows, but we should make an effort to resolve these issues and get back to using the latest versions.

jessebraham:
Updates `addr2line` to the latest published version. Did some basic testing against `espflash@main` and at least the behaviour has not changed with this PR, so I'll call it working.

@bjoernQ mentioned that updating `crossterm` did not cause issues for him, so I've just gone ahead and updated it without any further changes.

Closes #729

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,26 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


JurajSadel (approved):
Thanks!

Issue #691, opened 2024-10-23
Implement new USB SERIAL JTAG reset strategy


MabezDev:
As seen here: https://github.com/espressif/esptool/commit/8298cdcc68a8063df1ca932d42e21555e15269b3

This should solve the spurious issues where flashing is successful, but the chip stays in download mode.

Should close https://github.com/esp-rs/espflash/issues/556

Issue #556, opened 2024-01-19
esp32c6: Waiting for download _after_ flashing


t-moe:
I have a esp32c6 here that I flash using the command `espflash flash --monitor --flash-size 8mb --partition-table ./partitions.csv <elf>`

Sometimes I have the situation that after flashing the monitor shows:

```                                                                                                                                                                                                                        [2024-01-19T18:00:02Z INFO ] Flashing has completed!
Commands:
    CTRL+R    Reset chip
    CTRL+C    Exit

ESP-ROM:esp32c6-20220919
Build:Sep 19 2022
rst:0x15 (USB_UART_HPSYS),boot:0x4 (DOWNLOAD(USB/UART0/SDIO_FEI_FEO))
Saved PC:0x4080041e
0x4080041e - get_buffer_malloc
    at /tmp/backtrace-std/.embuild/espressif/esp-idf/master/components/spi_flash/spi_flash_os_func_app.c:193
waiting for download
```

Why is it still waiting for download? I have **not** pulled the strapping pin GPIO9 to low. 

I have not yet found out when this happens exactly, but it seems to somehow depend on the firmware I flash...

t-moe:
Minimal example // repro:

1. start with an erased esp32c6 (I'm using a esp32-c6-devkit-c1 v1.2 here)
2. Run `cargo generate esp-rs/esp-idf-template` to generate a hello world with std, based on idf v5.1.
3. Connect the target via JTAG USB. The pins are not connected to any external components, just a naked dev kit. (hook up the uart0 pins to the logic analyzer if you like)
4. Flash this to the target with `espflash flash --monitor <elf>`  ([flashproblems.elf.zip](https://github.com/esp-rs/espflash/files/14010380/flashproblems.elf.zip) )
5. So far everything works and the monitor prints the hello world. I can also press CTRL+R, still works
6. Exit the monitor (from step 4), then run `espflash monitor --no-stub`

espflash shows:
```
espflash monitor --no-stub
.... (version check with crates.io)
[2024-01-22T12:54:31Z INFO ] Serial port: '/dev/ttyACM0'
[2024-01-22T12:54:31Z INFO ] Connecting...
[2024-01-22T12:54:31Z DEBUG] Attempting flash enable with: SpiAttachParams { clk: 0, q: 0, d: 0, hd: 0, cs: 0 }
[2024-01-22T12:54:31Z DEBUG] Flash detect OK!
Commands:
    CTRL+R    Reset chip
[...]

PR #779, opened 2025-02-21
Implement new USB Serial JTAG reset strategy

JurajSadel:
This is mostly experimental, so opening as a draft. 

I've been playing with (cc cc https://github.com/esp-rs/espflash/issues/691)  and I was able to reproduce the issue with espflash AND esptool as well, following the same steps as mentioned [here](https://github.com/esp-rs/espflash/issues/556#issuecomment-1921143340)
I looked into esptool and specifically to [this commit](https://github.com/espressif/esptool/commit/8298cdcc68a8063df1ca932d42e21555e15269b3#diff-24560e81a6dac1814c9801703e7c13e2983585356da84fedc207ac9437c42dab). `esptool` not using this fix as default for SERIAL_JTAG, it can be used manually with: 
`--after reset_watchdog`

The reasons (from esptool perspective) it's not on by default are: 
1. The RTC WDT hack is not available on all devices (e.g., the H2 doesn't have it).
2. On Linux, it causes the port to re-enumerate, meaning that instead of /dev/ttyUSB0, a new /dev/ttyUSB1 appears. This can cause the monitor to disconnect and not reconnect automatically.
3. On the C6 and H2 USB-Serial/JTAG controller, there is a bug that causes the port to disappear after a system reset and not reappear until a power cycle is performed. The RTC WDT reset is a system reset, so this issue occurs. So they left it by default only for USB-OTG and is optional for SERIAL_JTAG. On C6 it is NOT SUPPORTTED at all in esptool.

### DIFF
--- a/espflash/src/connection/mod.rs
+++ b/espflash/src/connection/mod.rs
@@ -298,7 +298,9 @@ impl Connection {
 
                 match chip {
                     Chip::Esp32c3 => {
-                        wdt_reset(chip, self)?;
+                        if pid == USB_SERIAL_JTAG_PID {
+                            wdt_reset(chip, self)?;
+                        }
                     }
                     Chip::Esp32s2 => {
                         let esp32s2 = esp32s2::Esp32s2;
--- a/espflash/src/connection/mod.rs
+++ b/espflash/src/connection/mod.rs
@@ -13,6 +13,7 @@ use std::{
 
 use log::{debug, info};
 use regex::Regex;
+use reset::wdt_reset;
 use serialport::{SerialPort, UsbPortInfo};
 use slip_codec::SlipDecoder;
 
@@ -33,7 +34,10 @@ use self::{
         UsbJtagSerialReset,
     },
 };
-use crate::error::{ConnectionError, Error, ResultExt, RomError, RomErrorKind};
+use crate::{
+    error::{ConnectionError, Error, ResultExt, RomError, RomErrorKind},
+    targets::{esp32s2, esp32s3, Chip},
+};
 
 pub(crate) mod command;
 pub(crate) mod reset;
@@ -274,7 +278,7 @@ impl Connection {
     }
 
     // Reset the device taking into account the reset after argument
-    pub fn reset_after(&mut self, is_stub: bool) -> Result<(), Error> {
+    pub fn reset_after(&mut self, is_stub: bool, chip: Chip) -> Result<(), Error> {
         let pid = self.usb_pid();
 
         match self.after_operation {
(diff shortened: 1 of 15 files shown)


espflash/src/connection/mod.rs
```diff
@@ -277,6 +281,39 @@ impl Connection {
...
             }
+            ResetAfterOperation::WatchdogReset => {
+                info!("Resetting device with watchdog");
+
+                match chip {
+                    Chip::Esp32c3 => {
+                        wdt_reset(chip, self)?;
+                    }
+                    Chip::Esp32s2 => {
+                        let esp32s2 = esp32s2::Esp32s2;
```
MabezDev, line 292: You can use [match binding](https://doc.rust-lang.org/rust-by-example/flow_control/match/binding.html)'s to avoid this temporary variable

MabezDev: Ah sorry I missed that `esp32s2::Esp32s2` is a struct, not the enum we're matching on. Ignore me.

espflash/src/connection/reset.rs
```diff
@@ -294,6 +346,29 @@ pub fn soft_reset(
...
+    };
+
+    connection.write_reg(chip.wdt_wprotect(), chip.wdt_wkey(), None)?;
+    connection.write_reg(chip.wdt_config1(), 5000, None)?;
+    connection.write_reg(
+        chip.wdt_config0(),
+        (1 << 31) | (5 << 28) | (1 << 8) | 2,
+        None,
+    )?;
+    connection.write_reg(chip.wdt_wprotect(), 0, None)?;
```
SergioGasquez, line 367: Can we add some links/pointer to make this less black magic?

jessebraham: Probably wouldn't be a terrible idea to use [bitflags](https://docs.rs/bitflags/latest/bitflags/) for this, probably some other places we should be using this as well.

JurajSadel:
Should I add a P4 here as well?

espflash/src/connection/mod.rs
```diff
@@ -277,6 +281,48 @@ impl Connection {
...
+                        let esp32s3 = esp32s3::Esp32s3;
+                        if pid == USB_SERIAL_JTAG_PID || esp32s3.connection_is_usb_otg(self)? {
+                            // Check the strapping register to see if we can perform RTC WDT
+                            // reset
+                            if esp32s3.can_wtd_reset(self)? {
+                                wdt_reset(chip, self)?;
+                            }
+                        }
+                    }
+                    _ => println!("Unsupported chip for watchdog reset {:?}", chip),
```
SergioGasquez, line 321: This should probably be an error instead of a `println!`

JurajSadel: True

espflash/src/connection/reset.rs
```diff
@@ -199,6 +200,73 @@ impl ResetStrategy for UsbJtagSerialReset {
...
 }
 
+trait RtcWdtReset {
+    fn wdt_wprotect(&self) -> u32;
+    fn wdt_wkey(&self) -> u32;
+    fn wdt_config0(&self) -> u32;
+    fn wdt_config1(&self) -> u32;
+}
+
+impl RtcWdtReset for crate::targets::esp32c3::Esp32c3 {
```
jessebraham, line 210: I would really like to avoid leaking chip-specific details like these into other modules, this all probably belongs in the `targets` module.

espflash/src/targets/esp32p4.rs
```diff
@@ -30,9 +30,19 @@ const PARAMS: Esp32Params = Esp32Params::new(
...
 
 impl Esp32p4 {
+    /// Check if the magic value contains the specified value
     pub fn has_magic_value(value: u32) -> bool {
         CHIP_DETECT_MAGIC_VALUES.contains(&value)
     }
+
+    #[cfg(feature = "serialport")]
+    /// Check if the connection is USB OTG
+    pub(crate) fn connection_is_usb_otg(&self, connection: &mut Connection) -> Result<bool, Error> {
```
jessebraham, line 40: This function feels like it belongs on the `Connection` struct, so we should probably expose the necessary details from the `targets` module which can then be referenced there.

espflash/src/connection/mod.rs
```diff
@@ -277,6 +281,53 @@ impl Connection {
                 info!("Staying in flasher stub");
                 Ok(())
             }
+            ResetAfterOperation::WatchdogReset => {
+                info!("Resetting device with watchdog");
+
+                match chip {
+                    Chip::Esp32c3 => {
+                        if pid == USB_SERIAL_JTAG_PID {
```
SergioGasquez, line 289: #795 introduced a `is_using_usb_serial_jtag` method for this

espflash/src/targets/esp32p4.rs
SergioGasquez: Agree, also #795 introduced `is_using_usb_serial_jtag`, maybe this should be `is_using_usb_otg`

JurajSadel: I moved `connection_is_usb_otg` to `Connection` but I;m not sure about the others since most of them are not implemented for all chips

jessebraham:
Still not 100% happy with this, but I think it's in much better shape now at least. Remaining concerns are fairly minor, can be handled at another time (or just ignored 😁).

SergioGasquez (approved):
LGTM! Thanks

Issue #684, opened 2024-09-20
Add `--monitor` option to `write-bin` command


aiotter:
I'm using espflash for Rust based and non-Rust based ESP32 project. It is great.
However when I'm debugging non-Rust based app like Erlang with AtomVM, I'm stressed a bit to execute `espflash monitor` just after `espflash write-bin` again and again.

How about adding `--monitor` option to `write-bin` just like `flash` command?

It looks `cli::FlashArgs::{monitor,monitor_baud,log_format}` is safe to be copied into `WriteBinArgs`, and the same logic [here](https://github.com/esp-rs/espflash/blob/62cec888d732ba6a6148e7e8eff1884a6bc701c2/espflash/src/bin/espflash.rs#L281-L303) can be added at the end of `write_bin` function as well.

aiotter:
I don't expect that it is possible, because `espflash flash` is for ELF format.
I believe the binary file to be flashed into ESP32 is not in ELF format.

```bash
$ file MyProject.avm
MyProject.avm: a /usr/bin/env AtomVM script executable (binary data)
```

On AtomVM project, you write a program first with Erlang, Elixir or any other compatible languages, and then compiles it into BEAM format as usual.
AtomVM's `packbeam` utility will then packs it into AVM binary, which you can `espflash write-bin` into ESP32's `0x210000` address. AtomVM, flashed at bootloader partition in advance, will execute that AVM binary.

Do you think I can use `flash` command in my case?


> Regardless, I think we'd accept a PR that extended the monitor flag to `write-bin`.

Thank you for mentioning this! I'll try implementing it and open a PR when I have some time, hopefully.

PR #783, opened 2025-02-25
add `--monitor` option to `write-bin`

playfulFence:
closes https://github.com/esp-rs/espflash/issues/684

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add an environment variable to set monitoring baudrate (`MONITOR_BAUD`) (#737)
 - Add list-ports command to list available serial ports. (#761)
 - [cargo-espflash]: Add `write-bin` subcommand (#789)
-- Add --monitor option to write-bin. (#783)
+- Add `--monitor` option to `write-bin`. (#783)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


SergioGasquez (changes requested):
Changes LGTM! But Id love to have some HIL for this, at the moment, we do the following:
https://github.com/esp-rs/espflash/blob/main/.github/workflows/hil.yml#L114-L117, we could avoid using the `monitor.sh` by using the `--monitor` flag on the `save-image_write-bin.sh` and verifying the output!

espflash/tests/scripts/save-image_write-bin.sh
```diff
@@ -12,8 +12,12 @@ if [[ ! $result =~ "Image successfully saved!" ]]; then
     exit 1
 fi
 echo "Writing binary"
-result=$(espflash write-bin 0x0 app.bin 2>&1)
+result=$(timeout 60 espflash write-bin --monitor 0x0 app.bin --non-interactive 2>&1)
```
SergioGasquez, line 15: We already set the timeout in `hil.yml`, is there any reason to have this timeout?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add an environment variable to set monitoring baudrate (`MONITOR_BAUD`) (#737)
 - Add list-ports command to list available serial ports. (#761)
 - [cargo-espflash]: Add `write-bin` subcommand (#789)
+- Add --monitor option to write-bin. (#783)
```
SergioGasquez, line 17:
```suggestion
- Add `--monitor` option to `write-bin`. (#783)
```

PR #780, opened 2025-02-24
Fix write-bin for files whose lengths are not divisible by 4 by padding FF bytes

JurajSadel:
closes https://github.com/esp-rs/espflash/issues/698

Issue #622, opened 2024-04-07
`write-bin` does not flash binary


TheOnAndOnlyZenomat:
Hey,
I have a binary
```
00000000: 01a0                                     ..
```
which is just a jump to `0x0`.

When I try to write this binary with
```bash
espflash write-bin 0 img.img
```
nothing gets actually written. The output looks normal
```
[2024-04-07T17:54:34Z INFO ] Serial port: '/dev/ttyUSB0'
[2024-04-07T17:54:34Z INFO ] Connecting...
[2024-04-07T17:54:34Z INFO ] Using flash stub
Chip type:         esp32c3 (revision v0.3)
Crystal frequency: 40 MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       60:55:f9:c8:a7:6c
[00:00:00] [========================================]       1/1       0x10000                                                                        
[2024-04-07T17:54:36Z INFO ] Binary successfully written to flash!
```
but when reading the memory back using
```bash
espflash read-flash 0 64 out.bin
```
it is just all `0xff`.

When I try to flash the same binary with `esptool.py`
```bash
esptool.py write_flash 0 img.img
```
it works.
I can read back the memory with `espflash` and `esptool.py` and both show `01a0` at `0x0` in the flash.

Am I missing the point of the subcommand, or is this actually a bug. If the latter is the case, I'd be happy to do some debugging and testing for you.

MabezDev:
Hmm, looks like a bug. It's ignoring your input address of `0x0` and proceeding with the default application address of `0x10000`.

jordonsc:
I don't believe it's writing to the wrong location, if I have good data in the provided address, after using `espflash write-bin` that data is replaced by 0xFF bytes as if it was erased. 

The success of `write-bin` is dependent on my payload. If I change the payload slightly, it might succeed. Change again, it might fail - but it is deterministic.

PR #788, opened 2025-02-26
Fix write-bin padding

SergioGasquez:
While testing if https://github.com/esp-rs/espflash/issues/622 was resolved by #780, I noticed that #780 failed to flash a non-4-padded bin on the first try, the second would work as it would write the padding to the bin.

I updated the padding mechanism to not write the binary and it now works also on the first iteration.

Also added a HIL test.


JurajSadel (approved):
Thanks!

espflash/tests/README.md
SergioGasquez, line 1: Sorry to squeeze this unrelated commit, but this file was no longer correct

espflash/src/bin/espflash.rs
```diff
@@ -342,38 +342,21 @@ fn save_image(args: SaveImageArgs, config: &Config) -> Result<()> {
...
+
+    // If the file size is not divisible by 4, we need to pad `FF` bytes to the end
     let size = f.metadata().into_diagnostic()?.len();
+    let mut padded_bytes = 0;
+    if size % 4 != 0 {
+        padded_bytes = 4 - (size % 4);
+    }
     let mut buffer = Vec::with_capacity(size.try_into().into_diagnostic()?);
     f.read_to_end(&mut buffer).into_diagnostic()?;
+    buffer.extend_from_slice(&vec![0xFF; padded_bytes as usize]);
```
MabezDev, line 359: Biggest nitpick ever: we could avoid the vec allocation here, by using `buffer.extend` with `std::iter::repeat(0xFF).take(padded_bytes)`

MabezDev (approved):
LGTM, thanks for adding a HIL test!

PR #789, opened 2025-02-27
[cargo-espflash]: Add write-bin subcommand

SergioGasquez:
For some reason we were missing this command. Unless there is a reason for it, which I cant think off, we should have the same commands as in `espflash`

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -12,8 +12,8 @@
 
 use std::{
     collections::HashMap,
-    fs,
-    io::Write,
+    fs::{self, File},
+    io::{Read, Write},
     num::ParseIntError,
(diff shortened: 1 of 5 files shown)


espflash/src/cli/mod.rs
```diff
@@ -972,6 +986,31 @@ pub fn make_flash_data(
     )
 }
 
+pub fn write_bin(args: WriteBinArgs, config: &Config) -> Result<()> {
```
JurajSadel, line 989: We should add a docs here as well.

jessebraham: Why? The `cli` module will not be included in the documentation moving forward.

jessebraham (approved):
Thanks!

Issue #786, opened 2025-02-26
Update HIL and Release workflows to use newer version of ubuntu runner


jessebraham:
The `ubuntu-20.04` image will be EOL as of April 1st, and as such we need to update these workflows to use a newer version of the runner image.

Currently we experience GLIBC-related errors if we update to `ubuntu-22.04` or newer (AFAIK, due to the fact our self-hosted runners are running 20.04), so we need to either find a solution to this or update our runners to a newer version to resolve these.

jessebraham:
HIL has been updated, still working on the new release workflow.

PR #791, opened 2025-02-27
Update HIL workflow to use `ubuntu-22.04`, build in container to resolve GLIBC errors

jessebraham:
Part 1 of #786 

Still need to update the `release` workflow as well, but not entirely sure how to approach that yet. Can at least get HIL updated for now, though.


SergioGasquez (approved):
LGTM! Thanks

PR #795, opened 2025-02-28
Minor error handling and function naming improvements

jessebraham:
Nothing major here, just some simple cleanup:

- Improve error messages when receiving an incorrect response type, avoid some unwraps
- Remove `get_` prefix from `get_usb_pid`, remove useless `Result` from return type

Still a bit of untangling to do with regards to some imports from `flasher` in the `connection` module, but it's looking mostly okay I think. Not clear to me how to resolve those yet so will deal with it another time.


SergioGasquez (approved):
LGTM! Thanks for the cleanups!

Issue #792, opened 2025-02-27
ELF Module refactor


MabezDev:
We'll want to simplify this module.

We found that the `FirmwareImage` trait, is a bit of useless abstraction. _All_ firmware images start as ELFs, any binary (or hex etc.) data is written directly to flash without going through the image formatting process. Removing this, and the associated boiler plate should simplify this module.

PR #796, opened 2025-02-28
Refactor and simplify the `elf` module a bit

jessebraham:
- Eliminated the `ElfFirmwareImage` struct, as this was essentially just wrapping `xmas_elf::ElfFile`
- Merged `RomSegment` and `CodeSegment` structs into a new `Segment` struct, as they were identical
- Re-export a couple types from the `elf` module which are present in public APIs
- Removed unused `segments_with_load_addresses` function from `FirmwareImage` trait


SergioGasquez (approved):
LGTM! This is a nice step for #792, I also really like the renaming of some variables from `image` to `elf`, I had this pending.

Issue #799, opened 2025-03-04
cargo espflash fails when workspace Cargo.toml says edition=2024


acmorrow:
With a trivial `Cargo.toml` as generated by `cargo init`, the `cargo +esp espflash flash` command fails with an immediate error:

```
Error: cargo_espflash::invalid_workspace

  × The current workspace is invalid, and could not be loaded
  help: Ensure that a valid `Cargo.toml` file is in the executing directory
```

If `edition` is put back to `2021`, this error does not happen (though of course other errors do since this is not really a valid esp project).

jessebraham:
We've merged #800, which should resolve this error. I'd prefer not to backport this change to a patch release if possible, so if you don't mind installing `cargo-espflash` from the `main` branch here that should get you back on track.

If more people start reporting this and it becomes problematic I will reconsider the patch release, and will update here if we decide to do one.

PR #800, opened 2025-03-05
Update `cargo` dependency in `cargo-espflash`

jessebraham:
This resolves the error when using `edition = "2024"` in a package, see #799 for more details.


SergioGasquez (approved):
LGTM! Thanks

PR #801, opened 2025-03-05
Remove publish to crates.io job

### LINKED ISSUE #786: Update HIL and Release workflows to use newer version of ubuntu runner

SergioGasquez:


jessebraham (approved):
Thanks! 😅

Issue #745, opened 2025-02-05
Refactor image formatting, and allow supporting new image formats in a backward compatible way


MabezDev:
cc: #701 

This is quite a complex task, and we'll likely need to iterate a bit with the design here. Currently image formatting, is tightly coupled to the targets, which is then coupled to the flasher module. We need to find a way for elf inputs to be turned into the image format in a more self contained way.

In the meeting, we discussed some kind of API/Abstraction to take `CodeSegments` as an input, and then depending on the selected image format create an appropriate output. This was just a rough outline, and there is a lot of investigation and iteration needed to come up with a final solution here.

PR #802, opened 2025-03-05
Refactoring: remove the `FirmwareImage` trait and eliminate the `elf` module

### LINKED ISSUE #792: ELF Module refactor

jessebraham:
Changes should be pretty straightforward I hope. This is still just one of many refactoring PRs, so nothing is set in stone yet and there are still further changes to make.

One notable omission here is a generic interface for application image formats; I'm working on this but it's not ready yet, and I'd like to get these changes merged at least.

This also exposes `xmas_elf::ElfFile` in the public API, which is not a stable dependency, so this will need to be resolved. I will open an issue to keep track of this and resolve it in a later PR, probably when the aforementioned interface is created.

This closes #792, as the `elf` module no longer exists.

This is also the first steps towards #745, although there is still much work to be done regarding this issue.


SergioGasquez (approved):
LGTM! Thanks

Issue #787, opened 2025-02-26
`read-flash` doenst work with some lengths


SergioGasquez:
espflash `read-flash` times out with some lengths:

```
❯ espflash  read-flash 0 4 out.bin
[2025-02-26T12:19:03Z INFO ] Serial port: '/dev/ttyUSB0'
[2025-02-26T12:19:03Z INFO ] Connecting...
[2025-02-26T12:19:03Z INFO ] Using flash stub
[2025-02-26T12:19:04Z WARN ] Setting baud rate higher than 115,200 can cause issues
Chip type:         esp32c6 (revision v0.0)
Crystal frequency: 40 MHz
Flash size:        8MB
Features:          WiFi 6, BT 5
MAC address:       60:55:f9:f7:1b:6c
Error: espflash::timeout

  × Error while connecting to device
  ╰─▶ Timeout while running command
```
Not sure why, `11` works fine, but `12` fails. I tried a few different number (not with much detail), but looks like it only happens with low numbers

While `esptool`:

```
❯ esptools tool --chip esp32c6 read_flash  0 4 out.bin
Tool esptool mounted as `/home/sergio/.cache/esptools/78c195b08bd57bf27167b323600d417a546d3393/esptool`
Executing `esptool`
esptool.py v4.8.1
Found 33 serial ports
Serial port /dev/ttyUSB0
Connecting....
Chip is ESP32-C6 (QFN40) (revision v0.0)
Features: WiFi 6, BT 5, IEEE802.15.4
Crystal is 40MHz
MAC: 60:55:f9:00:00:f7:1b:6c
BASE MAC: 60:55:f9:f7:1b:6c
MAC_EXT: 00:00
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
4 (100 %)
4 (100 %)
Read 4 bytes at 0x00000000 in 0.0 seconds (3.1 kbit/s)...
Hard resetting via RTS pin...
Status Ok(ExitStatus(unix_wait_status(0)))
```

PR #804, opened 2025-03-05
Fix `read-flash` which didn't work with some lengths

playfulFence:
closes https://github.com/esp-rs/espflash/issues/787


jessebraham (approved):
Thanks, LGTM!

PR #805, opened 2025-03-06
Fix readmes links

SergioGasquez:
The third commit [39de5e1](https://github.com/esp-rs/espflash/pull/805/commits/39de5e1176faa019a6d8241494a0245ba77fedee) updates some docstrings, I can revert it but I think espflash does much more than just "flashing"


cargo-espflash/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name         = "cargo-espflash"
 version      = "4.0.0-dev"
 edition      = "2021"
 rust-version = "1.82"
-description  = "Cargo subcommand for flashing Espressif devices"
+description  = "Cargo subcommand for interacting Espressif devices"
```
jessebraham, line 6:
```suggestion
description  = "Cargo subcommand for interacting with Espressif devices"
```
😅

cargo-espflash/src/main.rs
```diff
@@ -88,7 +88,7 @@ enum Commands {
     Flash(FlashArgs),
     /// Hold the target device in reset
     HoldInReset(ConnectArgs),
-    /// List serial ports available for flashing.
+    /// List serial ports available.
```
jessebraham, line 91:
```suggestion
    /// List available serial ports.
```

If we're changing this anyway, this sounds more natural to me

jessebraham (approved):
Thanks!

PR #811, opened 2025-03-21
Remove unused dependencies, deduplicate some functions

jessebraham:
Nothing of interest here, just some cleanup:

- Removed a couple unused dependencies from `cargo-espflash`
- Noticed that both binaries contained two identical functions, so pulled them into the `cli` module


SergioGasquez (approved):
LGTM! Thanks

PR #816, opened 2025-03-25
Check chip using metadata

bugadani:
This PR reads chip name from the .elf and prevents flashing if it differs from the detected device.

### DIFF
--- a/espflash/src/cli/monitor/mod.rs
+++ b/espflash/src/cli/monitor/mod.rs
@@ -28,20 +28,18 @@ use strum::{Display, EnumIter, EnumString, VariantNames};
 
 use crate::{
     cli::{
-        monitor::{
-            parser::{InputParser, ResolvingPrinter},
-            symbols::Symbols,
-        },
+        monitor::parser::{InputParser, ResolvingPrinter},
         MonitorConfigArgs,
     },
     connection::{reset::reset_after_flash, Port},
+    image_format::Metadata,
 };
 
 pub mod external_processors;
 pub mod parser;
 
 mod line_endings;
-mod symbols;
+pub(crate) mod symbols;
 
 #[cfg_attr(feature = "cli", derive(clap::ValueEnum))]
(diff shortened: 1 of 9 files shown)


bugadani:
Only the last commit is new. I'm honestly not sure where exactly to add the checks. We could modify `connect`, pass the elf data to that and check where we check the `-c` argument, but it feels out of scope for `connect`. On the other hand, now we have to add a check to every place it matters, so the PR is messier than it should be. Ideas?

MabezDev:
I think I agree that it's a bit much for `connect`, but we already have `flasher.verify_minimum_revision()`, maybe a function on `Flasher` might make sense here. It might also avoid the `Option<&[u8]>` for elf data, as we can move that option check outside so that we only call `flasher.verify_elf_compat()` when there is elf data to verify against?

If you don't like that idea, I also think what you currently have isn't so bad, it's just three checks.

bugadani:
I've grouped the two bits of metadata into `Metadata`. Ideally, I think espflash should have a single Image struct that is parsed once, then used to extract metadata & symbols, but that's a bigger refactor than I'd like. Perf-wise we read all symbols twice because we extract Metadata twice, as opposed to looking up two symbols using linear search, so in the worst case we are in the same place.

SergioGasquez (approved):
LGTM! Thanks

espflash/src/cli/monitor/mod.rs
```diff
@@ -28,20 +28,18 @@ use strum::{Display, EnumIter, EnumString, VariantNames};
...
     connection::{reset::reset_after_flash, Port},
+    image_format::Metadata,
 };
 
 pub mod external_processors;
 pub mod parser;
 
 mod line_endings;
-mod symbols;
+pub(crate) mod symbols;
```
jessebraham, line 42: Sorry if I've missed something, but why does the visibility of this module need to change?

bugadani: It shouldn't I just forgot to undo it, thanks

jessebraham (approved):

PR #818, opened 2025-03-25
Option to supply custom defmt formats

### LINKED ISSUE #520: Add verbose defmt output like probe-rs
When using probe-rs with `defmt` We get timing and source info for each print. For example 
```
0.000061 INFO  Hello World!
└─ blinky::____embassy_main_task::{async_fn#0} @ src/bin/blinky.rs:14
```

For coloring, only the level word (INFO, DEBUG, etc.) is colored. I found it more readable than coloring the whole line.

It would be nice to be able to get the same with `espflash`.

I think it should not be the default but should be behind a cli flag (`-L defmt-verbose` for example).

### LINKED ISSUE #781: add defmt "--log-format" option
With probe-rs you can pass the argument `--log-format` as described in https://defmt.ferrous-systems.com/custom-log-output#passing-log-format-to-printers. Would it be possible to add this to espflash as well so we can edit the log output?

bugadani:
The formatting options are identical to probe-rs's, except in espflash we can't control timestamp/location with switches. The user can supply a custom format string if such control is necessary.

Closes #520 
Closes #781

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `watchdog-reset` strategy to `--after` subcommand (#779)
 - Add `ROM` version of `read-flash` command (#812)
 - `espflash` can detect the log format automatically from ESP-HAL metadata. Reqires `esp-println` 0.14 (presumably, yet to be released) (#809)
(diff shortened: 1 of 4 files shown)


SergioGasquez (approved):
LGTM! IIRC we have a c6 defmt test, we could test some output format(s) other than the default

jessebraham (approved):
Nice, thanks!

PR #819, opened 2025-03-25
Handle key events in non-interactive mode, too

### LINKED ISSUE #817: Unable to exit serial monitor with Ctrl-c when `--non-interactive` is passed
When using `--non-interactive` in conjunction with `--monitor`, it's impossible to exit the serial monitor without unplugging the device. My understanding was that the `--non-interactive` flag was meant to disable the forced user interaction with `dialoguer`, etc., and so this behaviour is somewhat of a surprise to me.

I think it would be best to allow Ctrl-c and Ctrl-r to be used even with `--non-interactive`, however the simple fix which allows this apparently causes issues with HIL testing. This will require some further discussion and investigation I suppose.

bugadani:
This PR implements two changes:

- The initial "Commands:" output is printed in non-interactive mode, too.
- `espflash` now ignores IO errors in non-interactive mode. This fixes #817 - when stdin is available, it will be read, otherwise stdin-related errors are ignores.


SergioGasquez (approved):
LGTM! Thanks

PR #820, opened 2025-03-27
Remove the dependency on `hex`

jessebraham:
- Removed the dependency on the `hex` package
- Renamed (de)serialization functions (just thought they could be a bit more clear, internal-only anyway)


SergioGasquez (approved):
LGTM! Thanks

Issue #803, opened 2025-03-05
Unstable `object::read::elf::ElfFile32` type is exposed in public API


jessebraham:
As of #821 this unstable type is being leaked, and this will need to be resolved before the next release.

PR #821, opened 2025-03-28
Remove the dependency on `xmas_elf`

jessebraham:
While these changes _technically_ have user-facing effects, we need to remove `ElfFile` from the public API anyway so I will not add a `CHANGELOG.md` entry here, since it will just be obsolete once #803 is resolved anyway.


SergioGasquez (approved):
LGTM!

PR #824, opened 2025-03-31
Remove `get_` prefixes from various functions

jessebraham:
This violates the Rust style guide, and is easy enough to fix while we're making breaking changes anyway.


SergioGasquez (approved):
LGTM! Thanks

PR #825, opened 2025-03-31
Don't leak `object` crate's `ElfFile32` type in public APIs

### LINKED ISSUE #803: Unstable `object::read::elf::ElfFile32` type is exposed in public API

jessebraham:
Closes #803

### DIFF
--- a/espflash/src/targets/esp32.rs
+++ b/espflash/src/targets/esp32.rs
@@ -1,7 +1,5 @@
 use std::ops::Range;
 
-use object::read::elf::ElfFile32 as ElfFile;
-
 #[cfg(feature = "serialport")]
 use crate::{connection::Connection, targets::bytes_to_mac_addr};
 use crate::{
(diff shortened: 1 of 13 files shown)


JurajSadel (approved):
Thanks!

espflash/src/targets/esp32.rs
```diff
@@ -157,12 +155,12 @@ impl Target for Esp32 {
 
     fn flash_image<'a>(
         &self,
-        elf: ElfFile<'a>,
+        elf_data: &'a [u8],
         flash_data: FlashData,
         _chip_revision: Option<(u32, u32)>,
         xtal_freq: XtalFrequency,
     ) -> Result<IdfBootloaderFormat<'a>, Error> {
-        let booloader: &'static [u8] = match xtal_freq {
```
SergioGasquez, line 165: https://github.com/esp-rs/esp-hal/pull/3281#discussion_r2018904119 funny that we found the same typo in two different projects in the same day

SergioGasquez (approved):
Thanks!

PR #828, opened 2025-04-03
Allow flashing binary to a named partition

### LINKED ISSUE #631: Allow a partition as a target for `write-bin`
Currently, `write-bin` requires the address for the data to be written. It would be nice if one could specify a partition instead. This would also allow a length check, so that partition boundaries are not violated.

bugadani:
Closes #631

This PR allows specifying a partition label for `write-bin`, in addition to a numeric address. If a partition label is passed, espflash tries to parse the provided/configured partition table, and resolve the label. This information is then used to a) check the size of the flashed binary and b) specify the address where the binary should be written.

As an alternative approach, we might read back the partition table from the device itself. Since users have the option to configure a partition table in a config file, this option would need a `--read-idf-partition-table-from-device` or similar flag, meaning it can still be implemented without breaking the command's current behaviour.


SergioGasquez (approved):
LGTM! Could we add a HIL test for this?

bugadani:
whoops that's a different pr 🙈

SergioGasquez (approved):

PR #829, opened 2025-04-04
Improve reading efuses

SergioGasquez:
Trying to improve the way we handle eFuses, any input is welcome

No public changes, so `skip-changelog`


MabezDev:
Can we leverage the CSV files that we use in esp-hal for efuse access here? I'd really like to avoid the manual impls if possible, because it will be a chore for new chip bring up.

For example, all chips appear to have a "MAC": https://docs.espressif.com/projects/rust/esp-hal/1.0.0-beta.0/esp32/esp_hal/efuse/constant.MAC.html constant in the CSV's so we can leverage that to only have to process this field in one way.

jessebraham:
As we've discussed privately, I've been working on an alternate approach for this. While not quite complete yet, I'm far enough along to be confident that it will work. So, thanks for the PR, but will close this in favour of my upcoming PR.

PR #830, opened 2025-04-04
Bail from flashing a partition table to too-small flash

bugadani:


jessebraham (approved):
Thanks! (Error just seems to be feature-gate related, otherwise looks good)

Issue #726, opened 2025-01-19
`espflash write-bin` not compatible with Secure Download mode


ivmarkov:
Once [Secure Download mode](https://docs.espressif.com/projects/esp-idf/en/v5.4/esp32s3/security/flash-encryption.html#best-practices) is enabled on the chip, no matter what I do I can't force the chip into writing a binary image with `espflash`.

Some commands I tried:
```sh
espflash write-bin -b no-reset-no-sync --chip esp32s3 0x10000 factory.bin
espflash write-bin -b no-reset-no-sync --chip esp32s3 0x10000 factory.bin
espflash write-bin -b no-sync --chip esp32s3 0x10000 factory.bin
espflash write-bin -b no-reset --chip esp32s3 0x10000 factory.bin
espflash write-bin -b usb-reset --chip esp32s3 0x10000 factory.bin
```

These fail either with:
```sh
Error: espflash::connection_failed

  × Error while connecting to device
  ╰─▶ Failed to connect to the device
  help: Ensure that the device is connected and the reset and boot pins are not being held down
```

or with:
```sh
Error: espflash::timeout

  × Error while connecting to device
  ╰─▶ Timeout while running ReadReg command
```

This ^^^ failure mode is interesting, because it seems `espflash` is capable of setting the chip in download mode after all (with some reset sequences), but then (I'm speculating) it tries to read (some?) register, which might not be supported in secure DL mode.

=========

### `esptool`

In contrast, `esptool` (a) is capable to detect that the chip is in Security Download mode (try `esptools tool get_security_info`) and (b) capable of writing to the flash (albeit with the default baud rate):
```sh
esptools tool write_flash 0x10000 factory.bin --force
```

ivmarkov:
For completeness, just to mention that I managed to convince the monitor to show logs when the chip is in secure DL mode by using:
```sh
espflash monitor -b no-reset-no-sync --chip esp32s3
```

chris-subtlebytes:
I hit this a while ago too and I think I just switched to esptool for the one device I accidentally enabled secure download mode on.

For my future development, I added this to my `sdkconfig.defaults` to prevent secure DL mode from enabling:

```
CONFIG_SECURE_BOOT_INSECURE=y # required for CONFIG_SECURE_INSECURE_ALLOW_DL_MODE to actually enable
CONFIG_SECURE_INSECURE_ALLOW_DL_MODE=y
[...]

PR #832, opened 2025-04-04
Add (some level of) SDM support

playfulFence:
closes https://github.com/esp-rs/espflash/issues/726

Initial issue contains just a little bit of lie: espflash didn't work with SDM almost at all 😅 
After this is merged, at least board-info and write-bin (mentioned in the original issue) functions will be supported.

No extra flags are needed, user will be able to just execute:
`espflash write-bin 0x0 --monitor <bin-name>`

I recommend setting a higher baudrate as otherwise flashing in SDM takes a lot of time.
Also, `--no-stub` is applied automatically when tool detects that connected chip is in SDM mode.

### DIFF
--- a/espflash/src/connection/mod.rs
+++ b/espflash/src/connection/mod.rs
@@ -523,12 +523,15 @@ impl Connection {
     ///  Write a command and reads the response
     pub fn command(&mut self, command: Command<'_>) -> Result<CommandResponseValue, Error> {
         let ty = command.command_type();
+        info!("deb: Why");
         self.write_command(command).for_command(ty)?;
-
+        info!("deb: is");
         for _ in 0..100 {
             match self.read_response().for_command(ty)? {
                 Some(response) if response.return_op == ty as u8 => {
+                    info!("deb: nefailing (return_op {} and ty {})\nwhole resp: {:?})", response.return_op, ty, response);
                     return if response.error != 0 {
+                        info!("deb: resp error {}", response.error);
                         let _error = self.flush();
                         Err(Error::RomError(RomError::new(
                             command.command_type(),
@@ -537,6 +540,7 @@ impl Connection {
                     } else {
                         // Check if the response is a Vector and strip header (first 8 bytes)
                         // https://github.com/espressif/esptool/blob/749d1ad/esptool/loader.py#L481
+                        info!("deb: fuck");
                         let modified_value = match response.value {
                             CommandResponseValue::Vector(mut vec) if vec.len() >= 8 => {
                                 vec = vec[8..][..response.return_length as usize].to_vec();
--- a/espflash/src/connection/mod.rs
+++ b/espflash/src/connection/mod.rs
@@ -449,11 +449,17 @@ impl Connection {
                 // - https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/serial-protocol.html?highlight=md5#response-packet
                 // - https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/serial-protocol.html?highlight=md5#status-bytes
                 // - https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/serial-protocol.html?highlight=md5#verifying-uploaded-data
+
+
+                // TODO: status len should be 2 for SDM
+
                 let status_len = if response.len() == 10 || response.len() == 26 {
                     2
                 } else {
                     4
                 };
+
+                let status_len = 2;
 
                 let value = match response.len() {
                     10 | 12 => CommandResponseValue::ValueU32(u32::from_le_bytes(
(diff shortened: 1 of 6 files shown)


espflash/src/connection/mod.rs
```diff
@@ -449,11 +451,13 @@ impl Connection {
                 // - https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/serial-protocol.html?highlight=md5#response-packet
                 // - https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/serial-protocol.html?highlight=md5#status-bytes
                 // - https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/serial-protocol.html?highlight=md5#verifying-uploaded-data
-                let status_len = if response.len() == 10 || response.len() == 26 {
-                    2
-                } else {
-                    4
-                };
+
+                let status_len =
```
MabezDev, line 455: This could benefit from some comments, and perhaps a permalink to either the esptool impl, or the docs explaining this return data from the ROM bootloader.

espflash/src/flasher/mod.rs
```diff
@@ -631,7 +631,7 @@ pub struct DeviceInfo {
     /// Device features
     pub features: Vec<String>,
     /// MAC address
-    pub mac_address: String,
+    pub mac_address: Option<String>,
```
MabezDev, line 634: The ROM bootloader doesn't send MAC in SDM?

espflash/src/flasher/mod.rs
```diff
@@ -1382,3 +1409,14 @@ fn detect_chip(connection: &mut Connection, use_stub: bool) -> Result<Chip, Erro
         }
     }
 }
+
+fn detect_sdm(connection: &mut Connection) -> Result<bool, Error> {
```
MabezDev, line 1413: Can we detect sd mode in the normal detection routine, instead of doing it twice?

espflash/src/flasher/mod.rs
```diff
@@ -709,11 +711,18 @@ impl Flasher {
 
         // Load flash stub if enabled
         if use_stub {
-            info!("Using flash stub");
-            flasher.load_stub()?;
+            if !sdm {
```
MabezDev, line 714: If you bring the `!sdm` if outside, you can avoid two seperate if's here and below

espflash/src/flasher/mod.rs
MabezDev: I.e, inside `connection.begin()`, can we return the download mode type there?

espflash/src/targets/flash_target/ram.rs
```diff
@@ -36,6 +36,15 @@ impl FlashTarget for RamTarget {
         Ok(())
     }
 
+    fn write_segment_sdm(
+        &mut self,
+        _connection: &mut Connection,
+        _segment: Segment<'_>,
+        _progress: &mut Option<&mut dyn ProgressCallbacks>,
+    ) -> Result<(), Error> {
+        todo!()
```
MabezDev, line 45: If we can't implement this, this should return a nice error, not panic.

MabezDev: If it can be implemented at some point, please also add a TODO with a relevant issue.

espflash/src/flasher/mod.rs
```diff
@@ -982,14 +991,25 @@ impl Flasher {
         let chip = self.chip();
         let target = chip.into_target();
 
-        let revision = Some(target.chip_revision(self.connection())?);
+        // chip_revision reads from efuse, which is not possible in Secure Download Mode
+        let revision = if !self.connection.secure_download_mode {
```
MabezDev, line 995: We can use https://doc.rust-lang.org/std/primitive.bool.html#method.then_some to remove the need for the if statements.

MabezDev (changes requested):
Thanks, I haven't tested it, but I'm sure it works :D. I've added some comments which I think will improve the PR.

jessebraham (commented):
I agree with all of the other review comments, plus a couple more to add.

playfulFence:
Thank you @MabezDev and @jessebraham for the reviews though, will address them when I'm back to this task.

espflash/src/flasher/mod.rs
playfulFence: Yeah, the MAC is read from efuse, but in SDM you can't read efuses, so that's the reason. I should probably add a comment to clarify this though, thanks!

espflash/src/flasher/mod.rs
playfulFence: @MabezDev
Not "then_some", as it eagerly evaluates arguments inside of it, which means, that even if chip IS in Secure Download Mode, "read_reg" inside of that "chip_revision" function will be evaluated -> everything will fail. 

Instead, me might want to use something like

```
let revision = (!self.connection.secure_download_mode)
            .then(|| target.chip_revision(self.connection()))
            .transpose()?;
```

PR #838, opened 2025-04-09
Add `board-info` command support in Secure Download Mode

playfulFence:
Some progress towards https://github.com/esp-rs/espflash/issues/726

Initial issue contains just a little bit of lie: espflash didn't work with SDM almost at all 😅
After this is merged, at least board-info command will be supported. `write-bin` support will be added later in https://github.com/esp-rs/espflash/pull/832

No extra flags are needed, user will be able to just execute:
```bash
espflash board-info
```

Also, --no-stub is applied automatically when tool detects that connected chip is in SDM mode.

### DIFF
--- a/espflash/src/connection/mod.rs
+++ b/espflash/src/connection/mod.rs
@@ -124,6 +124,7 @@ pub struct Connection {
     decoder: SlipDecoder,
     after_operation: ResetAfterOperation,
     before_operation: ResetBeforeOperation,
+    pub(crate) secure_download_mode: bool,
 }
 
 impl Connection {
@@ -139,11 +140,12 @@ impl Connection {
             decoder: SlipDecoder::new(),
             after_operation,
             before_operation,
+            secure_download_mode: false,
         }
     }
 
     /// Initialize a connection with a device
-    pub fn begin(&mut self) -> Result<(), Error> {
+    pub fn begin(&mut self) -> Result<bool, Error> {
         let port_name = self.serial.name().unwrap_or_default();
         let reset_sequence = construct_reset_strategy_sequence(
             &port_name,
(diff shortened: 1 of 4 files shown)


espflash/src/connection/mod.rs
```diff
@@ -483,8 +493,8 @@ impl Connection {
                     return_op: response[1],
                     return_length: u16::from_le_bytes(response[2..][..2].try_into().unwrap()),
                     value,
-                    error: response[response.len() - status_len],
-                    status: response[response.len() - status_len + 1],
+                    error: response[response.len() - status_len + 1],
+                    status: response[response.len() - status_len],
```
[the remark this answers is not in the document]
playfulFence: https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/serial-protocol.html#status-bytes

espflash/src/connection/mod.rs
```diff
@@ -139,11 +140,12 @@ impl Connection {
             decoder: SlipDecoder::new(),
             after_operation,
             before_operation,
+            secure_download_mode: false,
         }
     }
 
     /// Initialize a connection with a device
-    pub fn begin(&mut self) -> Result<(), Error> {
+    pub fn begin(&mut self) -> Result<bool, Error> {
```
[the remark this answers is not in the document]
playfulFence: Fixed

espflash/src/flasher/mod.rs
```diff
@@ -710,14 +712,19 @@ impl Flasher {
             return Ok(flasher);
         }
 
-        // Load flash stub if enabled
-        if use_stub {
-            info!("Using flash stub");
-            flasher.load_stub()?;
+        if !sdm {
```
jessebraham, line 715:
```suggestion
        if !connection.secure_download_mode {
```
Sorry to nitpick, but I don't think we need to create a local variable which is only referenced in one place.

espflash/src/flasher/mod.rs
```diff
@@ -1102,10 +1116,20 @@ impl Flasher {
         let mut target = self
             .chip
             .flash_target(self.spi_params, self.use_stub, false, false);
+
         target.begin(&mut self.connection).flashing()?;
+
         for segment in segments {
-            target.write_segment(&mut self.connection, segment.borrow(), &mut progress)?;
+            if self.connection.secure_download_mode {
```
jessebraham, line 1123: Why is this check being performed inside the loop? This should be the first item in the function body, there's no point running any of the code in this function if we're just going to error out anyway.

espflash/src/flasher/mod.rs
playfulFence: I was about to say that "connection" is moved to "flasher", but then realized that I can use "flasher.connection" actually 🤡 

Will be done, thanks haha 😄

espflash/src/flasher/mod.rs
playfulFence: Good catch, thanks! I didn't change it after separation from the original bigger PR.

SergioGasquez (approved):
LGTM!

jessebraham (approved):
Thanks!

PR #841, opened 2025-04-10
Remove the dependency on `cargo` from the `cargo-espflash` package

jessebraham:
Still needs more testing, but this PR removes the dependency on `cargo` from the `cargo-espflash` package. This dependency has caused issues a number of times, and it quite enormous, so glad to see it gone.

Based on some very limited testing this is still working the same as it did prior to the changes, but I would like to try it out in some more interesting situations to be sure I haven't subtly broken anything.

There are no user-facing changes here, so skipping the `CHANGELOG.md` entry.


MabezDev (approved):
+1 from me, but I'll let you finish your further tests

jessebraham:
I guess this is fine to merge, if there are any issues we can deal with them.

PR #846, opened 2025-04-17
Fix CI error by installing newly required dependency

jessebraham:
Not sure why this is suddenly required (CI worked fine yesterday), but easy fix at least 🤷🏻‍♂️


jessebraham:
Seems to be alternating passing/failing with each run now, fun 🙃

SergioGasquez (approved):
LGTM!

PR #847, opened 2025-04-17
Implement new API for reading eFuse values

### LINKED ISSUE #827: Current API for reading eFuse values is confusing/difficult to use
I've been meaning to open an issue for this for months now, but keep forgetting 😅 

The current API for reading eFuse values is quite confusing, and requires doing math in order to port features over from `esptool.py`. We should try to better align the APIs to make it easier to adapt functionality and remove some potential sources of errors.

This functionality, IIRC, is purely internal, so we should be able to change it whenever without breaking the public API. However, I will add it to the v4 milestone regardless, as I would like to get it done :)

jessebraham:
This turned out to be a large, so apologies for that. Plenty of the code can be disregarded in review though (i.e. the generated eFuse constants).

This PR introduces a new `xtask` package to the workspace. While I've been resisting adding this for some time, it's now become a bit of a necessity. This is used to generate the eFuse field definitions from the YAML files defined in the `esptool` repository.

Additionally, all code relating to reading of eFuse values has been completely rewritten. There are a few hacky bits here and there, mostly relating to the ESP32, but overall I think things are in much better shape. Perhaps we can iterate on this further and improve those areas, however for the time being I'm not too worried about it.

I have tested on all supported devices to confirm that things are working as expected; as far as I can tell, we're in good shape. There is one caveat, on the ESP32-S2 for whatever reason we are not reading the correct BLOCK2 version number; this does not really make sense to me, and honestly I'm getting kind of tired of thinking about eFuses at this point, so happy to just open an issue for this and worry about it later.

Closes #827

### DIFF
--- a/espflash/src/targets/esp32s2.rs
+++ b/espflash/src/targets/esp32s2.rs
@@ -2,11 +2,19 @@ use std::ops::Range;
 
 #[cfg(feature = "serialport")]
 use super::flash_target::MAX_RAM_BLOCK_SIZE;
+use super::{
+    Chip,
+    Esp32Params,
+    ReadEFuse,
+    SpiRegisters,
+    Target,
+    XtalFrequency,
+    efuse::esp32s2 as efuse,
+};
 use crate::{
     Error,
     flasher::{FlashData, FlashFrequency},
     image_format::IdfBootloaderFormat,
-    targets::{Chip, Esp32Params, ReadEFuse, SpiRegisters, Target, XtalFrequency},
 };
 #[cfg(feature = "serialport")]
 use crate::{connection::Connection, flasher::FLASH_WRITE_SIZE};
@@ -40,28 +48,20 @@ impl Esp32s2 {
     /// Return the block2 version based on eFuses
     #[cfg(feature = "serialport")]
     fn block2_version(&self, connection: &mut Connection) -> Result<u32, Error> {
-        let blk2_word4 = self.read_efuse(connection, 27)?;
-        let block2_version = (blk2_word4 >> 4) & 0x7;
-
-        Ok(block2_version)
+        // NOTE: Perplexingly, this does not seem to work correctly?
+        self.read_efuse(connection, efuse::BLK_VERSION_MINOR)
     }
 
     /// Return the flash version based on eFuses
     #[cfg(feature = "serialport")]
     fn flash_version(&self, connection: &mut Connection) -> Result<u32, Error> {
-        let blk1_word3 = self.read_efuse(connection, 20)?;
-        let flash_version = (blk1_word3 >> 21) & 0xf;
-
-        Ok(flash_version)
+        self.read_efuse(connection, efuse::FLASH_VERSION)
     }
 
     /// Return the PSRAM version based on eFuses
     #[cfg(feature = "serialport")]
     fn psram_version(&self, connection: &mut Connection) -> Result<u32, Error> {
-        let blk1_word3 = self.read_efuse(connection, 20)?;
-        let psram_version = (blk1_word3 >> 28) & 0xf;
-
-        Ok(psram_version)
+        self.read_efuse(connection, efuse::PSRAM_VERSION)
     }
 
     /// Check if the magic value contains the specified value
@@ -72,11 +72,23 @@ impl Esp32s2 {
 
 impl ReadEFuse for Esp32s2 {
     fn efuse_reg(&self) -> u32 {
-        0x3f41_a000
+        0x3F41_A000
+    }
+
+    fn block0_offset(&self) -> u32 {
+        0x2C
+    }
+
+    fn block_size(&self, block: usize) -> u32 {
+        efuse::BLOCK_SIZES[block]
     }
 }
 
 impl Target for Esp32s2 {
+    fn chip(&self) -> Chip {
+        Chip::Esp32s2
+    }
+
     fn addr_is_flash(&self, addr: u32) -> bool {
         FLASH_RANGES.iter().any(|range| range.contains(&addr))
     }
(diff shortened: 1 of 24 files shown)


espflash/src/targets/mod.rs
```diff
@@ -367,14 +412,31 @@ pub trait Target: ReadEFuse {
...
+        let (mac0_field, mac1_field) = match self.chip() {
+            Chip::Esp32 => (self::efuse::esp32::MAC0, self::efuse::esp32::MAC1),
+            Chip::Esp32c2 => (self::efuse::esp32c2::MAC0, self::efuse::esp32c2::MAC1),
+            Chip::Esp32c3 => (self::efuse::esp32c3::MAC0, self::efuse::esp32c3::MAC1),
+            Chip::Esp32c6 => (self::efuse::esp32c6::MAC0, self::efuse::esp32c6::MAC1),
+            Chip::Esp32h2 => (self::efuse::esp32h2::MAC0, self::efuse::esp32h2::MAC1),
+            Chip::Esp32p4 => (self::efuse::esp32p4::MAC0, self::efuse::esp32p4::MAC1),
+            Chip::Esp32s2 => (self::efuse::esp32s2::MAC0, self::efuse::esp32s2::MAC1),
+            Chip::Esp32s3 => (self::efuse::esp32s3::MAC0, self::efuse::esp32s3::MAC1),
+        };
```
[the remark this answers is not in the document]
SergioGasquez: I cant think of a better way neither

jessebraham:
@SergioGasquez pointed out that there may be a few more opportunities to use the field constants rather than `read_efuse_raw` for the ESP32, so will look into that shortly.

xtask/src/main.rs
```diff
@@ -0,0 +1,284 @@
...
+        .current_dir(workspace)
+        .output()?;
+
+    Ok(())
+}
+
+fn parse_efuse_fields(efuse_yaml_path: &Path) -> Result<EfuseFields> {
+    // TODO: We can probably handle this better, e.g. by defining a `Chip` enum
+    //       which can be iterated over, but for now this is good enough.
+    const CHIPS: &[&str] = &[
```
MabezDev, line 130: You can depend on esp-metadata and get this for free

xtask/src/main.rs
```diff
@@ -0,0 +1,284 @@
...
+        mac1_attrs.len = 16;
+
+        yaml.fields.remove("MAC").unwrap();
+        yaml.fields.insert("MAC0".into(), mac0_attrs);
+        yaml.fields.insert("MAC1".into(), mac1_attrs);
+    }
+
+    // The ESP32-S2 seems to be missing a reserved byte at the end of BLOCK0
+    // (Or, something else weird is going on).
+    for (chip, yaml) in efuse_fields {
```
MabezDev, line 178: I don't think this needs to be a loop, I think you can just grab the esp32s2 fields from the `EfuseFields` hashmap and modify it once?

MabezDev (commented):
This is looking great, way more grok-able than the random bit shifts :D.

I have a couple of initial comments to address. I'm also aware that we also have some efuse field generation in esp-hal's xtask, is the plan to retire that and keep field generation here and periodically copy the (presumably the .rs) files over to esp-hal?

espflash/src/targets/mod.rs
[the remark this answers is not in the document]
MabezDev: The only way I could think of of removing this would be to add more trait methods to `ReadEFuse` to get `mac0` and `mac1`. 

I think this is better though, because all that does it push this concise match into separate files, and I can't see that being a good thing.

xtask/src/main.rs
jessebraham: Yeah was just copy-pasting and being lazy

xtask/src/main.rs
jessebraham: `esp-metadata` doesn't have the P4 (solvable of course), and pulls in some additional dependencies, so I just didn't bother. I think given that we probably won't need to run the subcommand until we add support for a new chip, which who knows when that will be, it's okay as-is for the time being.

jessebraham:
Probably I should update the CI workflow to ensure the `xtask` package builds, too. Forgot about that.

SergioGasquez (approved):
LGTM! Thanks for working on this, it's definitely way better now!

jessebraham:
> > I'm also aware that we also have some efuse field generation in esp-hal's xtask, is the plan to retire that and keep field generation here and periodically copy the (presumably the .rs) files over to esp-hal?
> 
> I don't know what the plan is, meant to bring this up in the original comment. Probably it's not a bad idea to even have a separate `esp-efuse` package or something, but as we were discussing yesterday I'm hesitant to keep creating packages until we have improved our release workflow.
> 
> Open to suggestions here, was just going with the path of least resistance to get something working.

Regarding this, probably our options are:

1. Keep the eFuse field generation code in one repo only, either here or in `esp-hal`, and just copy-paste the files between repositories as needed
2. Create a new `esp-efuse` package which includes this functionality and provides the field definitions for all target devices
3. Duplicate the code between repositories and not worry about it too much

Probably 1 or 2 are the preferable choices, don't really care much which we go with. I also think this can be decided on at a later date and does not need to block this PR.

SergioGasquez:
> 1. Keep the eFuse field generation code in one repo only, either here or in `esp-hal`, and just copy-paste the files between repositories as needed
> 2. Create a new `esp-efuse` package which includes this functionality and provides the field definitions for all target devices
> 3. Duplicate the code between repositories and not worry about it too much

I would also lean towards 1 or 2, maybe I like having a standalone package a bit more, it shouldn't be too difficult to maintain and release.

espflash/src/targets/esp32s2.rs
```diff
@@ -72,11 +72,23 @@ impl Esp32s2 {
 
 impl ReadEFuse for Esp32s2 {
     fn efuse_reg(&self) -> u32 {
-        0x3f41_a000
+        0x3F41_A000
```
bugadani, line 75: What's up with the address? The TRM has 0x3ffc1a000
image

jessebraham: These addresses are taken from `esptool`, no idea why there's no consistency with this stuff.

bugadani: Apologies, I was looking at TRM v1.1, which has since been corrected.

espflash/src/targets/mod.rs
```diff
@@ -304,16 +308,57 @@ pub trait ReadEFuse {
...
+        connection: &mut Connection,
+        block: u32,
+        word: u32,
+    ) -> Result<u32, Error> {
+        let block0_addr = self.efuse_reg() + self.block0_offset();
+        let block_offset = if block == 0 {
+            0
+        } else {
+            self.block_size(block as usize - 1)
+        };
```
bugadani, line 349: This isn't correct. We shouldn't offset by the size of the previous block, we should offset by the total size of all previous blocks. Something like this works:

```rust
        let mut block_offset = 0;
        for b in 0..block {
            block_offset += self.block_size(b as usize);
        }
```

```
Chip type:         esp32s2 (revision v1.0)
Crystal frequency: 40 MHz
Flash size:        8MB
Features:          WiFi, No Embedded Flash, Embedded PSRAM 2MB, ADC and temperature sensor calibration in BLK2 of efuse V2
```

SergioGasquez (approved):

MabezDev:
1 or 2 seems best here, probably 1 initially and we can push it into it's own crate after.

PR #852, opened 2025-04-22
Correct bit start offset for `MAC1` field

jessebraham:
Oops 😅 

(Because of how we're reading the MAC address, this thankfully didn't affect `espflash`, however when porting this functionality over to `esp-hal` I noticed this change was required, so may as well be correct here anyway)


SergioGasquez (approved):
LGTM!

Issue #657, opened 2024-07-28
QIO does not work (at least with esp-s3-wroom-1-N16R8)


liebman:
If you specify the flash mode or `qio` to espflash then it fails to boot:

```
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x1 (POWERON),boot:0x2b (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:QIO, clock div:1
load:0x3fce3818,len:0x16f8
ets_loader.c 78 
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x7 (TG0WDT_SYS_RST),boot:0x2b (SPI_FAST_FLASH_BOOT)
Saved PC:0x400454d5
SPIWP:0xee
mode:QIO, clock div:1
load:0x3fce3818,len:0x16f8
ets_loader.c 78 
```

However If I build and copy a `bootloader.bin` from an `esp-idf` project that specifies the QIO flash mode and just let `espflash` flash as DIO it works. And the esp-idf bootloader switches to QIO.

[This section](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-guides/bootloader.html#spi-flash-configuration) in the api guide includes:

> The First Stage Bootloader in ROM reads the Second Stage Bootloader header information from flash and uses this information to load the rest of the Second Stage Bootloader from flash. However, at this time the system clock speed is lower than configured and not all flash modes are supported.

Also esp-idf seems to mark the bootloader as DIO when you have specified QIO.

liebman:
More info found [here](https://docs.espressif.com/projects/esptool/en/latest/esp32s3/advanced-topics/spi-flash-modes.html#how-is-flash-mode-communicated-to-the-espressif-chip) that confirms that the bootloader should be marked as DIO even for QIO.

USTHzhanglu:
i think is not espflash question
with idf, through config FLASHMODE_QIO, sdkconfig config CONFIG_ESPTOOLPY_FLASHMODE="dio" and couldn't change.
by cmd
```sh
esptool.py --chip esp32s3 merge_bin -o hello_world.bin --flash_mode qio --flash_freq 80m --flash_size 16MB 0x0 bootloader/bootloader.bin 0x10000 hello.bin 0x8000 partition_table/partition-table.bin
```
and flash,it also error
```log
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
[...]

PR #857, opened 2025-04-24
Update bootloaders with IDF release/v5.4 ones

JurajSadel:
closes https://github.com/esp-rs/espflash/issues/853


JurajSadel:
Some help with testing will be appreciated here 😅

bjoernQ:
There is one catch: newer bootloaders _need_ the app-descriptor which we support via the unreleased esp-bootloader-esp-idf - users will need to use that to have a app-descriptor

I guess updating to the latest bootloader which doesn't require the app-descriptor would be the best option for now

MabezDev:
> There is one catch: newer bootloaders need the app-descriptor which we support via the unreleased esp-bootloader-esp-idf - users will need to use that to have a app-descriptor
>
> I guess updating to the latest bootloader which doesn't require the app-descriptor would be the best option for now

but all v5.x versions need the app image desc, and no 4.x boot loader is in "maintenance", though the maintenance is probably not an issue.

I think we should switch now, and start tackling this breakage. We'll have various tests/CI workflows etc that will need to be updated. From that we can find edge cases and troubleshooting steps ready for a migration guide entry in esp-hal later (doesn't strictly belong there, but gets the most eyeballs). If we wait, we only delay this process until the espflash release.

bjoernQ:
> but all v5.x versions need the app image desc

I _think_ some older 5.x bootloaders don't strictly require it (at least w/o OTA support). Let me have a look

bjoernQ:
In any case we should document which exact version the bootloaders are from (ideally including the exact commit) and what config options were used (or maybe we should also define them here - see https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/kconfig.html#bootloader-config)

elipsitz:
This could also be a good opportunity to fix the issue with the esp32-s3 bootloader (#657)

bjoernQ:
> "I (76) esp_image: MMU page size mismatch, configured: 0x8000, found: 0x10000"

This is not great given we take all the effort to generate new bootloaders

esp-hal (for now) uses 0x10000 fixed/hard-coded (applies to C2, C6, H2 IIRC)

espflash/resources/bootloaders/README.md
```diff
@@ -0,0 +1 @@
+The listed bootloaders were built with [esp-idf/release/v5.4](https://github.com/espressif/esp-idf/tree/release/v5.4) at commit `3ad3632`, using the default settings.
```
bjoernQ, line 1: Maybe pasting the output of `grep BOOTLOADER sdkconfig` would be even better

JurajSadel: Good idea

jessebraham:
> > "I (76) esp_image: MMU page size mismatch, configured: 0x8000, found: 0x10000"
> 
> This is not great given we take all the effort to generate new bootloaders
> 
> esp-hal (for now) uses 0x10000 fixed/hard-coded (applies to C2, C6, H2 IIRC)

Probably we should build the bootloaders with the same values being used by the HAL, in this case.

JurajSadel:
> > > "I (76) esp_image: MMU page size mismatch, configured: 0x8000, found: 0x10000"
> > 
> > 
> > This is not great given we take all the effort to generate new bootloaders
> > esp-hal (for now) uses 0x10000 fixed/hard-coded (applies to C2, C6, H2 IIRC)
> 
> Probably we should build the bootloaders with the same values being used by the HAL, in this case.

I rebuilt C2, C6, and H2 with `0x10000` so the warning should be gone

MabezDev:
> This is not correct. The published version of espflash is using the bootloader from v5.1 and does not require it. IIRC v5.2 also does not require it. The dependency on the application descriptor was added with either v5.3 or v5.4 (I don't recall which).

Thanks for clearing that up, I thought it was all 5.x versions :sweat_smile:. We could only update to v5.2 here, but I think we should just rip this band aid off as this app desc thing will be a requirement going forward if we want to update the bootloaders again in the future.

espflash/resources/bootloaders/README.md
```diff
@@ -0,0 +1,42 @@
+The listed bootloaders were built with [esp-idf/release/v5.4](https://github.com/espressif/esp-idf/tree/release/v5.4) at commit `3ad3632`, using default settings:
```
jessebraham, line 1: Are these default settings the same for all chips?

jessebraham (approved):
Thanks, LGTM!

PR #863, opened 2025-05-22
Add support for the ESP32-C5

### LINKED ISSUE #861: Add support for the ESP32-C5

jessebraham:
Closes #861


espflash/src/targets/esp32c5.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    fn flash_image<'a>(
+        &self,
+        elf_data: &'a [u8],
+        flash_data: FlashData,
+        _chip_revision: Option<(u32, u32)>,
+        xtal_freq: XtalFrequency,
+    ) -> Result<IdfBootloaderFormat<'a>, Error> {
+        if xtal_freq != XtalFrequency::_40Mhz {
```
jessebraham, line 120: Oops, forgot to update this. Will fix shortly.

espflash/src/targets/esp32c5.rs
```diff
@@ -0,0 +1,145 @@
...
+use crate::connection::Connection;
+use crate::{
+    Error,
+    flasher::{FlashData, FlashFrequency},
+    image_format::IdfBootloaderFormat,
+};
+
+pub(crate) const CHIP_ID: u16 = 23;
+
+const CHIP_DETECT_MAGIC_VALUES: &[u32] = &[];
```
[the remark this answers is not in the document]
jessebraham: The P4 does as well, other chips support using security info as well (and in fact they've moved away from using magic values at all wherever possible)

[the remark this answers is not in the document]
SergioGasquez: I guess it only uses security info for chip detection

espflash/src/targets/esp32c5.rs
jessebraham: Ahh I guess we probably need two bootloaders as well, since it supports 40MHz and 48MHz xtals? 🙃

espflash/resources/README.md
```diff
@@ -1,7 +1,9 @@
 # Espflash Resources
 
-The bootloaders have been built by hand from ESP-IDF, using the `release/v5.1` branch:
-https://github.com/espressif/esp-idf/tree/release/v5.1
+The listed bootloaders from `espressif/esp-idf` were built with `release/v5.4` at commit `3ad3632`, using default settings:
+https://github.com/espressif/esp-idf/tree/release/v5.4
 
-The ESP32s flasher stubs are taken from the **esp-rs/esp-flasher-stub** repository:
-https://github.com/esp-rs/esp-flasher-stub/releases/latest
+For now, `esp-hal` uses MMU page size as `0x10000` (64k) therefore the flash size has to be changed to 64MB.
```
SergioGasquez, line 6: Do we want to have this duplicated? Id either remove it from this readme or from [`espflash/resources/bootloaders/README.md`](https://github.com/esp-rs/espflash/blob/main/espflash/resources/bootloaders/README.md) as we already forgot to update one of the two places, and it will happen again in the future.

jessebraham: Not sure what you mean, I've deleted that file as I'm not sure why it was added in the first place

espflash/src/targets/esp32c5.rs
SergioGasquez: Yes... And also we would need another branch in https://github.com/esp-rs/espflash/blob/main/espflash/src/bin/espflash.rs#L269-L275

SergioGasquez: Also, we need to update the `-x, --xtal-freq` arg of `save-image` subcommand

SergioGasquez: We could also just support the 40MHz for the moment, and leave the 48MHz support for the future (not even sure if we have 48MHz C5 available ?)

espflash/resources/README.md
SergioGasquez: Oh! I didn't noticed you deleted the `bootloaders/README`, sorry Im just blind!

espflash/src/targets/esp32c5.rs
jessebraham: 48MHz C5 is all I have available 😁

jessebraham: > Yes... And also we would need another branch in https://github.com/esp-rs/espflash/blob/main/espflash/src/bin/espflash.rs#L269-L275, note that this is duplicated in a few places

Do we? The ESP32 also has support for 26MHz xtal, but we don't do the fix here. I thought this was a C2-specific issue?

> Also, we need to update the `-x`, `--xtal-freq` arg of `save-image` subcommand

Doesn't this just use the `XtalFrequency` enum, which I've updated? Or am I misunderstanding?

jessebraham: Regarding the bootloader, seems the only option for xtal on the C5 is autodetect? So I will leave it as-is for now.

jamessizeland:
woop! 🎉

Issue #866, opened 2025-05-23
Detect applications where app desc is required, but not supplied


MabezDev:
The IDF bootloader output is kinda scary and unclear as to how to resolve the issue (without looking at idf docs).

espflash knows what format (i.e idf format) the binary is, plus it has the elf file, so we can do this check for the user and politely warn them that this is going to fail, and how to resolve it.

PR #872, opened 2025-05-30
Check if bootloader has app descriptor

JurajSadel:
This checks, if the bootloader has an app desc present, if not, returns an error. 
Also added an option for users to disable the check.

closes https://github.com/esp-rs/espflash/issues/866

### DIFF
--- a/espflash/src/bin/espflash.rs
+++ b/espflash/src/bin/espflash.rs
@@ -3,6 +3,7 @@ use std::{fs, path::PathBuf};
 use clap::{Args, CommandFactory, Parser, Subcommand};
 use espflash::{
     Error,
+    bt::check_bootloader,
     cli::{
         self,
         config::Config,
@@ -16,7 +17,6 @@ use espflash::{
 };
 use log::{LevelFilter, debug, info};
 use miette::{IntoDiagnostic, Result, WrapErr};
-use object::{File, Object, ObjectSymbol};
 
 #[derive(Debug, Parser)]
 #[command(about, max_term_width = 100, propagate_version = true, version)]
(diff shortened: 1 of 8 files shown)


espflash/src/bin/espflash.rs
```diff
@@ -235,6 +236,16 @@ fn flash(args: FlashArgs, config: &Config) -> Result<()> {
     // Read the ELF data from the build path and load it to the target.
     let elf_data = fs::read(&args.image).into_diagnostic()?;
 
+    if args.flash_args.image.check_app_descriptor.unwrap_or(true) {
```
MabezDev, line 240: We also need this check in cargo-espflash. I suggest taking the guts of this block and extracting into a function in the library, which both binaries can call to do the check.

MabezDev (changes requested):

espflash/src/cli/mod.rs
```diff
@@ -253,6 +253,9 @@ pub struct ImageArgs {
     /// MMU page size.
     #[arg(long, value_name = "MMU_PAGE_SIZE", value_parser = parse_u32)]
     pub mmu_page_size: Option<u32>,
+    /// Flag to check the app descriptor in bootloader
+    #[arg(long, default_value = "true", value_parser = clap::value_parser!(bool))]
+    pub check_app_descriptor: Option<bool>,
```
jessebraham, line 258:
```suggestion
    #[arg(long, default_value_t = true)]
    pub check_app_descriptor: bool,
```

I think we can simplify this? What reason is there to wrap the value in an `Option` here?

espflash/src/bin/espflash.rs
jessebraham: This check should also only be performed when using the `esp-idf` image format, which is why I had suggested holding off on this until those changes had been made.

espflash/src/cli/mod.rs
jessebraham: Wait sorry, I guess then there's no way to disable it 😅

espflash/src/lib.rs
```diff
@@ -76,3 +76,23 @@ pub mod update {
         }
     }
 }
+
+pub mod bt {
```
MabezDev: Let's remove the module and move this to `espflash/src/image_format
/esp_idf.rs`

espflash/src/image_format/esp_idf.rs
```diff
@@ -666,6 +674,19 @@ where
     s
 }
 
+// Check if the provided ELF contains the app descriptor.
```
MabezDev:
```suggestion
/// Check if the provided ELF contains the app descriptor required by [the IDF bootloader](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/bootloader.html).
```

PR #877, opened 2025-06-11
Refactor image format

SergioGasquez:
Draft PR as there are still some missing parts:
- [x] We need to check if the provided `EspIdfFormatArgs`  make sense for the selected subcommand
- There are still some esp-idf specifics that are not in the esp-idf module, maybe we should move them
- Not sure if I like the `make_...` methods, but I cant think of anything better
- Not sure if some things should live in `targets/<chip>.rs` or as method of the Chip enum
  - This can/should be done in a later PR if required.
- [x] There are also some `TODO`s in the code that need to be resolved
- [x] Maybe there is some docstrings missing
- [x] There are still some `ESP-IDF only` annotations

Any feedback or suggestion is more than welcome!

### DIFF
--- a/cargo-espflash/src/main.rs
+++ b/cargo-espflash/src/main.rs
@@ -9,7 +9,7 @@ use clap::{Args, CommandFactory, Parser, Subcommand};
 use espflash::{
     Error as EspflashError,
     cli::{
-        self, FormatArgs,
+        self,
         config::Config,
         monitor::{check_monitor_args, monitor},
         *,
@@ -21,7 +21,7 @@ use espflash::{
     update::check_for_update,
 };
 use log::{LevelFilter, debug, info};
-use miette::{IntoDiagnostic, Result, WrapErr, bail};
+use miette::{IntoDiagnostic, Result, WrapErr};
 
 use crate::{
     cargo_config::CargoConfig,
@@ -355,44 +355,13 @@ fn flash(args: FlashArgs, config: &Config) -> Result<()> {
     if args.flash_args.ram {
         flasher.load_elf_to_ram(&elf_data, Some(&mut EspflashProgress::default()))?;
     } else {
-        let format_args = match args.format {
-            ImageFormatKind::EspIdf => {
-                let mut esp_idf_format_args = args.esp_idf_format_args;
-
-                // Use bootloader from build context if not provided
-                if esp_idf_format_args.bootloader.is_none() {
-                    if let Some(bootloader_path) = build_ctx.bootloader_path {
-                        esp_idf_format_args.bootloader =
-                            Some(fs::read(bootloader_path).into_diagnostic()?);
-                    }
-                }
-
-                // Use partition table from build context if not provided
-                if esp_idf_format_args.partition_table.is_none() {
-                    if let Some(partition_table_path) = build_ctx.partition_table_path {
-                        esp_idf_format_args.partition_table = Some(parse_partition_table(
-                            partition_table_path.to_str().unwrap(),
-                        )?);
-                    }
-                }
-
-                if esp_idf_format_args.erase_parts.is_some()
-                    || esp_idf_format_args.erase_data_parts.is_some()
-                {
-                    erase_partitions(
-                        &mut flasher,
-                        esp_idf_format_args.partition_table.clone(),
-                        esp_idf_format_args.erase_parts.clone(),
-                        esp_idf_format_args.erase_data_parts.clone(),
-                    )?;
-                }
-
-                FormatArgs::EspIdf(esp_idf_format_args)
-            }
-            _ => {
-                bail!("Unsupported format: {:?}", args.format);
-            }
-        };
+        let format_args = cli::create_format_args(
+            args.format,
+            args.esp_idf_format_args,
+            Some(&mut flasher),
+            build_ctx.bootloader_path.as_deref(),
+            build_ctx.partition_table_path.as_deref(),
+        )?;
 
         // TODO: Atm, build_ctx bootloader and part are used over config, thats wrong!
         let flash_data = make_flash_data(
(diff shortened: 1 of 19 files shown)


espflash/src/cli/mod.rs
```diff
@@ -817,20 +812,8 @@ pub fn erase_region(args: EraseRegionArgs, config: &Config) -> Result<()> {
...
-) -> Result<()> {
-    // Load the ELF data, optionally using the provider bootloader/partition
-    // table/image format, to the device's flash memory.
-    flasher.load_elf_to_flash(
-        elf_data,
-        flash_data,
-        Some(&mut EspflashProgress::default()),
-        xtal_freq,
-    )?;
+pub fn flash_elf_image<'a>(flasher: &mut Flasher, image_format: ImageFormat<'a>) -> Result<()> {
```
MabezDev: Given that these don't strictly operate on ELF files anymore, we should rename these and related functions that operate on `ImageFormat` to things like `flash_image` or `load_image_to_flash` etc

cargo-espflash/src/main.rs
```diff
@@ -346,20 +369,32 @@ fn flash(args: FlashArgs, config: &Config) -> Result<()> {
...
+        // If using ESP-IDF image format, check if we need to erase partitions.
+        if let ImageFormat::EspIdf(idf_format) = &image_format {
+            if args.flash_args.erase_parts.is_some() || args.flash_args.erase_data_parts.is_some() {
+                erase_partitions(
+                    &mut flasher,
+                    Some(idf_format.partition_table.clone()),
+                    args.flash_args.erase_parts,
+                    args.flash_args.erase_data_parts,
+                )?;
+            }
```
SergioGasquez: This code snippet is (and was) duplicated in both `cargo-espflash` and `espflash`, shall we do a `maybe_erase_parts` (?) method? 

Also, `check_esp_idf_args` could return a `bool` that could indicate if we need to erase any partition to avoid double checking the format

espflash/src/cli/mod.rs
```diff
@@ -1187,6 +1168,20 @@ pub fn ensure_chip_compatibility(chip: Chip, elf: Option<&[u8]>) -> Result<()> {
...
+) -> Result<()> {
+    if format != ImageFormatKind::EspIdf && (erase_parts.is_some() || erase_data_parts.is_some()) {
+        return Err(miette::miette!(
+            "`erase-parts` and `erase-data` parts are only supported when using the `esp-idf` format."
+        ));
+    }
+
+    Ok(())
+}
+
```
SergioGasquez, line 1179: Creating a fake `Test` `ImageFormatKind` and using `espflash flash  espflash/tests/data/esp32c6 --monitor --format test --erase-parts nvs`  results in:
```
Error: 
  × `erase-parts` and `erase-data` parts are only supported when using the `esp-idf` format.
  
```

espflash/src/targets/esp32c2.rs
```diff
@@ -108,43 +94,6 @@ impl Target for Esp32c2 {
         HashMap::from(encodings)
     }
 
-    fn flash_image<'a>(
-        &self,
-        elf_data: &'a [u8],
-        flash_data: FlashData,
-        _chip_revision: Option<(u32, u32)>,
```
SergioGasquez, line 115: The current `flash_image` does not take/use chip revision as previously it was passed for the `flash` subcommand but then it wasn't used at all.

espflash/src/image_format/esp_idf.rs
```diff
@@ -158,27 +244,34 @@ impl AppDescriptor {
 /// ESP-IDF
 #[derive(Debug)]
 pub struct IdfBootloaderFormat<'a> {
-    params: Esp32Params,
-    bootloader: Cow<'a, [u8]>,
-    partition_table: PartitionTable,
+    boot_addr: u32,
+    pub bootloader: Cow<'a, [u8]>,
```
MabezDev: Do these need to be `pub`? Maybe a getter is cleaner here?

espflash/tests/scripts/write-bin.sh
```diff
@@ -21,22 +21,3 @@ if ! grep -q -a -F $'\x01\xa0' flash_content.bin; then
     echo "Failed verifying content"
     exit 1
 fi
-
-result=$(espflash write-bin nvs binary_file.bin --partition-table $part_table 2>&1)
```
MabezDev, line 25: Why did we remove this test?

SergioGasquez: We agreed on only allow specifying an address now, as the partition name only makes sense for esp-idf format. This tests was testing writing a bin with a part name

MabezDev: Duh, of course - sorry!

espflash/src/image_format/esp_idf.rs
SergioGasquez: Only partition-table needs to be `pub`, just created a getter for it and made it private!  https://github.com/esp-rs/espflash/pull/877/commits/f04650e7db587ffd2e4548e8a8ed7d5895da1497

cargo-espflash/src/main.rs
MabezDev: This check seems small enough that the duplication probably doesn't matter. Given that it's also in the binary/cli part, I care even less, we can refactor this at any time later. Let's leave it as it for now.

SergioGasquez:
Just rebased and added the changelog entry! Thanks @jessebraham and @MabezDev for all your work/help!

espflash/src/flasher/mod.rs
```diff
@@ -476,65 +475,31 @@ impl FlashSettings {
...
-    pub partition_table_offset: Option<u32>,
-    /// Target app partition.
-    pub target_app_partition: Option<String>,
-    /// Flash settings.
     pub flash_settings: FlashSettings,
     /// Minimum chip revision.
     pub min_chip_rev: u16,
     /// MMU page size.
     pub mmu_page_size: Option<u32>,
+    pub chip: Chip,
```
JurajSadel, line 484: Nit: We are missing a doc-string here

espflash/src/image_format/esp_idf.rs
```diff
@@ -506,53 +613,72 @@ impl<'a> IdfBootloaderFormat<'a> {
...
+        HashMap::from([
+            ("app_size", self.app_size.to_string()),
+            ("part_size", self.partition_table_size.to_string()),
+        ])
     }
 
-    /// Get the size of the partition.
-    pub fn part_size(&self) -> Option<u32> {
-        Some(self.part_size)
+    pub fn partition_table(&self) -> PartitionTable {
```
JurajSadel, line 632: Here as well

PR #878, opened 2025-06-11
Minor docs improvement 

JurajSadel:
This could definitely be improved in the future.

Revived https://github.com/esp-rs/espflash/pull/768

cc https://github.com/esp-rs/espflash/issues/874


espflash/src/cli/mod.rs
```diff
@@ -228,9 +228,11 @@ pub struct SaveImageArgs {
     #[arg(long, short = 'x')]
     pub xtal_freq: Option<XtalFrequency>,
     #[clap(flatten)]
+    /// Image arguments
```
MabezDev: Not related to the PR, but I see it and it's bothering me :D. Above this could you fix the typo in Cristal?

espflash/src/cli/mod.rs
```diff
@@ -228,9 +228,11 @@ pub struct SaveImageArgs {
...
     pub image: ImageArgs,
 }
 
+/// Image arguments needed for image generation
```
MabezDev: Most of your comments have a period, but the ones in this file don't. Let's make it consistent.

MabezDev: I noticed a few other places in this PR, could you fix those up too?

JurajSadel: We are inconsistent across whole repo, I can add  period to all comments but it would be huge amount of changes :D

MabezDev: Let's just change the ones added in this PR to have periods, we don't need to change everything.

SergioGasquez: We can do this just before the release if needed, as #877 would also change/remove lots of docstrings

JurajSadel: I added bunch of periods

MabezDev (approved):
Thank you!

Issue #854, opened 2025-04-23
Review public API prior to next release


jessebraham:
The goal is to minimize API surface while still being as useful as possible when using `espflash` as a library for various use cases.

Related to #794

MabezDev:
I spent a bit of time looking into this, here are some notes from me:

 * We likely want to cfg some error cases away that are dependent on the `serialport` feature
 * We have a `Target` trait, with no public implementors, seems to me we should ditch the trait and just implement these directly on the `Chip` enum (and use the chip enum in more places, instead of the `struct Esp32` stuff), especially given this sort of impl in the trait:
   ```rust
   fn chip(&self) -> Chip {
        Chip::Esp32
    }
   ``` 
* Same as above but for the Efuse trait
* Image format module is dependent on changes in https://github.com/esp-rs/espflash/issues/745 so ignoring for now
* A tonne of stuff not respecting the Rust API guidelines, but covered by https://github.com/esp-rs/espflash/issues/864
* Seems like we have some valuable input here: https://github.com/esp-rs/espflash/issues/794#issuecomment-2885299089. Let's take this into account, particularly about making sure `Flasher` can be constructed, and `reset()` via library calls is possible.

Overall, it's not looking to bad, I think most of this could be refactored in a few days. I think we should resolve #745 before tackling these.

I'll also note now, that I don't think we need this to be perfect, we can release another major version of this tool without too much churn later, so to me this is about resolving #745, cleaning up the API as much as we can within a reasonable time constraint then cutting the release.

MabezDev:
Other notes from my comment in #793:

* Regardless of above, I think `Connection` should be constructed outside of `Flasher` to avoid passing in `Connection`'s constructor params into `Flasher::connect`
* I don't know where `SpiAttachParams` should live, it acts like an input but we only ever use `SpiAttachParams::default()` or `SpiAttachParams::esp32_pico_d4()`. If we don't plan on allowing the users to configure this, we should try and associate it with `Chip`, it should simplify a lot of things.
* For cases where there are a large number of input params, that we can't untangle, we should consider a config struct to encapsulate it together.

PR #882, opened 2025-06-12
Move construction of Connection outside of flasher

MabezDev:
This PR implements the first suggestion from https://github.com/esp-rs/espflash/issues/854#issuecomment-2903849411 and enables the creation of a `Connection` outside of `Flasher::connect`.

In general this PR tries to keep the `Connection` specific things, like `sync`ing, baud rate control, reset etc in `Connection` and the flasher related things, like querying spi params, loading the flasher stub etc in `Flasher`.

### DIFF
--- a/espflash/src/flasher/mod.rs
+++ b/espflash/src/flasher/mod.rs
@@ -660,7 +660,7 @@ impl Flasher {
         let detected_chip = if connection.before_operation() != ResetBeforeOperation::NoResetNoSync
         {
             // Detect which chip we are connected to.
-            let detected_chip = detect_chip(&mut connection, use_stub)?;
+            let detected_chip = connection.detect_chip(use_stub)?;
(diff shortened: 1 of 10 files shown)


SergioGasquez (approved):
LGTM once you add the changelog entry! Was there any reason to rename `baud` into `speed`?  We use `baudrate`, `baud_rate`, `baud` and `speed` across the codebase.

espflash/src/flasher/mod.rs
```diff
@@ -1363,6 +1344,11 @@ impl Flasher {
 
         Ok(())
     }
+
+    /// Take the serial port out of the flasher, consuming self
+    pub fn into_serial(self) -> Port {
```
MabezDev: Hmm, this should probably be `into_connection`

PR #883, opened 2025-06-13
Migrate old config

SergioGasquez:
This approach tries to read project config and see if there is a `[connection]` or `[[ubs_device]]` section on it, this would mean that is an old config, and will parse the port_config out of it and write it to `espflash_ports.toml` file (it will not remove the ports sections from the project file though)

### DIFF
--- a/espflash/src/cli/config.rs
+++ b/espflash/src/cli/config.rs
@@ -141,6 +141,26 @@ impl Config {
 
         let mut port_config = if let Ok(data) = read_to_string(&port_config_file) {
             toml::from_str(&data).into_diagnostic()?
+        } else if let Ok(data) = read_to_string(&project_config_file) {
+            if data.contains("[connection]") || data.contains("[[usb_device]]") {
+                debug!(
+                    "espflash@3 configuration detected. Migrating port config to port_config_file: {:#?}",
+                    &port_config_file
+                );
+                let port_config = toml::from_str(&data).into_diagnostic()?;
+                let serialized = toml::to_string(&port_config)
+                    .into_diagnostic()
+                    .wrap_err("Failed to serialize config")?;
+                create_dir_all(port_config_file.parent().unwrap())
+                    .into_diagnostic()
+                    .wrap_err("Failed to create config directory")?;
+                write(&port_config_file, serialized)
+                    .into_diagnostic()
+                    .wrap_err("Failed to write config")?;
+                port_config
+            } else {
+                PortConfig::default()
+            }
         } else {
             PortConfig::default()
         };


espflash/src/cli/config.rs
```diff
@@ -141,6 +141,26 @@ impl Config {
 
         let mut port_config = if let Ok(data) = read_to_string(&port_config_file) {
             toml::from_str(&data).into_diagnostic()?
+        } else if let Ok(data) = read_to_string(&project_config_file) {
+            if data.contains("[connection]") || data.contains("[[usb_device]]") {
```
MabezDev, line 145: I think this as an idea is good, but I think we should at least parse the toml because they might have these commented out or something like that.

We can do this in two ways, we "recreate" the old config struct to try and parse it, or we use https://docs.rs/toml/0.8.23/toml/#parsing-toml to just parse the file as a whole and we can just inspect the keys before making a decision.

MabezDev, line 145: Once we decide that we need to do the migration, we also need to remove the keys from the project config, once the ports file has been succesfully saved. I'm pretty sure if we just call `Config::save_with` all this will be resolved right? It will write over the existing config file with the new format, plus write out the port info into the new file without any of this extra code?

MabezDev: Hmm `Config::save_with` doesn't save the project config... but maybe it should?

SergioGasquez: > because they might have these commented out or something like that.

If they are commented out, then [`let port_config = toml::from_str(&data).into_diagnostic()?;`](https://github.com/esp-rs/espflash/pull/883/files#diff-16aedab0b163d4e6c5350f585959afd77901030799b31de2a9c3c47e64b9c253R150) would not read it. But maybe its better your approach

SergioGasquez: > Hmm Config::save_with doesn't save the project config... but maybe it should?

Not sure if it should, the only scenario when we write config is to store port config when the users selects a port and wants to save it.

MabezDev: Ah okay, if it doesn't make sense to save the project config we can keep the special case save here then.

SergioGasquez:
With the current changes, having the following `espflash.toml`:

```toml
baudrate = 460800
[connection]
serial = "/dev/ttyUSB0"
```
Results in: 
```
[2025-06-16T13:30:02Z INFO ] espflash@3 configuration detected. Migrating port config to port_config_file: "/home/sergio/.config/espflash/espflash_ports.toml"
```
This is the resulting `espflash_ports.toml`
```toml
usb_device = []

[connection]
serial = "/dev/ttyUSB0"
```
And `espflash.toml` got overwritten to:
```toml
baudrate = 460800
format = "EspIdf"

[esp_idf_format_args]

[flash]
```

MabezDev (approved):
LGTM, thanks!

PR #885, opened 2025-06-13
Only change baud if neccesary

SergioGasquez:
https://github.com/esp-rs/espflash/pull/882/files#diff-e53a42b9f8c1ec16423891fcffa2c2a2fda60d4a70f42ddb0771db4ad11b05c0R709-R711 introduced an issue where it would timeout if we try changing the baud to the same baud that is using.

### DIFF
--- a/espflash/src/flasher/mod.rs
+++ b/espflash/src/flasher/mod.rs
@@ -1148,6 +1148,11 @@ impl Flasher {
             new_baud = new_baud * 40 / 26;
         }
 
+        if prior_baud == new_baud {
+            debug!("Baud rate is already set to {}", baud);
+            return Ok(());
+        }
+
         self.connection
             .with_timeout(CommandType::ChangeBaudrate.timeout(), |connection| {
                 connection.command(Command::ChangeBaudrate {


espflash/src/cli/mod.rs
```diff
@@ -453,6 +453,7 @@ pub fn connect(
         !no_verify,
         !no_skip,
         args.chip,
+        args.baud.or(config.project_config.baudrate),
```
SergioGasquez, line 456: This adds a weird arg to the flasher, but the other alternative was to have 2 getters in connection.

espflash/src/flasher/mod.rs
```diff
@@ -706,9 +707,11 @@ impl Flasher {
 
         // Now that we have established a connection and detected the chip and flash
         // size, we can set the baud rate of the connection to the configured value.
-        if flasher.connection.baud() > 115_200 {
-            warn!("Setting baud rate higher than 115,200 can cause issues");
-            flasher.change_baud(flasher.connection.baud())?;
```
SergioGasquez, line 711: This was causing the issue

MabezDev (commented):
LGTM, should we add a change baud rate test?

MabezDev (approved):
Thanks!

Issue #873, opened 2025-06-02
Avoid leaking inner error types


jessebraham:
See the *"Common mistake: inner error type leakage"* section of this blog post for more information:  
https://d34dl0ck.me/rust-bites-designing-error-types-in-rust-libraries/index.html

Given the relatively small amount of effort required, I think it's worth doing this. We can create a type alias for the `Box< .. >` mess and just use that everywhere that we are currently leaking error types.

This is only applicable to _library_ errors, but may be worth just doing everywhere. Less concerned about the CLI errors.

MabezDev:
Oops, I think some of those were added after the initial PR by me 😅. Do you know if there is a lint or check we can do to prevent PRs being merged like this?

bugadani:
[somewhat related rustc issue](https://github.com/rust-lang/rust/issues/120149)

PR #888, opened 2025-06-18
Avoid leaking inner error types in library

JurajSadel:
closes https://github.com/esp-rs/espflash/issues/873

### DIFF
--- a/espflash/src/error.rs
+++ b/espflash/src/error.rs
@@ -19,6 +19,10 @@ use crate::{
     targets::Chip,
 };
 
+// A type alias for a dynamic error that can be used in the library.
+// https://d34dl0ck.me/rust-bites-designing-error-types-in-rust-libraries/index.html
+type CoreError = Box<dyn core::error::Error + Send + Sync>;
+
 /// All possible errors returned by espflash.
 #[derive(Debug, Diagnostic, Error)]
 #[non_exhaustive]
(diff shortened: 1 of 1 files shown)


espflash/src/error.rs
```diff
@@ -19,6 +19,10 @@ use crate::{
     targets::Chip,
 };
 
+// A type alias for a dynamic error that can be used in the library.
+// https://d34dl0ck.me/rust-bites-designing-error-types-in-rust-libraries/index.html
+type CoreError = Box<dyn core::error::Error + Send + Sync>;
```
bugadani, line 24: This is a mistake, library users will not be able to inspect the error this way.

JurajSadel: Can you please elaborate?

bugadani: No, because I'm wrong :) Apparently if one uses `Error::source` they can downcast the result back into the original type if they want to.

espflash/src/error.rs
```diff
@@ -283,6 +287,25 @@ impl From<SlipError> for Error {
     }
 }
 
+impl From<TryFromSliceError> for Error {
+    fn from(err: TryFromSliceError) -> Self {
+        Self::InvalidResponse(err.to_string())
```
bugadani: I don't understand why this doesn't convert into `Error::TryFromSlice`, xor why that variant hasn't been removed.

JurajSadel: Without having it, [this](https://github.com/esp-rs/espflash/blob/main/espflash/src/flasher/mod.rs#L110) (and others) is not happy.

bugadani: But does this have to convert into the wrong error variant?

JurajSadel: Oh, I see now - you are right, probably copy-paste error.

SergioGasquez (approved):
Thanks!

PR #891, opened 2025-06-18
Remove chip target files

MabezDev:
This PR implements the suggestions 2 & 3 from https://github.com/esp-rs/espflash/issues/854#issuecomment-2901766410

As well as some minor refactoring and moving of various pieces to places that made more sense.


MabezDev:
I'd appreciate some thorough testing on this PR, I moved a lot of things around, hopefully all correctly :sweat_smile:.

espflash/src/flasher/mod.rs
```diff
@@ -52,6 +50,16 @@ pub(crate) mod stubs;
 pub(crate) const TRY_SPI_PARAMS: [SpiAttachParams; 2] =
     [SpiAttachParams::default(), SpiAttachParams::esp32_pico_d4()];
 
+/// Progress update callbacks
+pub trait ProgressCallbacks {
```
MabezDev, line 54: Not sure if this belongs here either, but I think it's better than where it was

espflash/src/error.rs
```diff
@@ -554,6 +554,7 @@ impl RomError {
     }
 }
 
+#[cfg(feature = "cli")]
```
MabezDev: Not particularly fond of this, but we get a dead code warning if we don't. Better ideas are most welcome here :D

MabezDev: Maybe it means some partition handling should be in core lib parts, not in the cli module?

MabezDev: Resolved this in last commit, these errors didn't have associated `Error` variants.

espflash/src/flasher/mod.rs
```diff
@@ -974,22 +980,20 @@ impl Flasher {
     /// Read and print any information we can about the connected device
     pub fn device_info(&mut self) -> Result<DeviceInfo, Error> {
         let chip = self.chip();
-        let target = chip.into_target();
-
         // chip_revision reads from efuse, which is not possible in Secure Download Mode
         let revision = (!self.connection.secure_download_mode)
-            .then(|| target.chip_revision(self.connection()))
+            .then(|| chip.chip_revision(self.connection()))
```
SergioGasquez: We should probably rename `chip_revision` to `revision`

espflash/src/image_format/mod.rs
```diff
@@ -18,12 +18,11 @@ use object::{
 };
 use serde::{Deserialize, Serialize};
 
-pub use self::{esp_idf::IdfBootloaderFormat, metadata::Metadata};
-use crate::targets::Chip;
+pub use self::metadata::Metadata;
+use crate::{image_format::idf::IdfBootloaderFormat, targets::Chip};
 
-pub mod esp_idf;
+pub mod idf;
```
SergioGasquez, line 24: I feel like we are not consistent with esp-idf/idf in naming, I think I like idf more (although "IoT Development Framework" doesn't say much), but we still have `EspIdfFormatArgs`, `check_esp_idf_args` `ImageFormatKind::EspIdf`...

espflash/src/targets/mod.rs
```diff
@@ -101,90 +78,215 @@ pub enum Chip {
...
+            Chip::Esp32s3,
+        ] {
+            if chip.has_magic_value(magic) {
+                return Ok(chip);
+            }
         }
+
+        Err(Error::ChipDetectError(format!(
+            "unrecognized magic value: {magic:#x}"
+        )))
```
SergioGasquez:
```suggestion
        Chip::iter()
            .find(|chip| chip.has_magic_value(magic))
            .ok_or_else(|| Error::ChipDetectError(format!("unrecognized magic value: {magic:#x}")))
```

Youd also need to include `strum::IntoEnumIterator`

espflash/src/targets/mod.rs
```diff
@@ -101,90 +78,215 @@ pub enum Chip {
...
+            Chip::Esp32h2 => [0xD7B7_3E80].contains(&value),
+            Chip::Esp32p4 => [0x0, 0x0ADDBAD0].contains(&value),
+            Chip::Esp32s2 => [0x0000_07c6].contains(&value),
+            Chip::Esp32s3 => [0x9].contains(&value),
+        }
+    }
+
+    /// Get the RTC watchdog write protect register address
+    #[cfg(feature = "serialport")]
+    pub fn wdt_wprotect(&self) -> Option<u32> {
```
SergioGasquez: We could use `wdt_wprotect`, `wdt_config0`... to avoid the match in https://github.com/esp-rs/espflash/blob/2c56b23fdf046be5019f22e4621d215ae01cfdc1/espflash/src/targets/flash_target/esp32.rs#L79-L134

we could do: 

``` rust
if chip::Esp32c3 || hip::Esp32s3 || hip::Esp32c6 {
      connection.command(Command::WriteReg {
          address: self.chip.wdt_wprotect().unwrap(),
          value: WDT_WKEY,
          mask: None,
      })?; // WP disable
      connection.command(Command::WriteReg {
          address: self.chip.wdt_config0().unwrap(),
          value: 0x0,
          mask: None,
      })?; // turn off RTC WDT
      connection.command(Command::WriteReg {
          address: self.chip.wdt_wprotect().unwrap(),
          value: 0x0,
          mask: None,
      })?; // WP enable
}
```

(I havent tested it) but something on this lines would avoid having duplicated hardcoded values

espflash/src/targets/mod.rs
```diff
@@ -369,58 +436,348 @@ pub trait ReadEFuse {
...
+    pub fn chip_revision(&self, connection: &mut Connection) -> Result<(u32, u32), Error> {
         let major = self.major_chip_version(connection)?;
         let minor = self.minor_chip_version(connection)?;
 
         Ok((major, minor))
     }
 
     #[cfg(feature = "serialport")]
-    fn major_chip_version(&self, connection: &mut Connection) -> Result<u32, Error>;
+    pub fn major_chip_version(&self, connection: &mut Connection) -> Result<u32, Error> {
```
SergioGasquez: Not sure about this one, but maybe we should also remove the `chip`part of the name (and if we do, ofc we should also update the minor method too)

espflash/src/targets/mod.rs
```diff
@@ -369,58 +436,348 @@ pub trait ReadEFuse {
...
+
+                Ok((hi << 3) + lo)
+            }
+        }
+    }
 
     #[cfg(feature = "serialport")]
     /// What is the crystal frequency?
-    fn crystal_freq(&self, connection: &mut Connection) -> Result<XtalFrequency, Error>;
+    pub fn crystal_freq(&self, connection: &mut Connection) -> Result<XtalFrequency, Error> {
```
SergioGasquez: I guess at this point I'm only complaining about names (feel free to ignore me hahaha), but also we are using Xtal/Crystal

espflash/src/targets/mod.rs
```diff
@@ -369,58 +436,348 @@ pub trait ReadEFuse {
...
+                    XtalFrequency::_40Mhz
+                };
+
+                Ok(norm_xtal)
+            }
+            Chip::Esp32c6 => Ok(XtalFrequency::_40Mhz), // Fixed frequency
+            Chip::Esp32h2 => Ok(XtalFrequency::_32Mhz), // Fixed frequency
+            Chip::Esp32p4 => Ok(XtalFrequency::_40Mhz), // Fixed frequency
+            Chip::Esp32s2 => Ok(XtalFrequency::_40Mhz), // Fixed frequency
+            Chip::Esp32s3 => Ok(XtalFrequency::_40Mhz), // Fixed frequency
```
SergioGasquez: extra nitpick: we can group all the 40 Mhz ones?

SergioGasquez (commented):
Niice `+789 −1,460`, thanks! Left some thoughts/suggestions

espflash/src/targets/mod.rs
MabezDev: Good call! I improved it a bit by matching on wdtprotect and wdtconfig0. That way we won't have to update this function anymore for new chips. Thanks!

espflash/src/targets/flash_target/esp32.rs
```diff
@@ -23,6 +23,8 @@ use crate::{
     targets::FlashTarget,
 };
 
+const WDT_WKEY: u32 = 0x50D8_3AA1; // Watchdog key for disabling the watchdog timers
```
SergioGasquez: This is also duplicated in `targets/mod.rs` maybe we can de duplicate it?

SergioGasquez (approved):
LGTM once we update the changelog! Feel free to dedup or ignore my latest comment

.gitignore
```diff
@@ -14,3 +14,5 @@ target/
 .DS_Store
 *.swp
 .idea/
+
+!espflash/src/target
```
SergioGasquez, line 18: 😂

MabezDev: I had some fun with that initially :D

Issue #886, opened 2025-06-14
`check-app-descriptor true` is `esp-hal`-specific


ivmarkov:
It is `esp-hal` specific, because it searches for the `esp_app_desc` symbol in an ELF section whose name is hard-coded to be [`.rodata_desc`](https://github.com/esp-rs/espflash/blob/dd0b18d3c6735f652ff4de1ae9318b7fd2e16841/espflash/src/image_format/esp_idf.rs#L819).

While the `esp-hal` linker script is putting the `esp_app_desc` symbol in a section under that name I'm sure, the linker script of ESP-IDF is a bit more complex than that, in that it merges the `.rodata_desc` section as well as a few others - in predefined order though - into a larger [`.flash.rodata`](https://github.com/espressif/esp-idf/blob/4c3d086c13b35ee671e5e7f3d92f1ccbaedc5fd0/components/esp_tee/subproject/main/ld/esp32c6/esp_tee.ld.in#L148) section.

So if `espflash` is to support ESP-IDF app images (Rust or C ones), then this check needs to be relaxed. Perhaps, by _also_ searching in a section named `.flash_rodata`.

Sure, one can always skip the check with `--check-app-descriptor false`, but you could imagine - with the new `espflash` release - we have to explain this to _every single user_ trying out the `esp-idf-*` crates which is not ideal.

Furthermore, I would suggest relaxing the error message to mention that using the `esp-bootloader-esp-idf` is `esp-hal` specific, i.e.:

- Now: "The app descriptor is not present in the project. You need to add the https://github.com/esp-rs/esp-hal/tree/main/esp-bootloader-esp-idf to your project."
- Then: "The app descriptor is not present in the project. If using `esp-hal`, add the https://github.com/esp-rs/esp-hal/tree/main/esp-bootloader-esp-idf to your project. If using `esp-idf-*`, the app descriptor is added implicitly, but you can override its default values by declaring `esp_idf_svc::sys::esp_app_desc! {}`".
- Then, Option 2: "The app descriptor is not present in the project. If using `esp-hal`, add the https://github.com/esp-rs/esp-hal/tree/main/esp-bootloader-esp-idf to your project. (As this error should never, ever appear when using ESP-IDF anyway.)

ivmarkov:
@bugadani FYI ^^^

ivmarkov:
> I started to agree with this, but your last point:
> 
[...]

PR #892, opened 2025-06-18
Check `app_desc` _only_ for esp-hal projects

JurajSadel:
Note: For testing non-esp-hal based project, I used ESP-IDF's hello_world example, and I hope it is valid test.

closes https://github.com/esp-rs/espflash/issues/886

### DIFF
--- a/espflash/src/error.rs
+++ b/espflash/src/error.rs
@@ -266,9 +266,9 @@ pub enum Error {
     )]
     PartitionTableDoesNotFit(FlashSize),
 
-    #[error("{error}")]
+    #[error("{0}")]
     #[diagnostic(code(espflash::app_desc::app_descriptor_not_present))]
-    AppDescriptorNotPresent { error: String },
+    AppDescriptorNotPresent(String),
 }
 
 #[cfg(feature = "serialport")]
--- a/espflash/src/error.rs
+++ b/espflash/src/error.rs
@@ -266,11 +266,9 @@ pub enum Error {
     )]
     PartitionTableDoesNotFit(FlashSize),
 
-    #[error(
-        "The app descriptor is not present in the project. You need to add the https://github.com/esp-rs/esp-hal/tree/main/esp-bootloader-esp-idf to your project."
-    )]
+    #[error("{error}")]
     #[diagnostic(code(espflash::app_desc::app_descriptor_not_present))]
-    AppDescriptorNotPresent,
+    AppDescriptorNotPresent { error: String },
 }
 
 #[cfg(feature = "serialport")]
--- a/espflash/src/image_format/idf.rs
+++ b/espflash/src/image_format/idf.rs
@@ -814,14 +814,33 @@ where
 /// Check if the provided ELF contains the app descriptor required by [the IDF bootloader](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/bootloader.html).
 pub fn check_idf_bootloader(elf_data: &Vec<u8>) -> Result<()> {
     let object = File::parse(elf_data.as_slice()).into_diagnostic()?;
-    let section = object.section_by_name(".rodata_desc").is_some();
-    let esp_hal = object.section_by_name(".espressif.metadata").is_some();
-    let symbol = object.symbols().any(|sym| sym.name() == Ok("esp_app_desc"));
 
-    // esp-hal specific, because it searches for the hard-coded esp_app_desc symbol
-    // in an ELF section
-    if esp_hal && (!section || !symbol) {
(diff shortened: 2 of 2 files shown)


espflash/src/image_format/esp_idf.rs
```diff
@@ -817,9 +817,12 @@ where
...
     let object = File::parse(elf_data.as_slice()).into_diagnostic()?;
     let section = object.section_by_name(".rodata_desc").is_some();
+    let esp_hal = object.section_by_name(".espressif.metadata").is_some();
     let symbol = object.symbols().any(|sym| sym.name() == Ok("esp_app_desc"));
 
-    if !section || !symbol {
+    // esp-hal specific, because it searches for the hard-coded esp_app_desc symbol
+    // in an ELF section
+    if esp_hal && (!section || !symbol) {
         return Err(Error::AppDescriptorNotPresent).into_diagnostic();
```
MabezDev: I think this bit still needs some more work, based on my comment here: https://github.com/esp-rs/espflash/issues/886#issuecomment-2975647493

We should still return an error if the section is missing from _either_ type of binary, but we should only tell users to use `esp-bootloader-idf` _if_ we detect that it's an esp-hal binary. I'm hoping we can customize the error context with `thiserror` or `miette` for this purpose :crossed_fingers:

espflash/src/error.rs
```diff
@@ -266,11 +266,9 @@ pub enum Error {
     )]
     PartitionTableDoesNotFit(FlashSize),
 
-    #[error(
-        "The app descriptor is not present in the project. You need to add the https://github.com/esp-rs/esp-hal/tree/main/esp-bootloader-esp-idf to your project."
-    )]
+    #[error("{error}")]
     #[diagnostic(code(espflash::app_desc::app_descriptor_not_present))]
-    AppDescriptorNotPresent,
+    AppDescriptorNotPresent { error: String },
```
jessebraham: Sorry to nitpick, but there's no benefit to naming the inner value here

```suggestion
    AppDescriptorNotPresent(String),
```

The attribute then turns into 

```rust
#[error("{0}")]
```

And you no longer need to specify `error:` when constructing this error

espflash/src/image_format/idf.rs
```diff
@@ -814,11 +814,33 @@ where
...
     let object = File::parse(elf_data.as_slice()).into_diagnostic()?;
-    let section = object.section_by_name(".rodata_desc").is_some();
-    let symbol = object.symbols().any(|sym| sym.name() == Ok("esp_app_desc"));
 
-    if !section || !symbol {
-        return Err(Error::AppDescriptorNotPresent).into_diagnostic();
+    // Check for section and symbol presence in esp-hal based projects
+    let section = object.section_by_name(".rodata_desc").is_some();
+    let is_esp_hal = object.section_by_name(".espressif.metadata").is_some();
+    let esp_hal_symbol = object.symbols().any(|sym| sym.name() == Ok("esp_app_desc"));
```
MabezDev: I think this check is a bit too complicated now.

* Both idf and esp-hal export the symbol as `esp_app_desc` so we need to check that in both. We should just check the symbol name (globally), as we do here (I would rename this variable to `has_app_desc`.
* Once we know if we need to report an error or not (in the case of missing desc), we must then decipher what kind of app it is, we already have that check too, `is_esp_hal`.
* I no longer think we need to check the section names, as `esp_app_desc` is global therefore only one instance can exist in a binary.

With this in mind, I believe the following if block becomes:

```rust
if !has_app_desc {
   if is_esp_hal {
		// esp-hal custom error
    } else {
		// idf custom error
    } 
}
```

JurajSadel: Hmm, for some reason grep didn't find any occurrence of `esp_app_desc` in idf app - I double-checked the relevant section manually and I found it - thanks!

MabezDev (approved):
LGTM, thanks!

espflash/src/image_format/idf.rs
```diff
@@ -814,11 +814,25 @@ where
...
-    if !section || !symbol {
-        return Err(Error::AppDescriptorNotPresent).into_diagnostic();
+    let has_app_desc = object.symbols().any(|sym| sym.name() == Ok("esp_app_desc"));
+    let is_esp_hal = object.section_by_name(".espressif.metadata").is_some();
+
+    if !has_app_desc {
+        if is_esp_hal {
+            return Err(Error::AppDescriptorNotPresent (
+                "The app descriptor is not present in the `esp-hal` based project.\n\
+                        You need to add the https://github.com/esp-rs/esp-hal/tree/main/esp-bootloader-esp-idf \
```
MabezDev: Oh actually, one last thing, could we update this to the crates.io link, now its released?

PR #893, opened 2025-06-18
Update deps and adapt code

SergioGasquez:
Since we plan to release soon :tm:, lets bump the deps and test them before the release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -426,7 +426,7 @@ version = "7.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4a65ebfec4fb190b6f90e944a817d60499ee0744e582530e2c9900a22e591d9a"
 dependencies = [
- "crossterm 0.28.1",
+ "crossterm",
  "unicode-segmentation",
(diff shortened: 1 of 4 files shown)


cargo-espflash/src/main.rs
```diff
@@ -545,7 +545,7 @@ fn build(
                     continue;
                 };
 
-                if package.name != "esp-idf-sys" {
+                if package.name.to_string() != "esp-idf-sys" {
```
bugadani:
```suggestion
                if package.name.as_ref() != "esp-idf-sys" {
```

No need to clone the string to compare it

Cargo.lock
```diff
@@ -451,11 +494,26 @@ version = "0.28.1"
...
- "bitflags 2.9.0",
+ "bitflags 2.9.1",
+ "crossterm_winapi",
+ "parking_lot",
+ "rustix 0.38.44",
+ "winapi",
+]
+
+[[package]]
+name = "crossterm"
```
bugadani: I'd prefer not pulling in a second version of crossterm

MabezDev (approved):

PR #896, opened 2025-06-20
Don't trip up on non-code sections

### LINKED ISSUE #894: Segments are overlapping
@bugadani discovered this issue, it fails at least with esp32 with:

```
Error: 
  × Main thread panicked.
  ├─▶ at espflash/src/image_format/esp_idf.rs:587:35
  ╰─▶ range end index 4294967289 out of range for slice of length 4
  help: set the `RUST_BACKTRACE=1` environment variable to display a backtrace.
``` 

And the error was introduced in https://github.com/esp-rs/espflash/pull/839, using the previous commit before merging that PR works.

Seems like for whatever reason, some segments are overlapping

bugadani:
Fixes #894


SergioGasquez (approved):
LGTM! Thanks!

PR #899, opened 2025-06-20
Improve docs

SergioGasquez:
Added some docstrings(some, where not necessary, ie: cargo-espflash) and fixed some typos

Regarding the period at the end of the sentences I can add it for all the docstrings if we want but that will introduce a lot of noise in the PR, maybe its better a different PR?


MabezDev (approved):
Thank you!

Regarding the periods, don't worry about it too much, I just think that new comments should have the proper punctuation.

PR #901, opened 2025-06-24
Address library feedback

MabezDev:
* Make `Command` and friends available outside the connection structure.
* Respect the `self.verify` and `self.skip` flash on `Flasher`

### DIFF
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -17,6 +17,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  RUST_LOG: debug
 
 # Cancel any currently running workflows from the same PR, branch, or
 # tag when a new workflow is triggered.
@@ -113,7 +114,7 @@ jobs:
 
       - name: save-image/write-bin test
         run: |
-          timeout 90 bash espflash/tests/scripts/save-image_write-bin.sh ${{ matrix.board.mcu }}
+          timeout 180 bash espflash/tests/scripts/save-image_write-bin.sh ${{ matrix.board.mcu }}
 
       - name: erase-region test
         run: timeout 30 bash espflash/tests/scripts/erase-region.sh
--- a/espflash/src/flasher/mod.rs
+++ b/espflash/src/flasher/mod.rs
@@ -1116,15 +1116,17 @@ impl Flasher {
 
     /// Get MD5 of region
     pub fn checksum_md5(&mut self, addr: u32, length: u32) -> Result<u128, Error> {
-        self.connection
-            .with_timeout(CommandType::FlashMd5.timeout(), |connection| {
+        self.connection.with_timeout(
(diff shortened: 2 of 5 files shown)


espflash/src/flasher/mod.rs
```diff
@@ -1102,9 +1099,9 @@ impl Flasher {
             });
         }
 
-        let mut target = self
-            .chip
-            .flash_target(self.spi_params, self.use_stub, false, false);
+        let mut target =
+            self.chip
+                .flash_target(self.spi_params, self.use_stub, self.verify, self.skip);
```
JurajSadel, line 1104: I tried to test this (`save-image_write-bin.sh`) and I can see both `self.verify` and `self.skip` are `true` now - just double checking if it was intended change or not

MabezDev: I think this PR is exposing that skip/verify is (and hasn't) be working correctly for a while, at least on an intermittent basis. They should have been true always, but they were hard coded to false.

MabezDev: It looks like it hasn't changed since the original impl in https://github.com/esp-rs/espflash/pull/538, cc: @bjoernQ, any reason why this was done this way?

bjoernQ: t.b.h. I don't remember why `write_bin` didn't have a skip or a verify command line flag

.github/workflows/hil.yml
```diff
@@ -17,6 +17,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  RUST_LOG: debug
```
MabezDev, line 20: I can remove this, but I think it's quite useful when things go wrong. If a different level is better, let me know.

espflash/tests/scripts/save-image_write-bin.sh
```diff
@@ -12,7 +12,7 @@ if [[ ! $result =~ "Image successfully saved!" ]]; then
     exit 1
 fi
 echo "Writing binary"
-result=$(timeout 45 espflash write-bin --monitor 0x0 app.bin --non-interactive 2>&1)
+result=$(timeout 90 espflash write-bin --monitor 0x0 app.bin --non-interactive 2>&1)
```
MabezDev, line 15: This looks like I'm trying to hide the issue, but I believe that now `verify` and `skip` are respected, they are now used by default. This adds time to CI, and on some chips this takes longer than 45s.

MabezDev: The same for the timeout in the hi.yml

.github/workflows/hil.yml
jessebraham: No strong feelings either way, only usually check the CI logs when things fail anyway so probably makes sense to keep it in, at least for now.

jessebraham (approved):

PR #903, opened 2025-06-25
Make eFuse field definitions public, minor public API cleanup

jessebraham:
Major change here is making the eFuse field definitions public. There was a minor change for the P4 which required some updates to read the version correctly again.

Also removed unnecessary parameters/Result types from a couple functions which do not require them.

### DIFF
--- a/espflash/src/target/mod.rs
+++ b/espflash/src/target/mod.rs
@@ -10,18 +10,15 @@ use serde::{Deserialize, Serialize};
 use strum::{Display, EnumIter, EnumString, IntoEnumIterator, VariantNames};
 
 #[cfg(feature = "serialport")]
-pub use self::{
-    efuse::EfuseField,
-    flash_target::{Esp32Target, FlashTarget, RamTarget},
-};
+pub use self::flash_target::{Esp32Target, FlashTarget, RamTarget};
 use crate::{
     Error,
     flasher::{FLASH_WRITE_SIZE, FlashFrequency},
 };
 #[cfg(feature = "serialport")]
-use crate::{connection::Connection, flasher::SpiAttachParams};
+use crate::{connection::Connection, flasher::SpiAttachParams, target::efuse::EfuseField};
 
-mod efuse;
+pub mod efuse;
 
 #[cfg(feature = "serialport")]
 pub(crate) mod flash_target;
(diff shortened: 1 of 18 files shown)


espflash/src/target/mod.rs
```diff
@@ -10,14 +10,16 @@ use serde::{Deserialize, Serialize};
 use strum::{Display, EnumIter, EnumString, IntoEnumIterator, VariantNames};
 
 #[cfg(feature = "serialport")]
-pub use self::flash_target::{Esp32Target, FlashTarget, RamTarget};
-use crate::{Error, flasher::FlashFrequency};
-#[cfg(feature = "serialport")]
+pub use self::{
+    efuse::EfuseField,
```
MabezDev: This needs a changelog entry

espflash/src/target/mod.rs
```diff
@@ -10,14 +10,16 @@ use serde::{Deserialize, Serialize};
...
-    connection::Connection,
-    flasher::{FLASH_WRITE_SIZE, SpiAttachParams},
-    target::{efuse::EfuseField, flash_target::MAX_RAM_BLOCK_SIZE},
+    Error,
+    flasher::{FLASH_WRITE_SIZE, FlashFrequency},
 };
+#[cfg(feature = "serialport")]
+use crate::{connection::Connection, flasher::SpiAttachParams};
 
 mod efuse;
```
MabezDev: Should this be public (along with the chip specific modules inside)? Or do we expect users craft the `EfuseField`'s themselves for now? I'm fine with that, as we can always add it later. Just checking this is the intention :smile:.

xtask/src/main.rs
```diff
@@ -188,7 +188,7 @@ fn process_efuse_definitions(efuse_fields: &mut EfuseFields) -> Result<()> {
 fn generate_efuse_definitions(espflash_path: &Path, efuse_fields: EfuseFields) -> Result<()> {
     let targets_efuse_path = espflash_path
         .join("src")
-        .join("targets")
+        .join("target")
```
MabezDev, line 193: Oops, forgot to update here when I renamed the module :sweat_smile:, thanks for fixing!

espflash/src/target/mod.rs
jessebraham: Oh, oops. Forgot to re-export the field definitions. Good catch

MabezDev (approved):
LGTM, thanks!

PR #904, opened 2025-06-25
Show progress when skipping and verifying

MabezDev:
The progress callbacks get initialized and updated even if we skip a segment. For verification, we add an extra step to the progress which is updated once verification is complete.

### DIFF
--- a/espflash/src/target/flash_target/mod.rs
+++ b/espflash/src/target/flash_target/mod.rs
@@ -1,5 +1,5 @@
 pub use self::{esp32::Esp32Target, ram::RamTarget};
-use crate::{Error, connection::Connection, flasher::ProgressCallbacks, image_format::Segment};
+use crate::{Error, connection::Connection, image_format::Segment};
 
 mod esp32;
 mod ram;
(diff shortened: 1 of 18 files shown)


MabezDev:
Eh, I went a different way. I removed the optionality and then added an empty impl as the "do nothing" value.

espflash/src/target/flash_target/mod.rs
```diff
@@ -14,9 +14,29 @@ pub trait FlashTarget {
...
         segment: Segment<'_>,
-        progress: &mut Option<&mut dyn ProgressCallbacks>,
+        progress: &mut dyn ProgressCallbacks,
     ) -> Result<(), Error>;
 
     /// Complete the flashing operation
     fn finish(&mut self, connection: &mut Connection, reboot: bool) -> Result<(), Error>;
 }
+
+/// Progress update callbacks
```
JurajSadel: Should we care about periods? 😆

jessebraham: Let's not worry about it right now, I will be submitting a PR working on documentation soon.

MabezDev: I added it because I can't be nitpicky on Juraj's PR and let this slide on mine :laughing:

jessebraham (approved):
Thanks, LGTM!

PR #905, opened 2025-06-25
Avoid leaking inner error types vol.2

JurajSadel:
This turned out to be more complicated than expected 

closes https://github.com/esp-rs/espflash/issues/873#issuecomment-3003826881 (again :D )


espflash/src/error.rs
```diff
@@ -271,6 +268,28 @@ pub enum Error {
...
+    }
+}
+
+#[cfg(feature = "serialport")]
+impl core::error::Error for SlipErrorWrapper {}
+
+#[cfg(feature = "serialport")]
+impl From<SlipError> for Error {
+    fn from(err: SlipError) -> Self {
+        Self::Connection(Box::new(SlipErrorWrapper(err)))
```
bugadani: You already have a `impl From<SlipError> for ConnectionError {` implementation in this file, you don't need a wrapper type.

espflash/src/error.rs
```diff
@@ -604,13 +607,24 @@ impl<T> ResultExt for Result<T, Error> {
...
-        match self {
-            Err(Error::Connection(ConnectionError::Timeout(_))) => {
-                Err(Error::Connection(ConnectionError::Timeout(command.into())))
-            }
-            Err(Error::Flashing(ConnectionError::Timeout(_))) => {
-                Err(Error::Flashing(ConnectionError::Timeout(command.into())))
+        fn remap_if_timeout(
+            err: Box<dyn std::error::Error + Send + Sync>,
+            command: CommandType,
+        ) -> Box<dyn std::error::Error + Send + Sync> {
```
jessebraham:
```suggestion
        fn remap_if_timeout(err: CoreError, command: CommandType) -> CoreError {
```

Can we just use the alias here to clean things up a bit and be consistent, please?

jessebraham (changes requested):
Just one minor thing I'd like changed please, otherwise LGTM

jessebraham (approved):

PR #906, opened 2025-06-26
Move non-interactive flag to connect args

SergioGasquez:
Moved the `non-interactive` flag to `ConnectArgs` so we also avoid asking the user to select a port. The logic added is the following: When using `non-interactive`, it will use any of the selected ports (via flag, environment variable or config file) and if no port was selected it will: Detect the number of ports, if there is only one, it will use this one, if there are multiple ports it will throw an error (MacOs usually duplicates ports so its not ideal, maybe we should error if >2 on macos?)

### DIFF
--- a/espflash/src/cli/serial.rs
+++ b/espflash/src/cli/serial.rs
@@ -113,6 +113,18 @@ pub(super) fn detect_usb_serial_ports(list_all_ports: bool) -> Result<Vec<Serial
     let ports = ports
         .into_iter()
         .filter(|port_info| {
+            // On macOS, each USB serial device is exposed twice: once as a
+            // "/dev/tty.*" callin device and once as a "/dev/cu.*" callout
+            // device. The latter is what users typically want because opening
+            // a tty.* device will block until a DCD signal is asserted.  To
+            // avoid confusing users with duplicate entries we hide the tty.*
+            // variants unless the caller explicitly asked to list *all* ports.
+            // See https://web.archive.org/web/20120602152224/http://lists.berlios.de/pipermail/gpsd-dev/2005-April/001288.html
+            #[cfg(target_os = "macos")]
+            if !list_all_ports && port_info.port_name.starts_with("/dev/tty.") {
+                return false;
+            }
+
             if list_all_ports {
                 matches!(
                     &port_info.port_type,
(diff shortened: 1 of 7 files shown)


JurajSadel (commented):
I've tried to run some test locally:
```rust
./espflash/tests/scripts/save-image_write-bin.sh --port /dev/cu.usbmodem11401  --non-interactive esp32s3
error: unexpected argument '--port' found

  tip: a similar argument exists: '--format'

Usage: espflash save-image --chip <CHIP> --merge --format <FORMAT> <IMAGE> <FILE>

For more information, try '--help'.
➜  espflash git:(1aa4970) ✗ ./espflash/tests/scripts/save-image_write-bin.sh -p /dev/cu.usbmodem11401  --non-interactive esp32s3
error: unexpected argument '-p' found

  tip: to pass '-p' as a value, use '-- -p'

Usage: espflash save-image [OPTIONS] --chip <CHIP> <IMAGE> <FILE>

For more information, try '--help'.
```

Am I missing something here?

jessebraham:
> MacOs usually duplicates ports so its not ideal, maybe we should error if >2 on macos?

Probably we can just filter the ports on macOS, e.g. to exclude `tty.*`, rather than allowing 2 ports.

JurajSadel (approved):
LGTM, thanks!

espflash/src/cli/serial.rs
```diff
@@ -110,6 +113,18 @@ pub(super) fn detect_usb_serial_ports(list_all_ports: bool) -> Result<Vec<Serial
...
         .filter(|port_info| {
+            // On macOS, each USB serial device is exposed twice: once as a
+            // "/dev/tty.*" callin device and once as a "/dev/cu.*" callout
+            // device. The latter is what users typically want because opening
+            // a tty.* device will block until a DCD signal is asserted.  To
+            // avoid confusing users with duplicate entries we hide the tty.*
+            // variants unless the caller explicitly asked to list *all* ports.
+            // See https://web.archive.org/web/20120602152224/http://lists.berlios.de/pipermail/gpsd-dev/2005-April/001288.html
+            #[cfg(target_os = "macos")]
+            if !list_all_ports && port_info.port_name.starts_with("/dev/tty.") {
```
MabezDev, line 124: Can the user set this via the CLI if this logic ends up being wrong on some mac systems?

MabezDev (approved):
Thanks!

PR #907, opened 2025-06-26
Avoid leaking inner error types vol.3

JurajSadel:


jessebraham (approved):
Thanks, sorry again for missing this in the previous review 😅

Issue #794, opened 2025-02-27
Investigate espflash library usage


MabezDev:
We don't have much from crates.io rev deps, but what we can, and should do is do a code search on Github/Gitlab to find library usage. From here, we can find what needs to be exposed in the library and what doesn't. It's also not that black and white, a module _may_ be used because our other APIs are insufficient, or are tangled together forcing users to use both modules so keep that in mind.

jessebraham:
I did a quick review of our public API in `main`, and as far as I can tell `esp_idf_part::PartitionTable` is the only type being leaked from a non-stable dependency. Should be easy enough to fix probably, only present in a few places and probably at least some of them will disappear with other refactoring we plan to do.

jessebraham:
A few projects which appears to be using `espflash` directly as a library:

- https://github.com/probe-rs/probe-rs/
- https://github.com/yanshay/esp-hal-app
- https://github.com/EyeTrackVR/FirmwareFlashingTool
- https://github.com/ivmarkov/espfactory
- https://github.com/3STechLabs-org/ESP32-Flasher
- https://github.com/QTShock/QTHub

This is not comprehensive, but investigating how these projects are using the library should at least give us a reasonable idea of how people are currently using `espflash`.

nullstalgia:
I'm currently working on a binary crate (serial-terminal adjacent, still an early WIP) that intends to use `espflash` as a library! (I had just started to sniff through issues here to get a clearer picture on what I might need to look out for, and ran into this issue near the top. 😄)

But so far, something I've **really** appreciated is the ability to create helpers like `Connection`s and `Flasher`s, and then be able to consume them and get back the `Port` object for my terminal to use again without reconnecting! The `ResetStrategy`'s request for a `&mut Port` also makes this easy, and is also appreciated.

If you have more questions about how I intend to use it as a library, or would like feedback as I get deeper into the project, I'd be happy to oblige!

i404788:
Also wanted to chime in and say I'm planning to use parts of espflash currently not yet exposed (mainly the `Command` struct) for an in-circuit esp programmer (which can't use `serialport` cause no_std). It's still early days & probably won't be fully fleshed out until a few months, but if anything of note comes up I'll post an update. Also happy to answer questions or contribute if needed.

nullstalgia:
Howdy, been a second but I'm finally at the stage in my application where I'm able to focus on the fancier things like espflash integration.

[...]

PR #908, opened 2025-06-28
Indicate in CLI when a partition was skipped or being verified, fix incorrect chunk size being shown.

nullstalgia:
This is meant to extend #904 to utilize the "skipped" flag in CLI, show an example usage of the `verifying` [signal I suggested](https://github.com/esp-rs/espflash/issues/794#issuecomment-3006635559) in the conversation in #794, and fix what I consider to be a (overall minor) mistake in #904.

Showcase for Verifying:

[Screencast_20250627_183019.webm](https://github.com/user-attachments/assets/4c6db327-b8f1-4736-8fb5-14b7784dfc7d)

Showcase for Skipping: 

[Screencast_20250627_182941.webm](https://github.com/user-attachments/assets/19d996f0-e97f-41b6-8a48-bbc4edc9fc53)

In the past, `espflash` seemed to pause for a moment during post-flash verification, and with #904, that was changed to add a ghost chunk to the reported total length of the segment that would be paused on during verification.

I personally disagree with this specific change, since it also changes the reported output to users on how large their uploads are, and could potentially lead some to the wrong idea on why things are taking extra time. So I chose to make it much more clear as to what actions are being taken.

Before #904:
image

After #904:
image

(Note the +1 to each segment's reported length).

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -745,13 +745,15 @@ pub(crate) fn display_image_size(app_size: u32, part_size: Option<u32>) {
 #[derive(Debug, Default)]
 pub struct EspflashProgress {
     pb: Option<ProgressBar>,
+    current_addr: Option<u32>,
+    verifying: bool,
 }
 
 impl ProgressCallbacks for EspflashProgress {
     /// Initialize the progress bar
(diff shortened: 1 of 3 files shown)


jessebraham (approved):
Very nice, thanks for this!

PR #909, opened 2025-06-30
Document all public types

jessebraham:
Probably still room for improvement with regards to docs, but this at least ensures all public types are documented and enforces documentation on public types.

Some other related changes here, nothing too interesting though. Will probably have one more small PR for documentation after this, but need to review everything first.


SergioGasquez (approved):
LGTM! Thanks

PR #910, opened 2025-06-30
Target module fixups

MabezDev:


espflash/src/target/mod.rs
```diff
@@ -91,7 +91,7 @@ impl Chip {
     pub fn from_magic(magic: u32) -> Result<Self, Error> {
         for chip in Chip::iter() {
             if chip.has_magic_value(magic) {
-                return Ok(chip);
+                Returns Ok(chip);
```
SergioGasquez: Mmmm you went too far with search and replace

MabezDev: Oops, I hit match whole word instead of match word and case in vscode :sweat_smile:

SergioGasquez (approved):
LGTM!

jessebraham (approved):
Thanks!

PR #912, opened 2025-06-30
Add xtask command for running HIL tests

JurajSadel:
Opening as a draft so I can test on the HIL runners.


xtask/src/test_runner.rs
```diff
@@ -0,0 +1,904 @@
...
+
+    fn setup_command(&self, cmd: &mut Command) {
+        cmd.current_dir(&self.workspace)
+            .stdout(Stdio::inherit())
+            .stderr(Stdio::inherit());
+    }
+
+    fn terminate_process(child_id: u32, child: &mut Option<&mut Child>) {
+        unsafe {
+            libc::kill(child_id as i32, libc::SIGTERM);
```
bugadani: Why do you need libc for this? Child has a kill method

JurajSadel: Removed, thanks!

xtask/Cargo.toml
```diff
@@ -11,3 +11,5 @@ env_logger = "0.11.8"
 log        = "0.4.27"
 serde      = { version = "1.0.217", features = ["derive"] }
 serde_yaml = "0.9.34"
+libc       = "0.2.152"
+winapi     = { version = "0.3.9", features = ["processthreadsapi", "handleapi"] }
```
bugadani: These don't seem to be used anywhere?

espflash/tests/flash_content.bin
```diff
@@ -0,0 +1 @@
...������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������������
```
SergioGasquez: This should be git ignored (?)

xtask/Cargo.toml
```diff
@@ -11,3 +11,4 @@ env_logger = "0.11.8"
 log        = "0.4.27"
 serde      = { version = "1.0.217", features = ["derive"] }
 serde_yaml = "0.9.34"
+libc       = "0.2.152"
```
jessebraham: This dependency is unused

jessebraham:
Couple small issues noted, will do a proper review tomorrow

xtask/src/main.rs
```diff
@@ -1,30 +1,24 @@
...
     /// Generate eFuse field definitions
-    GenerateEfuseFields(GenerateEfuseFieldsArgs),
-}
+    GenerateEfuseFields(efuse_generator::GenerateEfuseFieldsArgs),
 
-#[derive(Debug, Args)]
-struct GenerateEfuseFieldsArgs {
-    /// Local path to the `esptool` repository
-    esptool_path: PathBuf,
+    /// Run espflash tests (replacing bash scripts)
```
MabezDev:
```suggestion
    /// Run espflash tests
```

xtask/src/test_runner.rs
```diff
@@ -0,0 +1,904 @@
...
+        }
+
+        Self::restore_terminal();
+        log::debug!("Command completed after {:?}", start_time.elapsed());
+
+        Ok(naturally_terminated)
+    }
+
+    fn create_espflash_command(&self, args: &[&str]) -> Command {
+        let mut cmd = Command::new("espflash");
```
MabezDev: This tests the system installed espflash, not the changes in a given PR/local working tree

MabezDev (changes requested):

xtask/src/test_runner.rs
```diff
@@ -0,0 +1,936 @@
...
+        }
+
+        Self::restore_terminal();
+        log::debug!("Command completed after {:?}", start_time.elapsed());
+
+        Ok(naturally_terminated)
+    }
+
+    fn create_espflash_command(&self, args: &[&str]) -> Command {
+        let debug_path = PathBuf::from("target/debug/espflash");
```
MabezDev: This can be simplified to `cargo run -p espflash --release --`, I think?

JurajSadel: True, thanks!

SergioGasquez: I think this would execute `cargo run` on the VMs, which is something we dont want as that would be very slow. We could have a `--local` (or whatever) name that uses  `cargo run` instead of trying to invoke the `espflash` bin on the system `PATH`

MabezDev: Can we do what we do in esp-hal, and use a more powerful runner to build the espflash binary for the PR, and then download it onto the VM? We can use the macos self hosted, or even a github runner would be faster than these vms :D

SergioGasquez: Thats what we currently do, we [build `xtask` and `espflash` bins on a gh-runner,](https://github.com/esp-rs/espflash/pull/912/files#diff-3f888d600c3bcb11dd1c4600c03a6249d5741149bbcb7d04bf3e7b037f5cc693R52-R65) download them in the VMs and execute them. But if we try to do `cargo run -p espflash --release --...` on the VM, it will try to buil the binaries again and it will be sloooow

MabezDev: Gotcha, so for CI we should build it first and do what you mentioned above, but for local tests we should `cargo run -p espflash --release` for local code.

JurajSadel: I think the latest changes implement this 🤞

xtask/src/test_runner.rs
```diff
@@ -36,6 +37,10 @@ pub struct RunTestsArgs {
     /// Timeout for test commands in seconds
     #[clap(short, long, default_value = "15")]
     pub timeout: u64,
+
+    /// Whether to build espflash before running tests, true by default
+    #[arg(long, value_parser = clap::value_parser!(bool))]
+    pub local_espflash: Option<bool>,
```
SergioGasquez: We can just use a `bool` and have it `false` by default, this way we dont need a value parser and we can just use it as a flag

JurajSadel: The reason I used `Option` is to make sure when no flag is provided, the espflash will be built (because of local runs). With plain `bool` and `false` by default it felt weird - maybe I missed something here

SergioGasquez: Then we should just use default `true`

PR #913, opened 2025-06-30
Rust API guidelines: `connection` module

SergioGasquez:
Slightly touched `flasher` module too for` C-CONV` to have more sense


jessebraham (approved):
Thanks, LGTM!

PR #917, opened 2025-07-01
Fix SPI detection on s3 by using the right base address

MabezDev:
Skipping change log because we haven't released since I broke it :D


espflash/src/target/mod.rs
```diff
@@ -844,13 +844,19 @@ impl Chip {
...
                 miso_length_offset: Some(0x2c),
             },
-            Chip::Esp32c2
-            | Chip::Esp32c3
-            | Chip::Esp32c5
-            | Chip::Esp32c6
-            | Chip::Esp32h2
-            | Chip::Esp32s3 => SpiRegisters {
-                base: 0x6000_3000,
+            Chip::Esp32c2 | Chip::Esp32c3 | Chip::Esp32c5 | Chip::Esp32c6 | Chip::Esp32h2 => {
```
bugadani: C2 and C3 also used to use 0x6000_2000

SergioGasquez: Esp32c2 and Esp32c3 also have `0x6000_2000` base address

bugadani (approved):

PR #920, opened 2025-07-02
Add app descriptor check to save image

SergioGasquez:
- Add app descriptor check to save image
- ~~Rename `check_idf_bootloader` to `check_app_descriptor`~~
- Only check_app_descriptor for `EspIdf` format (useless for now as its the only supported, but...)
- `ImageArgs::check_app_descriptor` is now a `bool`, as it already has a default value

### DIFF
--- a/espflash/src/image_format/idf.rs
+++ b/espflash/src/image_format/idf.rs
@@ -806,7 +806,7 @@ where
 }
 
 /// Check if the provided ELF contains the app descriptor required by [the IDF bootloader](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/bootloader.html).
-pub fn check_app_descriptor(elf_data: &Vec<u8>) -> Result<()> {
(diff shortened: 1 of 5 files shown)


espflash/src/image_format/idf.rs
```diff
@@ -806,7 +806,7 @@ where
 }
 
 /// Check if the provided ELF contains the app descriptor required by [the IDF bootloader](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/bootloader.html).
-pub fn check_idf_bootloader(elf_data: &Vec<u8>) -> Result<()> {
+pub fn check_app_descriptor(elf_data: &Vec<u8>) -> Result<()> {
```
jessebraham: This is a breaking change

SergioGasquez: Upps forgot that we cant do breaking changes anymore

Issue #919, opened 2025-07-01
"efuse blk rev" error with new bootloaders on qemu


SFrijters:
I've read the similar issues that seem to have been resolved, but I'm still a bit out of my depth.

I maintain a ["hello world" type example for running Rust code on the Espressif fork of qemu](https://github.com/SFrijters/nix-qemu-esp32c3-rust-example) and I've updated to the freshly released espflash v4.0.0.

However, after this update I get the error below, bisected to the commit that updates to the [IDF v5.4 bootloaders](https://github.com/esp-rs/espflash/commit/d76820f30aac3108e113878eb9879ca700812e21). In particular `boot_comm: Image requires efuse blk rev >= v128.50, but chip is v1.3`.

I'm using `esp-hal` at v1.0.0-beta.1 and [the latest release of the espressif/qemu fork](https://github.com/espressif/qemu/releases/tag/esp-develop-9.2.2-20250228). 

Currently I'm not sure if this is caused by something I need to update in my code, or if this is expected behaviour until qemu is somehow updated?
I read that a flash side of 64MB would be needed (if I indeed interpreted that correctly), but qemu refuses to allow that: `qemu-system-riscv32: Drive size error: only 2, 4, 8, and 16MB images are supported`. So if that is a new hard requirement I need to hold back the bootloader updates (things work with 4.0.0 with that single commit reverted).

I saw a proposed migration guide in one of the issues / PRs - something like that would be appreciated.

```
qemu-check-blinky> ELF 32-bit LSB executable, UCB RISC-V, RVC, soft-float ABI, version 1 (SYSV), statically linked, with debug_info, not stripped
qemu-check-blinky> Chip type:         esp32c3
qemu-check-blinky> Merge:             true
qemu-check-blinky> Skip padding:      false
qemu-check-blinky> App/part. size:    76,880/4,128,768 bytes, 1.86%
qemu-check-blinky> [2025-07-01T21:26:06Z INFO ] Image successfully saved!
qemu-check-blinky> esptool.py v4.8.1
qemu-check-blinky> File size: 4194304 (bytes)
qemu-check-blinky> Detected image type: ESP32-C3
qemu-check-blinky> 
qemu-check-blinky> ESP32-C3 image header
qemu-check-blinky> =====================
qemu-check-blinky> Image version: 1
qemu-check-blinky> Entry point: 0x403cc71a
qemu-check-blinky> Segments: 3
qemu-check-blinky> Flash size: 4MB
[...]

PR #921, opened 2025-07-04
Improve missing app-desc errors

SergioGasquez:
Improves https://github.com/esp-rs/espflash/issues/919#issuecomment-3026933343, related to https://github.com/esp-rs/esp-hal/pull/3750

### DIFF
--- a/espflash/src/error.rs
+++ b/espflash/src/error.rs
@@ -341,10 +341,24 @@ pub enum Error {
     )]
     PartitionTableDoesNotFit(FlashSize),
 
-    /// App descriptor not present in binary
-    #[error("{0}")]
-    #[diagnostic(code(espflash::app_desc::app_descriptor_not_present))]
-    AppDescriptorNotPresent(String),
+    /// App descriptor not present in a esp-hal binary
+    #[error(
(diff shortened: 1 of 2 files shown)


espflash/src/error.rs
```diff
@@ -341,10 +341,24 @@ pub enum Error {
     )]
     PartitionTableDoesNotFit(FlashSize),
 
-    /// App descriptor not present in binary
-    #[error("{0}")]
-    #[diagnostic(code(espflash::app_desc::app_descriptor_not_present))]
-    AppDescriptorNotPresent(String),
```
bugadani, line 347: Removing this is a breaking change

espflash/src/image_format/idf.rs
```diff
@@ -814,18 +814,9 @@ pub fn check_idf_bootloader(elf_data: &Vec<u8>) -> Result<()> {
 
     if !has_app_desc {
         if is_esp_hal {
-            return Err(Error::AppDescriptorNotPresent(
```
MabezDev, line 817: I think instead of making the breaking change to error, we can just improve these error messages, whilst we're here I think we could also address https://github.com/esp-rs/espflash/issues/919#issuecomment-3026933343

Issue #929, opened 2025-07-16
espflash ignores partition table config setting


adrian-aaen:
The saga of #156 continues!

I'm using the latest version (4.0.1) of espflash, on an ESP32S3.
Edit: I'm also using esp-hal 1.0.0-beta.1. I just noticed the rc update.

I have a `partitions.csv` file in my workspace root that looks like this:

```csv
# ESP-IDF Partition Table
# Name,   Type, SubType, Offset,  Size, Flags
nvs,      data, nvs,     0x9000,   0x6000,
phy_init, data, phy,     0xf000,   0x1000,
factory,  app,  factory, 0x10000,  0x200000,
storage,  data, fat,     0x210000, 0x5F0000,
```

Using espflash with  `--partition-table partitions.csv` gives the expected partition table:

```
I (51) boot: Partition Table:
I (54) boot: ## Label            Usage          Type ST Offset   Length
I (60) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (67) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (73) boot:  2 factory          factory app      00 00 00010000 00200000
I (80) boot:  3 storage          Unknown data     01 81 00210000 005f0000
I (86) boot: End of partition table
```

However, when creating an `espflash.toml` file with the line `partition_table = "partitions.csv"` in the workspace root, it ignores the partition table and gives me the default partition layout:

```
I (51) boot: Partition Table:
I (54) boot: ## Label            Usage          Type ST Offset   Length
I (60) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (67) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (73) boot:  2 factory          factory app      00 00 00010000 007f0000
I (80) boot: End of partition table
```

Setting a custom serial port in `espflash_ports.toml`, or setting a custom baudrate in `espflash.toml` are both recognized and applied, so espflash seems to specifically ignore only the partition table config setting.

PR #931, opened 2025-07-21
Fix using `cli::IdfFormatArgs` in `espflash.toml`

JurajSadel:
The image format refactor introduced some layers to `ProjectConfig` struct so it needs some additional parsing/checking so `espflash.toml` doesn't ignore it's options.
  
 Closes https://github.com/esp-rs/espflash/issues/929


SergioGasquez (approved):
LGTM! Thanks!

PR #933, opened 2025-07-29
Fix it so that --check-app-descriptor can actually be turned off

ivmarkov:
Subject says it all.

I also took the liberty to sophisticate the flag documentation, because "Flag to check the app descriptor in bootloader" does not sound completely correct. The app descriptor struct is not in the bootloader image in the first place (it is in the app image), and the fact that newer IDF bootloaders want it is a whole separate topic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


bugadani:
This is probably a bit more complicated than necessary, see https://github.com/esp-rs/espflash/pull/932

SergioGasquez:
> ... but is not a breaking change. Up to you, I'm fine either way.

Since its in the `cli` mod, we can allow breaking changes, just merged @bugadani PR, thanks @ivmarkov for the PR anyway!

PR #937, opened 2025-08-05
Fix compatibility with newer serialport api

stef-steos-engineering:
Newer versions have "interface" field that needs to be initialized. When using espflash as a library, if the calling application needs this feature, espflash won't compile anymore due to the dependency being added in the resulting serialport build.

Enable the feature and initialize the field in espflash, so that applications using espflash as a library can use serialport crate with this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- `espflash` enables interface feature on serialport allowing tools to build with this dependency (#937)
 
 ### Fixed
 
--- a/espflash/Cargo.toml
+++ b/espflash/Cargo.toml
@@ -51,7 +51,7 @@ miette          = "7.6"
 object          = "0.37"
 regex           = { version = "1.11", optional = true }
 serde           = { version = "1.0", features = ["derive"] }
-serialport      = { version = "4.7", default-features = false, optional = true }
+serialport      = { version = "4.7", default-features = false, features = ["usbportinfo-interface"], optional = true }
 sha2            = "0.10"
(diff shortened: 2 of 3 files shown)


stef-steos-engineering:
> This is a design issue in serialport. We don't need this (unstable!) feature, so we don't enable this feature. The solution is to fix serialport, not to modify espflash.

I agree that the docs are not up to date for serialport-rs, but the feature is present already in 4.6 and 4.7 branches of serialport crate? Is it marked unstable somewhere?

What would be a nice fix in your opinion, add the default trait in the serialport lib for that struct and the default method in espflash?

Their Cargo.toml states:
# TODO: Make the feature unconditionally available with the next major release
# (5.0) and remove this feature gate.

So from serialport-rs version 5 onward, espflash would need to add this anyway, so why not now already?

BR,
Stef

bugadani:
Quoting the [PR](https://github.com/serialport/serialport-rs/pull/101) that added the feature:

> Adding this field unconditionally is a major breaking change with
respect to semantic versioning. Putting it behind a feature gate allows
to ship this information early and also still cut 4.x releases from the
main branch while still working towards 5.0.

What @sirhcel didn't take into account is that this is a breaking feature anyway: if you enable it in one crate, it needs to be enabled in all other crates in the project. espflash shouldn't require such a breaking change from its neighbours. Therefore, the only viable course of action we can take is to wait for serialport 5.0 where this feature becomes the default.

stef-steos-engineering:
> > 
> 
> What @sirhcel didn't take into account is that this is a breaking feature anyway: if you enable it in one crate, it needs to be enabled in all other crates in the project. espflash shouldn't require such a breaking change from its neighbo

Ah cool, make sense. Thanks for taking the time to figure this out :)

PR #940, opened 2025-09-10
Add magic value for esp32c5

CameronCTS:
Add `0x5fd1_406f` as the magic value for identifying ESP32-C5. This value matches the value reported by my ESP32-C5. The attached image shows the result of running `espflash board-info` with the ESP32-C5 connected:

espflash


JurajSadel (approved):
Thanks!

SergioGasquez (commented):
How did you check the magic number of your device? Using the `espflash` logs? According to [`esp-serial-flasher` the magic number is `0x1101406F`](https://github.com/espressif/esp-serial-flasher/blob/master/src/esp_targets.c#L189)

CameronCTS:
> How did you check the magic number of your device? Using the `espflash` logs?

I ran `espflash board-info` using `espflash` 4.0.1 with my ESP32-C5-DevKitC-1 connected via USB and it reported an unknown magic value of `0x5fd1_406f`.

SergioGasquez (commented):
Sorry for the late reply! I did investigated a bit more and found out that it has several magic values, see https://github.com/espressif/esptool-js/blob/4d5231289af64659480c1119a0ba476c6cd402ea/src/targets/esp32c5.ts#L52, mind adding the other two?

Thanks!

Issue #681, opened 2024-09-16
espflash monitor takes over the terminal


cacharle:
It happens often that I cannot kill `espflash monitor` with `Ctrl-C` or `Ctrl-\` and the logs stop coming aswell. I'm not sure if my code crashed or if it's the connection to the device but it would be nice to have some error message or at least be able to kill the process instead of having to relaunch my terminal.

I use alacritty as a terminal btw:

```
❯ alacritty --version
alacritty 0.13.2 (bb8ea18e)
```

```
❯ espflash --version
espflash 3.1.0
```

SergioGasquez:
Seems similar to https://github.com/esp-rs/espflash/issues/464, I do use the same `allacritty` version on Linux and I don't recall encountering this issue.

cacharle:
My colleagues and I have observed that behavior across different terminals (Windows terminal, ubuntu's terminal, alacritty) and different machines with different OSes (Linux, Windows)

I really seems independent of the platform you're running on

cacharle:
Similar behavior when it cannot find a usb:

```
$ espflash monitor --log-format defmt --elf <elf>
[2024-10-04T15:20:49Z INFO ] 🚀 A new version of espflash is available: v3.1.1
[2024-10-04T15:20:49Z INFO ] Detected 3 serial ports
[2024-10-04T15:20:49Z INFO ] Ports which match a known common dev board are highlighted
[2024-10-04T15:20:49Z INFO ] Please select a port
[2024-10-04T15:20:49Z INFO ] Serial port: '/dev/ttyACM0'
[2024-10-04T15:20:49Z INFO ] Connecting...
r^C^C^C^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^\^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^Z^Z^Z^Z
```

cacharle:
This bug is the bane of my existence, any clue where I could start looking for it in the espflash codebase?

SergioGasquez:
[...]

PR #943, opened 2025-09-12
[windows] Don't crash during typing when the MCU isn't listening

bugadani:
This PR fixes the Windows manifestation of #681 - where instead of a hang, we have `The semaphore timeout period has expired. (os error 121)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a crash in monitor when espflash is connected via USB Serial/JTAG, and the user is typing into the monitor but the device is not reading serial input. (#943)
```
```suggestion
- [Windows] Fixed a crash in monitor when espflash is connected via USB Serial/JTAG, and the user is typing into the monitor but the device is not reading serial input. (#943)
```

PR #944, opened 2025-09-12
[Linux] Fix hanging terminal

bugadani:
Some terrible timer+signal hackery but this seems to be the recommended way in case tcdrain blocks indefinitely.

I've also added a bit of a flush-throttling so that we're not completely swamped by timeouting blocking calls on every keypress if the user mashes the keyboard.

The reason for all this is that nothing moves the data out of the USB endpoing on the device, so it doesn't accept new data.

I don't feel qualified, prepared or well enough equipped to try this on macOS.

cc #681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Windows] Fixed a crash in monitor when espflash is connected via USB Serial/JTAG, and the user is typing into the monitor but the device is not reading serial input. (#943)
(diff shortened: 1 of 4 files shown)


sirhcel:
> I don't feel qualified, prepared or well enough equipped to try this on macOS.

Kudos for this deep dive @bugadani! I'm sorry that did not manage to to look deeper into it so far, but I gave the fix from this PR a spin on macOS 15.6 on a M1 machine. I still end up in a locked state after about ten key strokes in the monitor.

bugadani:
Thanks for trying! I think it's not too surprising that a 1:1 port doesn't work - it took me quite a bit of trial and error to figure out something working. Can you perhaps share your code? Maybe it's a good enough start to work on.

sirhcel:
> Can you perhaps share your code?

I just ran espflash as of e05601d on macOS as follows:

```
$ cargo run -- flash --monitor test.elf
```

There was no bespoke code involved.

PR #950, opened 2025-09-17
Fix esp32s2 spi register base address

SergioGasquez:


espflash/src/target/mod.rs
```diff
@@ -872,7 +872,7 @@ impl Chip {
                 miso_length_offset: Some(0x28),
             },
             Chip::Esp32s2 => SpiRegisters {
-                base: 0x3f40_3000,
+                base: 0x3f40_2000,
```
SergioGasquez, line 875: Source: https://github.com/esp-rs/espflash/blob/v3.3.0/espflash/src/targets/esp32s2.rs#L186

MabezDev (approved):
Oops, that my bad I think. Thanks for fixing!

PR #951, opened 2025-09-17
move padding to 4 bytes from CLI into library

Be-ing:
Downstream applications needed to duplicate the padding to 4 bytes logic that the CLI application does (although that requirement was not documented). Otherwise flashing could fail with a RomError with error code 0.


SergioGasquez (approved):

Be-ing:
Thanks for the quick review and merge.

PR #955, opened 2025-09-26
A minimal stack-dump decoder (RV)

bjoernQ:
This is the companion PR to https://github.com/esp-rs/esp-hal/pull/4189

TL;DR in addition to the address resolving we now detect the stack-dump in the format emitted by the mentioned PR - if espflash detects such a stack dump it will try to decode it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Add chip detection based on security info, where supported (#953)
-- Support for decoding esp-backtrace's stack-dump output (#955)
+- Support for decoding `esp-backtrace`'s RISC-V stack-dump output (#955)
 
 ### Changed
 - Moved `SecurityInfo` to the `connection` module from the `flasher` module (#953)


espflash/src/cli/monitor/stack_dump.rs
```diff
@@ -0,0 +1,204 @@
...
+                                }
+                            }
+
+                            if index + func.stack_frame_size as usize > stack_bytes.len() {
+                                break;
+                            }
+
+                            let next_pc_pos = index + (func.stack_frame_size as usize - 4);
+
+                            pc = u32::from_le_bytes(
```
bugadani, line 74: Won't this blow up in trap handlers? MEPC is not on the stack, at least not in a stable position, so we can't unwind past a trap handler, can we? Likely we should look out for `.cfi_undefined ra`/`.cfi_register ra,...` markers, they show up as `RegisterRule::Undefined` or `RegisterRule::Register` in the debuginfo.

bjoernQ: Yes it will stop at the trap handler with the simple approach (and might print some garbage after it)

bjoernQ: mhhh - `_start_trap_rust` looks like this

```
0x00000e50: FDE
        length: 0x00000010
   CIE_pointer: 0x00000000
    start_addr: 0x40800502
    range_size: 0x18 (end_addr = 0x4080051a)
  Instructions:
                DW_CFA_nop
                DW_CFA_nop
                DW_CFA_nop
```

it should end up with a fame-size of 0 and we stop here

bugadani: works for me

espflash/Cargo.toml
```diff
@@ -44,6 +44,7 @@ directories     = { version = "6.0", optional = true }
 env_logger      = { version = "0.11", optional = true }
 esp-idf-part    = "0.6"
 flate2          = "1.1"
+gimli           = "0.32.3"
```
SergioGasquez, line 47: Do we need the `gimli` dependency? `addr2line` reexports it.

bjoernQ: Unfortunately `addr2line` doesn't activate all the needed features (`read-all`) on Gimli - also enabling `std` on `addr2line` doesn't (while it activates `std` in Gimli which doesn't help)

We can certainly use the re-export but still need the dependency to activate the features - which would be weird

espflash/src/cli/monitor/stack_dump.rs
```diff
@@ -0,0 +1,204 @@
+use std::{collections::HashMap, io::Write, rc::Rc};
+
+use gimli::{Section, UnwindSection};
+use object::{Object, ObjectSection};
+
+use crate::cli::monitor::symbols::Symbols;
+
+pub(crate) const MARKER: &str = "STACKDUMP: ";
```
MabezDev, line 8: Is this the identifier esp-idf uses?

* If we can use the same, I think we should (this also applies in esp-backtrace ofc)
* I'm a bit concerned about false positives with such a simple marker

bjoernQ: It is not - I decided to not adopt the same format in esp-backtrace since we just don't have all the registers (anymore) which ESP-IDF spits out - and making up register values is probably not a great solution.

I considered adding support for the ESP-IDF support in espflash - which should be trivial once we have this (hopefully). But I wanted to avoid additional noise in this PR

> I'm a bit concerned about false positives with such a simple marker

This PR does quite a few checks to make sure we don't do the wrong thing - at least I wasn't able to get it to do something harmful (i.e. we just return if something is wrong, e.g. not a hex number, the number of digits for the dump is not even, etc.). If there is something not covered yet we should definitely fix that before merging

bjoernQ: Thinking about it ... the ESP-IDF format is not trivial to integrate here since it's multi-lines ... we would need to change a bit more in espflash for that (and also another reason I opted for the "just one line with minimal information" solution as the first step)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Add chip detection based on security info, where supported (#953)
+- Support for decoding esp-backtrace's stack-dump output (#955)
```
```suggestion
- Support for decoding `esp-backtrace`'s RISC-V stack-dump output (#955)
```

SergioGasquez (approved):
LGTM! Thanks

PR #961, opened 2025-11-08
efuse: Correct block0 offset definitions for ESP32-C2, -C3, and -S3

kyrias:
According to both the technical reference manuals and to the esptool definitions these are all at 0x02C.

### DIFF
--- a/espflash/src/target/mod.rs
+++ b/espflash/src/target/mod.rs
@@ -386,14 +386,14 @@ impl Chip {
     pub fn block0_offset(&self) -> u32 {
         match self {
             Chip::Esp32 => 0x0,
-            Chip::Esp32c2 => 0x35,
-            Chip::Esp32c3 => 0x2D,
+            Chip::Esp32c2 => 0x2C,
+            Chip::Esp32c3 => 0x2C,
             Chip::Esp32c5 => 0x2C,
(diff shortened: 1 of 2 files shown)


espflash/src/target/mod.rs
```diff
@@ -386,14 +386,14 @@ impl Chip {
     pub fn block0_offset(&self) -> u32 {
         match self {
             Chip::Esp32 => 0x0,
-            Chip::Esp32c2 => 0x35,
-            Chip::Esp32c3 => 0x2D,
+            Chip::Esp32c2 => 0x2C,
```
SergioGasquez, line 389: For future reference: 
-  C2: https://github.com/espressif/esptool/blob/206970acf821635dbf9adfba99537ac420fef4d2/espefuse/efuse/esp32c2/mem_definition.py#L74
- C3: https://github.com/espressif/esptool/blob/206970acf821635dbf9adfba99537ac420fef4d2/espefuse/efuse/esp32c3/mem_definition.py#L103
- S3: https://github.com/espressif/esptool/blob/206970acf821635dbf9adfba99537ac420fef4d2/espefuse/efuse/esp32s3/mem_definition.py#L88

SergioGasquez (approved):
Thanks for the fix!

PR #962, opened 2025-11-08
eFuse write support

kyrias:
This pull requests add support for writing eFuses to all currently supported chips.  From reading both the relevant TRMs and the `espefuse` flashing process I believe that this should be correct for all of them, though I only have ESP32-S3 and ESP32-C3 devkits to test this out with.  In particular I think it would be good if it could be explicitly tested on at least an ESP32 as well since its eFuse controller is structured completely differently from the rest, as well as ESP32-S2 just to be sure since it has rather different timing parameters from the rest.

---

Since I only have a couple of devkits on hand that limits the number of tests that can be done for this feature, so I would appreciate a review and at least a tentative approval before I perform my final tests.

### DIFF
--- a/espflash/Cargo.toml
+++ b/espflash/Cargo.toml
@@ -50,6 +50,7 @@ log             = "0.4"
 md-5            = "0.10"
 miette          = "7.6"
 object          = "0.38"
+reed-solomon    = { version = "0.2.1", optional = true }
 regex           = { version = "1.11", optional = true }
 serde           = { version = "1.0", features = ["derive"] }
 serialport      = { version = "4.7", default-features = false, optional = true }
@@ -86,4 +87,4 @@ cli = [
 ]
 
 # Enables connecting to a device via serial port
-serialport = ["dep:regex", "dep:serialport", "dep:slip-codec", "dep:toml"]
+serialport = ["dep:reed-solomon", "dep:regex", "dep:serialport", "dep:slip-codec", "dep:toml"]
--- a/espflash/src/error.rs
+++ b/espflash/src/error.rs
@@ -367,6 +367,22 @@ pub enum Error {
     /// Specified eFuse block does not exist
     #[error("specified eFuse block does not exist: {0}")]
     InvalidEfuseBlock(u32),
+
+    /// Unsupported crystal frequency
+    #[error("Unsupported crystal frequency: {0}")]
+    UnsupportedXtalFrequency(String),
+
+    /// Failed to write eFuse
+    #[error("Failed to write eFuse: {0}")]
+    WritingEfuseFailed(String),
+
+    /// Timed out while waiting for eFuse controller to return to idle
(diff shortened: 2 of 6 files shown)


espflash/src/target/efuse/mod.rs
```diff
@@ -15,15 +15,28 @@ pub mod esp32p4;
...
 pub(crate) struct EfuseBlock {
+    pub(crate) index: u8,
     /// Number of registers that this block contains.
     ///
     /// Each register is a single 4-byte word.
     pub(crate) length: u8,
     /// Read address for this eFuse block.
-    #[allow(dead_code)]
     pub(crate) read_address: u32,
+    /// Write address for this eFus block.
```
SergioGasquez:
```suggestion
    /// Write address for this eFuse block.
```

espflash/src/error.rs
```diff
@@ -367,6 +367,22 @@ pub enum Error {
     /// Specified eFuse block does not exist
     #[error("specified eFuse block does not exist: {0}")]
     InvalidEfuseBlock(u32),
+
+    /// Unsupported crystall frequency
```
SergioGasquez:
```suggestion
    /// Unsupported crystal frequency
```

espflash/Cargo.toml
```diff
@@ -50,6 +50,7 @@ log             = "0.4"
 md-5            = "0.10"
 miette          = "7.6"
 object          = "0.38"
+reed-solomon    = "0.2.1"
```
SergioGasquez: This could be optional as it requires `serialport` feature if im not missing anything

SergioGasquez (commented):
Thanks for your work on this! A few fixs/comments, I wish we could test  this features on HIL, but sadly I cant think of a way with efuse beings only writable once

kyrias:
Actually, nevermind, I did something silly (or well, expected the doc examples to be correct  :sweat_smile:).  It seems connecting to the QEMU VM with `esptool`/`espefuse` *does* work.

I'll try to rig something up to at least be able to test it manually this way.  I think doing it in CI will still be too much work for this PR though, because it requires compiling the Espressif QEMU fork and figuring out how to run it properly in CI and stuff.

---

One annoyance is that we'd need to change how `espflash` handles serial ports.  Currently `espflash` *only* allows using serial ports returned by `serialport::available_ports()`, but on Linux it uses `libudev` to detect serial ports which only cares about real, physical devices and will therefore not return the PTY allocated by QEMU.

On the other hand, both QEMU and `esptool` supports using the UART over TCP instead of as a PTY, using e.g. `socket://localhost:5555`.  I wonder if it would make sense to support that as well?  (Though it's a bit awkward that they use `socket://`, hmm.)

So I think we'd need to either allow specifying any device using `--port` and not just ones returned by `serialport::available_ports()`, or implement support for connecting over TCP.  (Or both, I guess.)

PR #963, opened 2025-11-10
Allow specifying a `rom-elf`

bjoernQ:
This adds a way to specify a rom-elf ( see https://github.com/espressif/esp-rom-elfs )

The additional ELF will be used to resolve names, e.g. on ESP32-S3

Before

image

After

image

It is also used by the stack-dump analyzer used on RISC-V if stack frames are not enabled

Before (ESP32-S3)

image

After

image


bjoernQ:
> IMHO the default should be that espflash tries to figure out the chip revision and loads the appropriate elf automatically, there should be no need to faff about with the elfs unless the user uses a chip that is not supported.

I thought about that but - the way we currently do things, we probably would add all the ROMs to the resources (increasing binary size) for something which is relatively rarely really needed 🤔 

I don't have a strong oppinion about that however 🤷‍♂️

SergioGasquez (approved):
LGTM! Not sure about whats the best approach for storing the elfs though, but we can go with this way until we see it becomes an issue

PR #969, opened 2025-11-19
Fix read efuse

### LINKED ISSUE #960: eFuse API misleading with regards to eFuse field widths
The new eFuse API implemented in #847 is a bit misleading to use, and in fact many of the generated `EfuseField` definitions that it included lead to a panic if you try to read them using `Chip::read_efuse`.

`EfuseField` appears to support defining fields that are wider than a single 32-bit word, and many of the generated definitions do exactly that, but `Chip::read_efuse` only supports reading fields of at most 32 bits.  Passing in a field wider than that leads to a "attempt to shift left with overflow" panic.

bjoernQ:
Closes #960
Closes #968

This is an overhaul of our efuse reading API.

The original function is now deprecated in favor of `read_efuse_le`  (which is mostly just a port of the code found in esp-hal).

Additionally there was a problem in the xtask in calculating the block size - in general the whole approach is something we might want to reconsider.

On a side note: The current API is quite easy to use in a wrong way. e.g. `Chip::Esp32c2.read_efuse_le::<[u8;16]>(flasher.connection(), efuse::esp32c3::OPTIONAL_UNIQUE_ID)` is valid code which will produce garbage.
(Guess how I learnt about this!)

This also fixes some problems which are even visible when using the espflash CLI (look at the shown revision and MAC)

```
❯ espflash board-info
[2025-11-19T14:08:14Z INFO ] Serial port: 'COM18'
[2025-11-19T14:08:14Z INFO ] Connecting...
[2025-11-19T14:08:15Z INFO ] Using flash stub
[2025-11-19T14:08:15Z WARN ] Setting baud rate higher than 115,200 can cause issues
Chip type:         esp32c2 (revision v3.4)
Crystal frequency: 26 MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       98:7a:0a:18:20:00

Security Information:
=====================
Flags: 0x00000000 (0)
Key Purposes: [0, 0, 0, 0, 0, 0, 0]
Chip ID: 12
API Version: 1
Secure Boot: Disabled
Flash Encryption: Disabled
SPI Boot Crypt Count (SPI_BOOT_CRYPT_CNT): 0x0


❯ cargo run -- board-info
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.17s
     Running `d:\projects\upstream\espflash\target\debug\espflash.exe board-info`
[2025-11-19T14:08:05Z INFO ] Serial port: 'COM18'
[2025-11-19T14:08:05Z INFO ] Connecting...
[2025-11-19T14:08:05Z INFO ] Using flash stub
[2025-11-19T14:08:06Z WARN ] Setting baud rate higher than 115,200 can cause issues
Chip type:         esp32c2 (revision v1.0)
Crystal frequency: 26 MHz
Flash size:        4MB
Features:          WiFi, BLE
MAC address:       34:98:7a:0a:18:20

Security Information:
=====================
Flags: 0x00000000 (0)
Key Purposes: [0, 0, 0, 0, 0, 0, 0]
Chip ID: 12
API Version: 1
Secure Boot: Disabled
Flash Encryption: Disabled
SPI Boot Crypt Count (SPI_BOOT_CRYPT_CNT): 0x0


❯ c:\tools\esptool\esptool.exe get_security_info
esptool.py v4.7.0
Found 1 serial ports
Serial port COM18
Connecting....
Detecting chip type... ESP32-C2
Chip is ESP32-C2 (revision v1.0)
Features: WiFi, BLE
Crystal is 26MHz
MAC: 34:98:7a:0a:18:20
Uploading stub...
Running stub...
Stub running...

Security Information:
=====================
Flags: 0x00000000 (0b0)
Key Purposes: (0, 0, 0, 0, 0, 0, 0)
Chip ID: 12
API Version: 1
Secure Boot: Disabled
Flash Encryption: Disabled
SPI Boot Crypt Count (SPI_BOOT_CRYPT_CNT): 0x0
Hard resetting via RTS pin...
```

### DIFF
--- a/espflash/src/target/mod.rs
+++ b/espflash/src/target/mod.rs
@@ -507,6 +507,7 @@ impl Chip {
         Ok(unsafe { output.assume_init() })
     }
 
+    #[cfg(feature = "serialport")]
     fn block_address(&self, block: u32) -> u32 {
         let block0_addr = self.efuse_reg() + self.block0_offset();
 


espflash/src/target/mod.rs
```diff
@@ -416,19 +416,104 @@ impl Chip {
...
+        fn block_address(chip: &Chip, block: u32) -> u32 {
+            let block0_addr = chip.efuse_reg() + chip.block0_offset();
+
+            let mut block_offset = 0;
+            for b in 0..block {
+                block_offset += chip.block_size(b as usize);
+            }
+
+            block0_addr + block_offset
+        }
```
kyrias: Since this is the same logic as `read_efuse_raw`, any reason for not making this a method so the logic is shared between the two methods?

espflash/src/target/mod.rs
```diff
@@ -416,19 +416,104 @@ impl Chip {
...
+
+        fn read_raw(connection: &mut Connection, addr: u32) -> Result<u32, Error> {
+            connection.read_reg(addr)
+        }
```
kyrias, line 447: This seems unnecessary?

espflash/src/target/mod.rs
```diff
@@ -416,19 +416,104 @@ impl Chip {
...
+
+        // Represent output value as a bytes slice:
+        let mut output = std::mem::MaybeUninit::<T>::uninit();
+        let mut bytes = unsafe {
+            std::slice::from_raw_parts_mut(output.as_mut_ptr() as *mut u8, std::mem::size_of::<T>())
         };
```
kyrias, line 455: Could we get a comment explaining that this is safe due to the `bytemuck::AnyBitPattern` bound?

espflash/src/target/mod.rs
bjoernQ: it's just here to make is easy(ier) to keep the code here aligned with the code in esp-hal

PR #971, opened 2025-11-19
Replace old `BLOCK_SIZES` with definitions of the blocks themselves

kyrias:
The actual block definitions only exist within Python classes so this commit uses PyO3 to import the esptool.py eFuse block definitions and generates arrays of the length and read address of each eFuse block.

We need the actual read address here rather than using the old offset method because not all eFuse blocks are sequential.  (On ESP32 the block 0 read registers are followed by the block 0 write registers.)

---

TODO:
- [x] Should we deprecate the `block_size` method?
    
    It's unfortunately a public method, but it doesn't seem to me like there's any safe way for API consumers to use it for anything.

    On the other hand, that kind of also applies to `efuse_reg` and `block0_offset`.  To use them safely you have to have pretty good knowledge of the register layout per-chip, and we don't expose that right now.
- [ ] Should we also replace `efuse_reg` with ones generated from the `DR_REG_EFUSE_BASE` definitions in `esptool.py`?
- [x] I decided to leave `block` and `EfuseBlock` private for now.  It might make sense to eventually make them public, but it's much harder to make any necessary changes later if we make them public.

I'll wait for #969 to be merged and rebase on top of that before undrafting this.

### DIFF
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -4,11 +4,14 @@ version = "0.0.0"
 edition = "2024"
 publish = false
 
+[features]
+efuse-generator = ["pyo3"]
+
 [dependencies]
 chrono     = "0.4"
 clap       = { version = "4.5", features = ["derive"] }
 env_logger = "0.11"
 log        = "0.4"
-pyo3       = { version = "0.27.1", features = ["auto-initialize"] }
+pyo3       = { version = "0.27.1", features = ["auto-initialize"], optional = true }
 serde      = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.9"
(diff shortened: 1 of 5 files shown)


xtask/Cargo.toml
```diff
@@ -9,5 +9,6 @@ chrono     = "0.4"
 clap       = { version = "4.5", features = ["derive"] }
 env_logger = "0.11"
 log        = "0.4"
+pyo3       = { version = "0.27.1", features = ["auto-initialize"] }
```
SergioGasquez: Can we have this behind a feature? This way we would only need `python3` when regenerating the efuses, but the rest of the `xtask` commands can be run without it. We could add an alias, just like in https://github.com/esp-rs/esp-hal/blob/cd8addbedde320c9352020b2dde142e7e83c2aa7/.cargo/config.toml#L4 so its easier too.

kyrias: Like this?  https://github.com/esp-rs/espflash/pull/971/commits/03f86c80853e3c60702ea479a6913b2dca400791

SergioGasquez: Yeah! Thanks!

Issue #973, opened 2025-11-21
Rename `ignore_app_descriptor` flag to `ignore-app-descriptor`


SergioGasquez:
> For now, I can use --ignore_app_descriptor (which, by the way, is the only flag separated by _)
> 
> Oh that's bad UX
>  

 _Originally posted by @bjoernQ in [#927](https://github.com/esp-rs/espflash/issues/927#issuecomment-3562646692)_

We shouldn't have one flag using underscores while the rest use dash between words

PR #974, opened 2025-11-25
Change long form of save-image --ignore-app-descriptor

DaneSlattery:
Solves #973


SergioGasquez (changes requested):
Can you modify this PR to support both `ignore_app_descriptor` and `ignore-app-descriptor` while adding a TODO that the underscore version should be removed for `v5`??

Issue #927, opened 2025-07-11
ESP-IDF App Descriptor missing in your`esp-idf` application.


hazer-hazer:
Hi, I encountered this error after updating to the latest version of `espflash`:
```
Error:   × ESP-IDF App Descriptor (https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-
  │ reference/system/app_image_format.html#application-description) missing in your`esp-idf`
  │ application.
```

Cargo.toml:
```toml
# ...
esp-bootloader-esp-idf = "0.1.0"
esp-hal = { version = "^1.0.0-beta.0", features = [
  "esp32c3",
  "unstable",
  "log",
] }
esp-hal-embassy = { version = "0.7", features = ["esp32c3"] }
# ...
```

Error does not appear on espflash@3.3.0

SergioGasquez:
Hi @hazer-hazer, do you have any updates regarding this issue?

ImplFerris:
> Image contains multiple DROM segments. Only the last one will be mapped.

I checked it with esptool, the DROM gets split into two. This seems to be the reason behind the error message. 

```sh
7 segments

Segment 1: len 0x00100 load 0x3f400020 file_offs 0x00000018 [DROM]
Segment 2: len 0x00050 load 0x3ffb0000 file_offs 0x00000120 [BYTE_ACCESSIBLE,DRAM]
Segment 3: len 0x0f2f4 load 0x3f400180 file_offs 0x00000178 [DROM]
Segment 4: len 0x00674 load 0x3ffb0050 file_offs 0x0000f474 [BYTE_ACCESSIBLE,DRAM]
Segment 5: len 0x00520 load 0x40080000 file_offs 0x0000faf0 [IRAM]
Segment 6: len 0x1aa0c load 0x400d0020 file_offs 0x00010018 [IROM]
Segment 7: len 0x00d04 load 0x40080520 file_offs 0x0002aa2c [IRAM]
```

The error message i belive it is coming from "bootloader_utility.c" because of two different DROM segments there.

```c
// Find DROM & IROM addresses, to configure MMU mappings
for (int i = 0; i < data->image.segment_count; i++) {
    const esp_image_segment_header_t *header = &data->segments[i];
    if (header->load_addr >= SOC_DROM_LOW && header->load_addr < SOC_DROM_HIGH) {
        if (drom_addr != 0) {
[...]

PR #975, opened 2025-11-26
Improve app-descriptor detection

bjoernQ:
A Rust project with `strip = true` will discard the  symbol we are looking for but the section is kept.

See https://github.com/esp-rs/espflash/issues/927#issuecomment-3580431363

This should fix that situation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Corrected eFuse BLOCK0 definitions for ESP32-C2, ESP32-C3, and ESP32-S3 (#961)
 - Fixed Secure Download Mode detection on ESP32-P4 (#972)
 - Several fixes in `read_efuse` (#969)
-- Fixed a problem in detecting the app-descriptor for a Rust-std project if `strip = true` is used
-- Fixed a problem in detecting the app-descriptor for a Rust-std project if `strip = true` is used (#975)
 - Fixed a problem in detecting the app-descriptor for a project if `strip = true` is used (#975)
 
 ### Removed


bugadani:
Why the "for unknown reasons" part? https://doc.rust-lang.org/cargo/reference/profiles.html#strip says:

> strip = true is equivalent to strip = "symbols"

bjoernQ:
> Why the "for unknown reasons" part? https://doc.rust-lang.org/cargo/reference/profiles.html#strip says:
> 
> > strip = true is equivalent to strip = "symbols"

ah yeah ... ok makes a lot of sense

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Corrected eFuse BLOCK0 definitions for ESP32-C2, ESP32-C3, and ESP32-S3 (#961)
 - Fixed Secure Download Mode detection on ESP32-P4 (#972)
 - Several fixes in `read_efuse` (#969)
+- Fixed a problem in detecting the app-descriptor for a Rust-std project if `strip = true` is used
+- Fixed a problem in detecting the app-descriptor for a Rust-std project if `strip = true` is used (#975)
+- Fixed a problem in detecting the app-descriptor for a project if `strip = true` is used (#975)
```
```suggestion
- Fixed a problem in detecting the app-descriptor for a project if `strip = true` is used (#975)
```

I guess this is some rebase/merging thing?

bjoernQ: Oh yes - probably a bad rebase - thanks

espflash/src/image_format/idf.rs
```diff
@@ -809,10 +809,17 @@ where
 pub fn check_idf_bootloader(elf_data: &Vec<u8>) -> Result<()> {
```
bugadani, line 809: I've just tried using this function in probe-rs to fix https://github.com/probe-rs/probe-rs/issues/3564 and come on why are we returning a miette diagnostic from this...?

bjoernQ: we should change that! maybe there are more places like this?

bjoernQ: but ... probably for 5.x since this is already published public API :(

bugadani: OK I've managed to work around it by [this](https://github.com/probe-rs/probe-rs/pull/3664/files#diff-9945f9c276d3bffeeb0de70fc683263997ac376af5b8c69d91e321e56596d239R276) line - turning the diagnostic back into a string, and wrapping it into the `espflash::Error::AppDescriptorNotPresent` error, which it was supposed to be in the first place...

bugadani: We either shouldn't use miette in the library, or embrace it and don't expose espflash's error type, but this hybrid mess is just sad.

> but ... probably for 5.x since this is already published public API :(

We can extract the implementation into another function that returns the `Result<(), Error>` type, and just call `that().into_diagnostic()` in the current function.

bjoernQ: > We either shouldn't use miette in the library, or embrace it and don't expose espflash's error type, but this hybrid mess is just sad.

+1 ... we probably shouldn't embrace it but scrape it from the API surface

> We can extract the implementation into another function that returns the Result<(), Error> type, and just call that().into_diagnostic() in the current function.

sure - and deprecate the current function for now

Might be worth a separate issue ... pretty sure comments on a merged PR will be lost in an instant

PR #984, opened 2025-12-15
Prepare v4.3.0 release

SergioGasquez:


cargo-espflash/src/error.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused_assignments)]
```
MabezDev:
```suggestion
#![expect(unused_assignments)]
```

So we can remove this in the future when we use them

PR #993, opened 2026-01-05
Use higher default baud rate and chunks for esp32p4.

mdwn:
When flashing esp32p4, esptool is significantly faster for a simple hello world program (~5 seconds in esptool vs. ~23 seconds or so for espflash). Using some (removed) performance timing capture:

This adjusts the baud rate and chunk size for esp32p4 to match esptool's behavior, making it much quicker to flash things.

### DIFF
--- a/espflash/src/flasher/mod.rs
+++ b/espflash/src/flasher/mod.rs
@@ -597,9 +597,8 @@ impl Flasher {
 
         // Now that we have established a connection and detected the chip and flash
         // size, we can set the baud rate of the connection to the configured value.
-        if let Some(baud) = baud {
+        if let Some(baud) = flasher.baud_to_use(baud) {
             if baud > 115_200 {
-                warn!("Setting baud rate higher than 115,200 can cause issues");
                 flasher.change_baud(baud)?;
             }
         }
@@ -619,6 +618,19 @@ impl Flasher {
             .flash_target(self.spi_params, self.use_stub, false, false);
         target.begin(&mut self.connection).flashing()?;
         Ok(())
+    }
+
+    /// Determine the baud rate to use based on the chip and the provided baud
+    /// rate.
+    fn baud_to_use(&self, baud: Option<u32>) -> Option<u32> {
+        if let Some(baud) = baud {
+            return Some(baud);
+        }
+
+        match self.chip {
+            Chip::Esp32p4 => Some(921_600), // USB-OTG uses 921,600 baud for better performance
+            _ => None,
+        }
     }
 
     fn load_stub(&mut self) -> Result<(), Error> {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-Adjusted default chunk size for esp32p4 for faster flashing
+Adjusted default chunk size and baud rate for esp32p4 for faster flashing
 
 ### Fixed
 


espflash/src/flasher/mod.rs
```diff
@@ -597,9 +597,11 @@ impl Flasher {
 
         // Now that we have established a connection and detected the chip and flash
         // size, we can set the baud rate of the connection to the configured value.
-        if let Some(baud) = baud {
+        if let Some(baud) = flasher.baud_to_use(baud) {
             if baud > 115_200 {
-                warn!("Setting baud rate higher than 115,200 can cause issues");
+                if detected_chip != Chip::Esp32p4 {
+                    warn!("Setting baud rate higher than 115,200 can cause issues");
```
bugadani: I'd keep this warning either unchanged or removed completely. If you happen to have a setup where your UART-USB converter is connected with a cable (like it is the case with an esp-prog), you can run into issues with the higher baud rate. That being said, I have the baudrate set to 2MBaud on my system globally and never had any issues on any of my chips.

espflash/src/flasher/mod.rs
```diff
@@ -621,6 +623,19 @@ impl Flasher {
...
 
+    /// Determine the baud rate to use based on the chip and the provided baud
+    /// rate.
+    fn baud_to_use(&self, baud: Option<u32>) -> Option<u32> {
+        if let Some(baud) = baud {
+            return Some(baud);
+        }
+
+        match self.chip {
+            Chip::Esp32p4 => Some(921_600), // USB-OTG uses 921,600 baud for better performance
```
bugadani: Does this actually matter over USB? Virtual COM ports (other than USB-UART converters) usually ignore baud rate. USB-UART converters, on the other hand, have the issue of cable length and noise, and defaulting to 900k may be a bit more... brave than we should be.

mdwn: Truth be told I'm pretty new to this space, so I don't know the answer to this overall. I can tell you it's a pretty drastic difference for my particular setup (esp32p4 shared with WSL2).

espflash/src/flasher/mod.rs
mdwn: I've removed it here.

espflash/src/flasher/mod.rs
bugadani: I can see the difference, I'm just not sure what your actual setup is :) Which USB port do you connect to? The Function_EV_Board has 3 that you can use to download code, only one of them should be affected by this PR.

mdwn: So I'm connecting a Waveshare esp32p4 (https://www.amazon.com/dp/B0FKN8GCW6) over a USB3.0 port in Windows. I'm using `usbipd` to share this with my Windows WSL2 installation:

```
PS C:\Users\Michael Wilson> usbipd list
Connected:
BUSID  VID:PID    DEVICE                                                        STATE
...
2-13   1a86:55d3  USB-Enhanced-SERIAL CH343 (COM9)                              Shared
...
```

This shows up as `ttyACM0` in my WSL2 install:

```
➜  espflash git:(esp32p4-speed) ✗ esptool chip-id
esptool v5.1.0
Connected to ESP32-P4 on /dev/ttyACM0:
Chip type:          ESP32-P4 (revision v1.3)
Features:           Dual Core + LP Core, 400MHz
Crystal frequency:  40MHz
MAC:                30:ed:a0:e3:27:f0

Stub flasher running.

Warning: ESP32-P4 has no chip ID. Reading MAC address instead.
MAC:                30:ed:a0:e3:27:f0

Hard resetting via RTS pin...
```

So my setup may be somewhat unorthodox.

bugadani: Ah great, so that board has a USB-UART converter for some reason, probably because the P4 chip is an older revision - I'm not sure if the USB_SERIAL_JTAG interface works on those or not, that converter shouldn't actually be necessary.

I'm not entirely sure we should change the default baud rate. You can set your own, but the defaults should be conservative - otherwise we may just as well raise it for other chips. The chunk size is a safer change, but I'll let someone else decide on this one.

mdwn: I'll revert the baud rate changes.

bugadani: I wouldn't be too hasty, my opinion is one thing, the precedent in esptool is another :)

mdwn: Hahah sounds good! I'll unresolve and see what the discussion leads to here.

espflash/src/target/mod.rs
```diff
@@ -945,7 +945,10 @@ impl Chip {
 
     /// Write size for flashing operations
     pub fn flash_write_size(&self) -> usize {
-        FLASH_WRITE_SIZE
+        match self {
+            Chip::Esp32p4 => 0x800, // USB-OTG uses 2KB blocks for better performance
```
SergioGasquez, line 949: When using OTG `esptool` also uses this for S2 and S3, but we should only change it for USB-OTG

espflash/src/flasher/mod.rs
SergioGasquez: Do you mind sharing some esptool references for this change?

mdwn: So, embarrassingly, I didn't notice a baud rate override in a script I'm using to flash to an esp32p4 target using esptool... soooo let's pretend this PR never happened. 😰 Thanks for pushing back on this, I appreciate it.

bugadani: Thanks for trying :)

PR #996, opened 2026-01-08
Use esp-hal runners for HIL

SergioGasquez:
This PR does a few things:
-  Migrates espflash HIL runners from the current VMs to esp-hal runners (RPis) 
- Adds P4 support
  - This required adding an elf file for HIL tests (created with esp-idf)
    - This required updating the P4 bootloader (I only updated this one, maybe I should update all the targets?) 
- Adds C5 support
  - This required adding an elf file for HIL tests (created with esp-idf)
- Adapted the `test_list_ports` to also accept USB Serial/Jtag ports
- HIL tests are now run in `USB` and `UART` ports when possible
- Runners now have an `udev` rule that symlinks devices to `/dev/serial_ports/uart ` and `/dev/serial_ports/usb`

HIL CI Run: https://github.com/esp-rs/espflash/actions/runs/21246847880

Pending things:
- [x] Connect S2
- [x] Connect C6 uart port
- [x] Test uart and usb ports when possible
- [x] Fix C5 errors
- [x] Fix P4 errors
- [x] Edit required jobs for merging a PR


.github/workflows/hil.yml
```diff
@@ -120,3 +120,103 @@ jobs:
...
+        uses: docker/setup-docker-action@v4
+
+      - name: Install toolchain
+        uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: stable
+          components: rust-src
+
+      - name: Install cross
+        run: cargo install cross
```
bugadani: You can `binstall` cross. For probe-rs, we use:

```yaml
      - name: Install cross
        uses: taiki-e/install-action@v2.65.6
        with:
          tool: cross
```

Which I believe skips building it.

.github/workflows/hil.yml
```diff
@@ -28,95 +28,116 @@ concurrency:
...
-    runs-on:
-      [
-        self-hosted,
-        linux,
-        x64,
-        "${{ matrix.board.mcu }}${{ matrix.board.freq }}",
-      ]
-
+    name: HIL | ${{ matrix.target.soc }}
+    needs: build-packages
```
SergioGasquez, line 77: We could duplicate a bit of code to speed things up, and only wait for the packages of their arch to be finished, not sure if its worth though

espflash/resources/70-espflash.rules
SergioGasquez, line 1: Not sure if this is the best way, and I guess this could supersede probe-rs udev rules in the runner (https://github.com/esp-rs/esp-hal/tree/main/hil-test#rpi-setup), but having the symlink to `/dev/serial_ports/usb` and `/dev/serial_ports/uart` is quite useful I think.

I only tested this locally, I haven't yet added this file to the runners, I'll wait for some input before

MabezDev: I like this a lot! We should use these symlinks on all runners if possible.

Regarding the use rules, I think we still need both but we can put both in the udev rules directory.

SergioGasquez: Sorry, the comment was outdated, all the runners already have this udev rules settled, we are using both probe-rs and this ones, but since this is `70-` and probe-rs is `69-`, this ones should take preference.

PR #998, opened 2026-01-29
feat: include serial number in list-ports output

vfabr:
The current list-ports output lacks serial numbers, making it impossible to identify specific devices when multiple ESP32 are connected and now it is possible to get a port for the flash by PROBE.

```
$ probe-rs list  
The following debug probes were found:
[0]: ESP JTAG -- 303a:1001:B4:3A:45:AE:52:FC (EspJtag)
[1]: ESP JTAG -- 303a:1001:B8:F8:62:FC:17:8C (EspJtag)
$ espflash list-ports
/dev/ttyACM0  303A:1001:B4:3A:45:AE:52:FC          Espressif  USB JTAG/serial debug unit
/dev/ttyACM1  303A:1001:B8:F8:62:FC:17:8C          Espressif  USB JTAG/serial debug unit
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+
 ### Added
 
+- Add serial number to `list-ports` output for easier device identification (#998)
 - eFuse write support (#962)
 - esp32p4rc1 support for esp32p4 chips with revision < 3.0 (#922)
 - Support for binaries with INIT_ARRAY sections, which is needed for esp32p4 support. (#991)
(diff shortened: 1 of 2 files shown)


SergioGasquez (commented):
Thanks for the contribution! I think it makes sense to add this information, but maybe on a different format? For USB-Serial/JTAG the format is good but for UART ports:
`/dev/ttyUSB0 10C4:EA60:dc414a793059ed11a99bb5301d62bc44 Silicon Labs CP2102N USB to UART Bridge Controller`

SergioGasquez:
> For my current aims this is a good enough solution because I use only JTAG boards now. But I agree that could be more common and better implementation here. I need more time to try to do it :)

Hi @vfabr, would you like to finish the PR? We are planning a minor release soon, also, it's completely fine if the PR doesn't make it and its merged for the next release, which will likely be a major.

PR #1002, opened 2026-02-03
Add SDM HIL 

SergioGasquez:
This PR add HIL SDM tests on a ESP32C6 (using one of our VMs) and also:
- Disables the bootloader write protection (see https://github.com/esp-rs/espflash/pull/990#issuecomment-3852930827)
- Unifies the `save-image` and `wirte-bin` tests into a single one

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - esp32p4rc1 support for esp32p4 chips with revision < 3.0 (#922)
 - Support for binaries with INIT_ARRAY sections, which is needed for esp32p4 support. (#991)
 - Add sha256 calculation to match esptool generated binaries (#991)
-- Support flashing in secure download mode (#990)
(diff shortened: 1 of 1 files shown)


xtask/src/test_runner.rs
```diff
@@ -137,16 +141,21 @@ impl TestRunner {
...
-        while start_time.elapsed() < timeout {
+        while start_time.elapsed() < timeout + grace {
             if let Ok(Some(_)) = child.try_wait() {
                 terminated_naturally = true;
                 break;
             }
             thread::sleep(Duration::from_millis(100));
         }
 
+        if !terminated_naturally && let Ok(Some(_)) = child.try_wait() {
```
MabezDev: I don't understand this sequence, why we have another iteration of the loop unrolled here?

SergioGasquez: It was there to check they `try_wait` one last time after the 100ms sleep, as we dont check after the last iteration. I was investigating some transient HIL failures (where it would timeout and still succeed), but this didnt help either, so I'll just remove it)

PR #1005, opened 2026-02-11
Read xtal freq directly from PCR

bugadani:
The previous algorithm estimated a very inaccurate 0 as the xtal freq.

### DIFF
--- a/espflash/src/target/mod.rs
+++ b/espflash/src/target/mod.rs
@@ -911,12 +911,14 @@ impl Chip {
             }
             Chip::Esp32c3 => Ok(XtalFrequency::_40Mhz), // Fixed frequency
             Chip::Esp32c5 => {
-                const UART_CLKDIV_REG: u32 = 0x6000_0014; // UART0_BASE_REG + 0x14
-                const UART_CLKDIV_MASK: u32 = 0xfffff;
-                const XTAL_CLK_DIVIDER: u32 = 1;
+                const PCR_SYSCLK_CONF_REG: u32 = 0x6009_6110; // PCR_BASE_REG + 0x110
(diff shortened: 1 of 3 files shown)


espflash/src/target/mod.rs
```diff
@@ -911,12 +907,14 @@ impl Chip {
             }
             Chip::Esp32c3 => Ok(XtalFrequency::_40Mhz), // Fixed frequency
             Chip::Esp32c5 => {
-                const UART_CLKDIV_REG: u32 = 0x6000_0014; // UART0_BASE_REG + 0x14
-                const UART_CLKDIV_MASK: u32 = 0xfffff;
-                const XTAL_CLK_DIVIDER: u32 = 1;
+                const PCR_SYSCLK_CONF_REG: u32 = 0x6009_6110; // PCR_BASE_REG + 0x110
+                const PCR_CLK_XTAL_FREQ_MASK: u32 = 0x7F;
+                const PCR_CLK_XTAL_FREQ_SHIFT: u32 = 24;
```
SergioGasquez, line 916: Source for reference: https://github.com/espressif/esptool/blob/deca775e38ce5f096b6e94a8bd5db74c6fc44c45/esptool/targets/esp32c5.py#L59-L61

espflash/src/target/mod.rs
SergioGasquez, line 926: I wonder if we still need this if we only support 48Mhz

espflash/src/target/mod.rs
bugadani: Actually I just read this out of the TRM, but sure we can say it comes from esptool :D

bugadani:
I think we might want to keep the 40MHz branch just in case, I'm also not sure if there are preproduction chips floating around that should at least not be misreported.

PR #1032, opened 2026-04-29
flasher: add fallible `try_connect()`

### LINKED ISSUE #1019: Don't consume `Connection` when `Flasher::new()` fails
## Motivations

A serial connection can be turned into a `Connection`, which can then be used to flash using a `Flasher`. When this is done, it can be turned back into a `Connection` and then back into a `Port`.

However, if `Flasher::connect()` fails, it returns an `Error` and the `Connection` is lost. This makes it not possible to recover the stream, requiring the serial port to be re-opened.

- Would you like to implement this feature? y

## Solution

This could be accomplished with a `::try_connect()` that returns `Result<Self, (Connection, Error)>`. The existing `::connect()` would be rewritten using this paradigm, and would be replaced with `Self::try_connect(...).map_err(|e| e.1)`.

## Alternatives

I see talk of a breaking API change to pull `Connection` out of `Flasher`, which could be an ideal alternative if breaking changes were allowed.

xobs:
Add a function to `Flasher` to support returning the `Connection` if an error occurs. This allows for retrying the operation with the same file descriptor without closing and opening the serial port.

This closes #1019

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `Flasher::try_connect()` as a variant of `Flasher::connect()` that returns the `Connection`
+
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


SergioGasquez (approved):

PR #1039, opened 2026-06-15
Fix C3 HIL failures

playfulFence:
If I understand things correctly, on C3 USB_UART_CHIP_RESET doesn't seem reinit the debug unit.
So, this is probably why we saw that DMI timeout thing.

Testing proves my theory, let's see how HIL tests will behave

### DIFF
--- a/espflash/src/cli/mod.rs
+++ b/espflash/src/cli/mod.rs
@@ -730,10 +730,22 @@ pub fn serial_monitor(args: MonitorArgs, config: &Config) -> Result<()> {
             .map_err(Error::from)
             .wrap_err_with(|| format!("Failed to reopen serial port {port_name}"))?;
         monitor_args.no_reset = true;
-        return monitor(serial, elf_refs, pid, monitor_args, args.connect_args.non_interactive);
+        return monitor(
+            serial,
+            elf_refs,
+            pid,
+            monitor_args,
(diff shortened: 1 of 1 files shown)


bugadani (approved):
No issues from me if this really fixes the issue. +1 for the non-interactive check.

espflash/src/cli/mod.rs
```diff
@@ -709,6 +709,36 @@ pub fn serial_monitor(args: MonitorArgs, config: &Config) -> Result<()> {
...
     let elf_refs = elfs.refs();
 
+    // UsbJtagSerialReset issues USB_UART_CHIP_RESET, which does not re-initialize
+    // the RISC-V debug module, which causes probe-rs JTAG sessions to timing
+    // out on DMI access. rtc_wdt_reset performs a full chip reset that fixes this,
+    // but also causes USB re-enumeration, so the port must be reopened.
+    if args.connect_args.non_interactive
+        && !monitor_args.no_reset
+        && flasher.connection().is_using_usb_serial_jtag()
+        && chip.can_rtc_wdt_reset(flasher.connection())?
```
SergioGasquez: `can_rtc_wdt_reset()` returns `UnsupportedFeature` for USB-Serial-JTAG chips that don't support this helper (C6/H2/C5/C61) this makes HIL tests fail

PR #1043, opened 2026-06-18
Fix P4 flashing

SergioGasquez:
- Add ESP32-P4 ECO6/ECO7 flash power-on handling before SPI flash attach                                          
- Fix flash finish command encoding to use a 4-byte payload, matching `esptool`                                     
- Avoid rebooting directly from the stub `FLASH_DEFL_END` command

cc @bjoernQ


bjoernQ (approved):
Thanks a lot!

PR #1044, opened 2026-06-22
Update esp32p4 v3 bootloader

SergioGasquez:


bjoernQ (approved):
This works fine for me

PR #1046, opened 2026-06-25
honor minimal revision in ELF from esp-hal

playfulFence:
see https://github.com/esp-rs/esp-hal/pull/5812

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Update ESP32-P4 stub (#1037)
 - Update ESP32-P4 rev >= 300 bootloader (#1044, #1045)
-
-- Update ESP32-P4 rev > 300 bootloader (#1044)
 - MIN_CHIP_REVISION from ELF metadata is now taken into consideration when setting min_chip_rev_full in the flash image header. (#1046)
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


MabezDev:
We should probably have an escape hatch to flash it anyways I think, do you know if esptool has something like this?

playfulFence:
I think their `--force` flag is basically what we're looking for 🤔. They use it as an escape hatch basically for everything, I've implemented it. However this might have some problems architecture-wise 🤔 

Like, if we` --force` espflash, then we'll get an assertion error in esp-hal. Then we change the limitation in esp-hal config, then `--force` in espflash makes no sense, since espflash will be able to flash it anyway with the min revision bumped in esp-hal config

CHANGELOG.md
```diff
@@ -11,12 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Added `Flasher::try_connect()` as a variant of `Flasher::connect()` that returns the `Connection`
 - Add ESP32-P4 ECO6/ECO7 flash power-on handling (#1043)
+- `--force` flag to ignore min_chip_revision (#1046)
 
 ### Changed
 
 - Update ESP32-P4 stub (#1037)
 - Update ESP32-P4 rev >= 300 bootloader (#1044, #1045)
 
+- Update ESP32-P4 rev > 300 bootloader (#1044)
```
```suggestion
```

SergioGasquez (changes requested):
I think we would also need to update cargo-espflash

espflash/src/bin/espflash.rs
SergioGasquez, line 298: When using `--ram` we are not checking it

SergioGasquez (approved):
LGTM! Thanks

PR #1055, opened 2026-07-17
Add xtask command to generate bootloaders

### LINKED ISSUE #1054: Repo/project/xtask command for reliably producing boot loaders with the correct configurations
More and more chips have individual settings we must configure, and or amend. We've been able to do this manually for a while but as our chip line up expands, we'll need some infra to automate most of this for us.

SergioGasquez:


xtask/src/bootloader_builder.rs
```diff
@@ -0,0 +1,349 @@
...
+    run(Command::new("cmd")
+        .arg("/C")
+        .arg(script)
+        .env("IDF_TOOLS_PATH", idf_tools_path))
+}
+
+#[cfg(unix)]
+fn shell_quote(value: impl AsRef<OsStr>) -> String {
+    let value = value.as_ref().to_string_lossy();
+    format!("'{}'", value.replace('\'', "'\\''"))
```
bugadani, line 370: This replaces a single `'` quote with `\''`, is that right? 🤔

SergioGasquez: To prevent paths with spaces/quotes from breaking the generated shell script

bugadani: I get the idea, I just don't know if the replacement has one more quote than intended or not.

SergioGasquez: I think it turns `/tmp/foo'bar` into `/tmp/foo'\''bar` and then wraps into quotes: `'/tmp/foo'\''bar' `, which is ok for posix  shells

playfulFence (approved):
I like the changes, thanks!

PR #1056, opened 2026-07-27
Bump MSRV to 1.95.0, memory range sourcing from metadata

### LINKED ISSUE #1047: Use esp-metadata-generated instead of duplicating `Chip` info in espflash
A lot of them flash range addresses are wrong, but given how addr_is_flash is used, not critical:

https://github.com/esp-rs/espflash/blob/main/espflash/src/target/mod.rs#L591-L664

playfulFence:
closes https://github.com/esp-rs/espflash/issues/1047
this is waiting for https://github.com/esp-rs/esp-hal/pull/5966 to get merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  MSRV: "1.88"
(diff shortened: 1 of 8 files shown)


SergioGasquez (approved):
Thanks! Looks much better now!

PR #1061, opened 2026-08-14
Skip stub for non-flashing commands

SergioGasquez:


bugadani:
This is fine, but this still runs through the connection process. For `reset` we really just need to toggle DTR/RTS, there's no need for any connection unless we really want to make sure that we mess with our own MCUs. 🤔

SergioGasquez:
I just created https://github.com/esp-rs/espflash/issues/1062 too

PR #1058, opened 2026-07-31
HIL optimizations

SergioGasquez:
- Speeds up HIL tests with higher transfer baud rates.                                                            
- Improves test coverage and timeout handling.                                                                    
- Recovers unresponsive native USB devices through UART.                                                          
- Prevents device reset commands from hanging indefinitely.                                                       
- Hardens workflow concurrency and fork PR support.                                                               
- Reduces artifact uploads and retention.

PR #1063, opened 2026-08-14
Add esp32h4 support

playfulFence:


espflash/src/target/mod.rs
```diff
@@ -144,6 +146,7 @@ impl Chip {
             Chip::Esp32c6 => [0x2CE0_806F].contains(&value),
             Chip::Esp32c61 => [0x7211_606f].contains(&value),
             Chip::Esp32h2 => [0xD7B7_3E80].contains(&value),
+            Chip::Esp32h4 => [0x0].contains(&value),
```
SergioGasquez: Both P4 and H4  have `0` magic number, since they rely on security info. Since H4 is matched first, fallback detection can misidentify P4 as H4 when `security_info()` fails. Not sure if we should cover this case

playfulFence: I'd honestly just write "return" on both of them. 0 is not the real magic number IIRC, it's just that it doesn't exist and things are preferably done via security info 🤷🏼

playfulFence:
HIL tests are failing independently from this PR, the changes here do not touch anything failure-related for any other chip ([PR](https://github.com/esp-rs/espflash/pull/1058) that intents to resolve this)

espflash/src/target/mod.rs
SergioGasquez: Agreed, `esptool` treats H4 and P4 as not using magic values. We should just return `false` for them so a failed `security_info()` produces an error.

espflash/resources/bootloaders/manifest.yaml
```diff
@@ -66,6 +66,14 @@ bootloaders:
       - CONFIG_ESPTOOLPY_FLASHSIZE_64MB=y
       - CONFIG_ESPTOOLPY_FLASHSIZE="64MB"
 
+  - name: esp32h4
+    target: esp32h4
+    output: esp32h4-bootloader.bin
+    preview: true
+    configs:
+      - CONFIG_ESPTOOLPY_FLASHSIZE_64MB=y
+      - CONFIG_ESPTOOLPY_FLASHSIZE="64MB"
```
bugadani, line 75: Per datasheet and TRM this should be 32MB

playfulFence: that tool we have is wrong, both esptool and espflash detect 8MBs of flash size on the chip which corresponds to 64KB MMU page size 
image


I'd leave everything as is here, what I'm doing here looks the most correct.

playfulFence: CONFIG_ESPTOOL_FLASHSIZE_32MB (nor 64MB lol) doesn't exist it seems like, not sure where does it come from 😅 
image

bugadani (approved):

PR #1065, opened 2026-08-26
Add threshold of failed HIL jobs to prevent merge block

### LINKED ISSUE #1064: Rework HIL CI
In case something dies for whatever reason(which happens way too often with esp32/s2 😄  ) - it effectively makes merging in espflash (which currently requires all the HIL tests) impossible without bypassing rules (as it happened in https://github.com/esp-rs/espflash/pull/1063). 

We might need to rework this, either un-require esp32 + some other problematic runners or implement the same failure percentage concept as we do in esp-hal (if more than X% of HIL runs failed - kick out of the merge queue)

I'd prefer the second option here

playfulFence:
closes https://github.com/esp-rs/espflash/issues/1064

Mostly copied approach from esp-hal + tweak: instead of strict percentage allow a certain amount of failed HIL runs

### DIFF
--- a/.github/scripts/hil-gate.js
+++ b/.github/scripts/hil-gate.js
@@ -25,6 +25,7 @@ function classifyMatrixJob(job) {
   if (
     conclusion === "failure" ||
     conclusion === "cancelled" ||
+    conclusion === "timed_out" ||
     conclusion === null
   ) {
     return { kind: "failed" };
--- a/.github/scripts/hil-gate.js
+++ b/.github/scripts/hil-gate.js
@@ -1,0 +1,119 @@
+// Merge-queue gate for HIL matrix legs.
+//
+// Passes when at most MAX_FAILED_RUNS executed `HIL | <soc> | <port>` jobs
+// failed or were cancelled. Dual-port chips (uart + usb) count as two jobs.
+// Skipped jobs are excluded. `HIL (SDM)` is informational and is not gated.
+//
+// Classification uses the job conclusion rather than a single step: a dead
+// runner typically fails "Prepare device" and never reaches "Run all tests".
+
+const MAX_FAILED_RUNS = 6;
+
+function isHilRunMatrixJob(name) {
+  // Matches "HIL | …" but not "HIL (SDM) | …".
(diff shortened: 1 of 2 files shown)


.github/scripts/hil-gate.js
```diff
@@ -0,0 +1,119 @@
...
+  let page = 1;
+
+  while (true) {
+    const { data } = await github.rest.actions.listJobsForWorkflowRun({
+      owner,
+      repo,
+      run_id,
+      per_page: 100,
+      page,
+    });
```
SergioGasquez, line 91: I think `listJobsForWorkflowRun` defaults to `filter: "latest"`, see https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#list-jobs-for-a-workflow-run--parameters. So, rerunning one HIL job makes the gate evaluate only that job, if there were some previous failures, not sure if they will be picked

playfulFence: Hmmm, makes sense 🤔 

Does it mean we have the same "hole" in esp-hal CI then?

SergioGasquez: Yeees! I just checked and I think we also have the same issue there!

playfulFence: I'll take care of this in esp-hal too 👍🏼

.github/scripts/hil-gate.js
```diff
@@ -0,0 +1,147 @@
...
+    return { kind: "skipped" };
+  }
+  if (conclusion === "success") {
+    return { kind: "passed" };
+  }
+  if (
+    conclusion === "failure" ||
+    conclusion === "cancelled" ||
+    conclusion === null
+  ) {
```
SergioGasquez, line 30: [Speaking of similar issues as esp-hal](https://github.com/esp-rs/esp-hal/pull/6199#discussion_r3861552526) for this, a `timed_out` job, will be treated as unexpected, so it will not count as one fail for the merge. I think, `timed_out` should be considered as failed.

SergioGasquez (approved):
LGTM! Thanks

PR #1072, opened 2026-09-16
Decode stack memory dumps generated by ESP-IDF firmware

ivmarkov:
**Code generated by Fable**

Turned out slightly more involved than implementing an ESP-IDF-specific stack memory dump screen scraper.

Background:
The existing stack walker in `espflash` uses the `.eh_frame` section instead of the `.debug_frame` section; eh frames are typically used for C++-style unwinding and - unless special measures are taken - are therefore **flashed**, as they need to be available during program execution.

(https://github.com/esp-rs/esp-hal/pull/4189 is therefore an additional trick in the `esp-hal` linker script that makes sure the eh_frame is NOT flashed by marking it with the "INFO" attribute.)

My hypothesis as to why debug frames were not used instead (they are automatically "info" and therefore never flashed) is **not** because they are inferior in any way for stack walking, but because eh frames have a nice side effect: Rust functions which are typically never returning (`foo() -> !`, and there could be **a lot** of those when a program panics/aborts at the tip of the stack) - in the absence of `-C force-unwind-tables` (which generates the eh frame section for a Rust program) are marked with `noreturn nounwind` LLVM attributes which means the LLVM codegen for these does NOT emit assembly that saves the **ra** (and other) registers of the caller because, well, the function is not expected to return anyway. Since the `ra` (= return address) register is not in there, the frame becomes much more difficult to walk therefore (though Fable came up with a way).

In contrast, when `-C force-unwind-tables` is there, the `-> !` functions are additionally marked with `uwtable` which seems to force LLVM to emit assembly for saving `ra` and friends.

===

Now, it is easy to ask ESP-IDF to generate an eh_frame section as well (a Kconfig param), but very difficult to ask it to NOT put it in the firmware. Therefore Fable implemented support for regular debug frames, and the ability to walk stack frames that don't have a return address (as long as these are < 8).

===

There is also now a new `Collector` trait, with one impl for the esp-backtrace stack memory screen scraping, and another - for ESP-IDF. And a chain of the two.


espflash/src/cli/mod.rs
```diff
@@ -34,7 +34,7 @@ use serialport::{FlowControl, SerialPortInfo, SerialPortType, UsbPortInfo};
 
 use self::{
     config::Config,
-    monitor::{LogFormat, check_monitor_args, monitor},
+    monitor::{LogFormat, UnwindTables, check_monitor_args, monitor},
```
SergioGasquez, line 37: This comment should be in `check_monitor_args` implementation (`monitor/mod.rs`) but I'm putting it here for simplicity, as you didn't edit that function.

`check_monitor_args` should also check the new `unwind_tables` option to produce a warning and avoid commands like `espflash write-bin --unwind-tables eh-frame` without `--monitor` being silently ignored.

SergioGasquez (commented):
Thanks Ivan for the PR! I'm definitely not an expert in this topic, but the implementation looks good to me, just a few things: 
- Should we add some HIL test to validate the new `montior` args? Maybe we could add an ESP-IDF app bin for a single target, not to bloat the repo full of bins.
- Added a nit comment

espflash/src/cli/mod.rs
ivmarkov: Done.

ivmarkov:
> * Should we add some HIL test to validate the new `montior` args? Maybe we could add an ESP-IDF app bin for a single target, not to bloat the repo full of bins.

Right, I've asked it to add a very simple ESP-IDF app - basically the "hello world" one with an `abort` at the end.

ivmarkov:
The esp32c6 HIL failed but is this flaky? It used not to fail before? not sure how to re-trigger it...

SergioGasquez:
> The esp32c6 HIL failed but is this flaky? It used not to fail before? not sure how to re-trigger it...

Rerunning it made it pass, so I would assume it's part of his flakyness

SergioGasquez (approved):

PR #1074, opened 2026-09-17
Add slash commands for HIL testing

### LINKED ISSUE #978: Use /<commands> for HIL testing like we do in esp-hal
See https://github.com/esp-rs/esp-hal/issues/4429

SergioGasquez:
Adds slash commands for selectively running HIL tests: 
                                                        
 - /hil quick                                           
 - /hil full                                            
 - /hil sdm                                             
 - /hil <chip1> [chip2...]                              
 - /trust user                                         
 - /revoke user                                        
                                                        
 Full HIL remains required in the merge queue.


.github/workflows/hil.yml
```diff
@@ -61,74 +124,17 @@ jobs:
...
-            runner: esp32s31-jtag
-            port: uart
-
+      matrix: ${{ fromJSON(needs.resolve.outputs.matrix) }}
     steps:
-      - uses: actions/checkout@v6
-        with:
-          repository: ${{ github.event.inputs.repository || github.repository }}
-          ref: ${{ github.event.inputs.branch || github.ref }}
-
```
playfulFence, line 131: I think we need checkout for run-tests, we're using stuff from `tests` directory, so just espflash binary is not enough here 🤔

.github/README.md
playfulFence, line 1: Please rename it to be HIL.md, Git seems to be working in a way that it prefers `README.md` in `.github`, not in the root of the repo. Proof - espflash repo in your fork: https://github.com/SergioGasquez/espflash/tree/feat/slash-commands. The new README is the main page README there 😄

.github/workflows/hil.yml
```diff
@@ -264,20 +241,16 @@ jobs:
...
     timeout-minutes: 30
     env:
       ESPFLASH_PORT: /dev/serial_ports/esp32c6
-
     steps:
-      - uses: actions/checkout@v6
-        with:
-          repository: ${{ github.event.inputs.repository || github.repository }}
-          ref: ${{ github.event.inputs.branch || github.ref }}
-
```
playfulFence, line 280: Same here, I think it needs a checkout 🤔

.github/workflows/hil.yml
```diff
@@ -11,33 +16,87 @@ on:
...
   CARGO_TERM_COLOR: always
   ESPFLASH_SKIP_UPDATE_CHECK: "true"
   RUST_LOG: debug
 
-# Cancel any currently running workflows from the same PR, branch, or
-# tag when a new workflow is triggered.
-#
-# https://stackoverflow.com/a/66336834
 concurrency:
   cancel-in-progress: true
```
playfulFence, line 47: I'd avoid `cancel-in-progress` at all, we've cancelled it for esp-hal in https://github.com/esp-rs/esp-hal/pull/4758, I don't think it'll do much good for us here either. 

I think we can just remove the entire concurrency here

.github/scripts/hil.test.js
```diff
@@ -0,0 +1,75 @@
...
+});
+
+test("rejects unknown chips", () => {
+  assert.match(parseHilCommand("/hil esp8266").error, /Unsupported/);
+});
+
+test("resolves all configured targets and rejects invalid selections", () => {
+  const all = resolveMatrix("all");
+  assert.deepEqual(all.chips, allowedChips);
+  assert.equal(all.matrix.target.length, 16);
```
playfulFence, line 42: will break on a new target? 
maybe `config.targets.length` is better than a hard-coded value

playfulFence (changes requested):

.github/README.md
SergioGasquez: Thats nasty, good catch!

.github/workflows/hil.yml
```diff
@@ -292,3 +268,73 @@ jobs:
...
+        with:
+          script: |
+            const suite = process.env.SUITE;
+            const requireSuccess = (name, result) => {
+              if (result !== 'success') core.setFailed(`${name} did not succeed (${result})`);
+            };
+            requireSuccess('Matrix resolution', process.env.RESOLVE_RESULT);
+            if (suite !== 'sdm') requireSuccess('HIL build', process.env.BUILD_RESULT);
+            // SDM remains informational in the merge queue, matching the
+            // existing policy, but an explicit command must report its result.
```
playfulFence: this is true, but hil-gate will still wait for run-sdm-tests. We should split SDM-part of this step into a separate job with ` needs: [resolve, build-sdm-packages, run-sdm-tests]` that will only do the `Check SDM Results`

.github/workflows/hil.yml
```diff
@@ -292,3 +268,73 @@ jobs:
...
+            }
+
+      - name: Evaluate device results
+        if: >-
+          github.event_name != 'pull_request' &&
+          needs.resolve.outputs.suite != 'sdm' &&
+          needs.build-packages.result == 'success'
+        uses: actions/github-script@v7
+        env:
+          MAX_FAILURES: ${{ github.event_name == 'merge_group' && '6' || '0' }}
```
playfulFence: I'm a bit lost in MAX_FAILURES vs DEFAULT_MAX_FAILED_RUNS 😄 

This feels like we can do something like 


```
env: 
    STRICT: ${{ github.event_name != 'merge_group' }}
```

and then pass `maxFailures: process.env.STRICT === 'true' ? 0 : undefined` to the evaluateHilRun, `undefined` will take default value => `DEFAULT_MAX_FAILED_RUNS` from the script

playfulFence (changes requested):

playfulFence (approved):
Thanks! Good first step, we can fine-tune it in case of emergency 👍🏼

